PR #15, opened 2014-12-31
Typeahead

noahmanger:
Adds typeahead support on all search fields using the [Twitter Typeahead library](https://github.com/twitter/typeahead.js).


static/js/modules/typeahead.js
```diff
@@ -0,0 +1,152 @@
+'use strict';
+var Handlebars = require('handlebars');
+var events = require('./events.js');
+
+module.exports = {
+  init: function() {
+    function filterCandidates(result) {
+      // Label House, President, Senate
+      var officeFull;
+      switch(result.elections[0].office_sought) {
```
theresaanna: It would be a little bit nicer if we could map these values instead of using a `switch` statement. Something like:

```
var officeMap = {
  'H': 'House of Representatives'
  // and so on
}

officeFull = officeMap[result.elections[0].office_sought]
```

static/js/modules/typeahead.js
```diff
@@ -0,0 +1,152 @@
+'use strict';
+var Handlebars = require('handlebars');
+var events = require('./events.js');
+
+module.exports = {
+  init: function() {
+    function filterCandidates(result) {
```
theresaanna: I think it would be nice if we could move these functions defined within the `init()` method outside of the `module.exports` object. You will still be able to call them from within `init()` and then `init()` reads more clearly.

static/js/modules/typeahead.js
```diff
@@ -0,0 +1,152 @@
...
+          officeFull = "House of Representatives";
+          break;
+        case "S":
+          officeFull = "Senate";
+          break;
+        case "P":
+          officeFull = "President";
+          break;
+      }
+      var filteredResults =   {
```
theresaanna: Its considered best practice to declare all `var` statements at the top of a function. You can populate them later. For example, you could do one of two things here:

```
function filterCandidates(result) {
  var officeFull,
        filteredResults;

  // stuff

  filteredResults = {
    // stuff
  }
}
```
or
```
function filterCandidates(result) {
  var officeFull,
        filteredResults = {};

  // stuff

  filteredResults.name = result.name;
  // and so on
}
```

theresaanna:
:+1:

PR #16, opened 2015-01-03
Selenium!

theresaanna:
This introduces Selenium tests run in Sauce Labs. They will run in Travis CI. It uses mock API responses taken from the API running with the test database.

It also includes some refactor bits here and there. @noahmanger, this will affect the candidate/committee API request work you've done, so let's merge this PR first and I will update your changes accordingly.

###### Browse candidates and committees
- navigate to browse page
- content is correct
- results count is correct
- pagination works
- pagination counts update
- candidates: state filter works
- committees: type filter works

###### Search
- main page search works
- search results shows
- can click view all to results
- name filter is active and populated when table page is loaded
- headers are correct

##### Running tests
If you want to run tests from your local environment, you will need to run https://docs.saucelabs.com/reference/sauce-connect/

Then, run each test using `node tests/selenium/test-name.js`.

If you want to simulate the test environment, load your local with ?test=true. Also, see static/js/init.js and uncomment the require for mocks.js and any mocks you may need included in the doc.ready.


theresaanna:
Will fail until this is sorted out. https://github.com/18F/openFEC-web-app/commit/8d14589661aab9ea4878358efe1e3e8951b9ce4c#commitcomment-9141791

static/js/init.js
```diff
@@ -11,6 +11,7 @@ var tables = require('./modules/tables.js');
 var progress = require('./modules/progress.js');
 var errors = require('./modules/errors.js');
 var router = require('./modules/router.js');
+//var mocks = require('../../tests/mocks/mocks.js');
```
theresaanna: I'm leaving these comments in place because they are useful in writing/debugging tests off of a local environment and it gets stripped out when Browserify condenses the JS into a production version.

micahsaul:
Tests look good to me. I leave the JS refactor stuff to someone who does not have the JS skills of a drugged water buffalo.

theresaanna:
@noahmanger reviewed the refactor stuff and said it was fine so I'm merging to be able to work off of these changes.

PR #19, opened 2015-01-13
Totals summary in templates

theresaanna:
Implements totals summary for principal candidate committees on candidate pages on the client side.


shared/candidate-helpers.js
```diff
@@ -1,5 +1,28 @@
+var _ = require('underscore');
 var committeeHelpers = require('./committee-helpers.js');
```
noahmanger: I don't think we need this line. I think I originally put it in there because I was using the committeeHelpers.buildCommitteeContext() function but no more.

static/js/modules/api.js
```diff
@@ -64,6 +90,35 @@ var filterLoadHandler = function(e) {
...
+
+var loadCandidate = function(e) {
+    var candidatePromise = callAPI(buildURL(e));
+
+    candidatePromise.done(function(data) {
+        var totalsPromises = [],
+            cycle = data.results[0].elections[0].election_year;
+
+        totalsPromises.push(callAPI('/rest/total/' + data.results[0].elections[0].primary_committee.committee_id + '?election_cycle=' + cycle));
+// data.results[0].elections[0].affiliated_committees[0].committee_id
```
noahmanger: Remove this line?

theresaanna: Oh good call. Thanks. That's for when I implement affiliated committee totals, but I should remove it from the code.

PR #22, opened 2015-01-30
Moving to Flask

theresaanna:
Sorry, this is a giant one.

Removing almost all of the client side app. Leaving behind pieces that deal with filter formatting and activation, and Typeahead. (Which works, btw. The back end just times out sometimes, presumably because certain requests take too long. I imagine "Obama" and "Romney" like we're all trying have a bazillion committees to pull totals for. Try "pelosi" and choose Nancy Pelosi and her first committee.)

I've only successfully written tests for the `openfecwebapp.data_mappings` module. Maybe I'm just tired and have been working too long, but I can't quite figure out how to test the views and api_caller modules without headaches. I am happily taking suggestions.

I really want to remove the `load_totals` calls from `openfecwebapp.views.render_page` but am not sure what is best. Again, would like feedback.

Also includes some of @noahmanger's markup and css work.

The JS version is frozen in https://github.com/18F/openFEC-web-app/tree/js-version.


__init__.py
```diff
@@ -0,0 +1,73 @@
...
+from openfecwebapp.api_caller import (load_search_results,
+    load_single_type, load_totals)
+
+app = Flask(__name__)
+
+def _filter_empty_params(params):
+    new_params = {}
+    for p in params:
+        if params[p]:
+            new_params[p] = params[p]
```
cmc333333: One liner:
```python
{key: value for key, value in params.items() if value}
```

theresaanna: Oooh, I likes me a good one liner. Thanks.

__init__.py
```diff
@@ -0,0 +1,73 @@
...
+
+    return new_params
+
+def _add_fields_star(params):
+    # move from immutablemultidict -> multidict -> dict
```
cmc333333: Are you sure you *want* a mutable dict? If you want this to be a docstring, it needs to be a string rather than a pound.

__init__.py
```diff
@@ -0,0 +1,73 @@
...
+def home_page():
+    return render_template('search.html');
+
+@app.route('/search')
+def search():
+    query = request.args.get('search')
+    if query:
+        return render_search_results(load_search_results(query), query)
+    else:
+        return home_page()
```
cmc333333: This means you have two urls with the same content -- did you mean to redirect instead?

theresaanna: Yes, I do.

openfecwebapp/api_caller.py
```diff
@@ -0,0 +1,40 @@
...
+
+from openfecwebapp.local_config import api_location
+
+# api urls are singular, public urls are plural
+type_map = {
+    'candidates': 'candidate',
+    'committees': 'committee',
+    'candidate': 'candidate',
+    'committee': 'committee'
+}
```
cmc333333: It looks like you're using a plural form in the __init__.py page, but immediately converting to singular here. Why not just use singular throughout? I'd also recommend using constants rather than strings.

__init__.py
theresaanna: So, the reason for this is that I don't think we want / to always go to /search. Conversely, I don't think the links to search result pages should be at /. Thoughts?

openfecwebapp/data_mappings.py
```diff
@@ -0,0 +1,195 @@
...
+
+    if c.get('address'):
+        committee['state'] = c['address'].get('state')
+
+    if c.get('status'):
+        committee['type'] = c['status'].get('type_full', '')
+        committee['designation'] = c['status'].get(
+            'designation_full', '')
+
+    committee['nameURL'] ='/committees/' + c.get('committee_id', '')
```
cmc333333: Just wanted to highlight this potential gotcha for the future. If the committee_id isn't present, the nameURL won't point to a committee but it also won't be _empty_, so you can't do any sort of `if nameURL` style logic.

openfecwebapp/data_mappings.py
```diff
@@ -0,0 +1,195 @@
...
+    return committee
+
+# maps committee values shown on candidate pages
+def _map_committee_values(ac):
+    c = {}
+    c['id'] = ac.get('committee_id', '') 
+    c['name'] = ac.get('committee_name', '')
+    c['designation'] = ac.get('designation_full', '')
+    c['designation_code'] = ac.get('designation', '')
+    c['url'] = '/committees/' + ac.get('committee_id', '')
```
cmc333333: Same highlight as above.

openfecwebapp/data_mappings.py
```diff
@@ -0,0 +1,195 @@
...
+        'total_disbursements': totals.get('total_disbursements'),
+        'total_cash': reports.get('cash_on_hand_end_period'),
+        'total_debt': reports.get('debts_owed_by_committee')
+    }
+
+    # format totals data in US dollars
+    for v in value_map:
+        if value_map[v]:
+            totals_mapped[v] = locale.currency(
+            value_map[v], grouping=True)
```
cmc333333: Nice use of `locale`

openfecwebapp/data_mappings.py
```diff
@@ -0,0 +1,195 @@
...
+            candidate['related_committees'] = True
+
+        # affiliated committees = committee_type_map, 
+        # plus more we ignore for the candidate pages
+        if c_e.get('affiliated_committees'):
+            candidate['affiliated_committees'] = []
+            for i in range(len(c_e['affiliated_committees'])):
+                candidate['affiliated_committees'].append(
+                    _map_committee_values(c_e[
+                        'affiliated_committees'][i]))
```
cmc333333: Might be easier to read:
```python
candidate['affiliated_committtees'] = [
    _map_committee_values[c]
    for c in c_e['affiliated_committees']
]
```

openfecwebapp/data_mappings.py
```diff
@@ -0,0 +1,195 @@
...
+
+            candidate['authorized_committees'] = []
+            candidate['leadership_committees'] = []
+            candidate['joint_committees'] = []
+
+            for i in range(len(candidate['affiliated_committees'])):
+                cmte = candidate['affiliated_committees'][i]
```
cmc333333: Might be easier to read if you iterate over the elements rather than the index:

```python
for cmte in candidate['affiliated_committtees']:
```

openfecwebapp/data_mappings.py
```diff
@@ -0,0 +1,195 @@
...
+                cmte_type = cmte['designation_code']
+                # drop anything that's not of the types we're
+                # interested in
```
cmc333333: I wonder if this sort of filter should eventually make its way into the API, rather than forcing the client to do it.

openfecwebapp/data_mappings.py
```diff
@@ -0,0 +1,195 @@
...
+    committee = map_committee_table_values(c)
+
+    if c.get('address'):
+        committee['address'] = {
+            'state': c['address'].get('state'),
+            'zip': c['address'].get('zip'),
+            'city': c['address'].get('city'),
+            'street_1': c['address'].get('street_1'),
+            'street_2': c['address'].get('street_2')
+        }
```
cmc333333: Would it make sense to just assign a copy rather that specifying the fields?
```python
committee['address'] = copy(c['address'])
```

openfecwebapp/tests/data_mappings_test.py
```diff
@@ -0,0 +1,163 @@
+from flask import Flask
+from flask.ext.testing import TestCase
+from openfecwebapp.data_mappings import *
+
+class TestDataMappings(TestCase):
```
cmc333333: Yay tests

openfecwebapp/views.py
```diff
@@ -0,0 +1,70 @@
...
+generate_pagination_values)
+from werkzeug.exceptions import abort
+
+def render_search_results(results, query):
+    candidates = []
+    committees = []
+
+    if results.get('candidates') and results['candidates'].get(
+        'results'):
+        for c in results['candidates']['results']:
```
cmc333333: Style question: which do you like more, the explicit guard or the implicit iteration?

```python
if results.get('candidates'):
    for c in results['candidates'].get('results', []):
```

You could take it a step further, but this seems ugly in Python
```python
for c in results.get('candidates', {}).get('results', []):
```

openfecwebapp/views.py
```diff
@@ -0,0 +1,70 @@
...
+            candidates.append(map_candidate_table_values(c))
+
+    if results.get('committees') and results['committees'].get(
+        'results'):
+        for c in results['committees']['results']:
+            committees.append(map_committee_table_values(c))
+
+    # if true will show "no results" message
+    no_results = True if not len(candidates) and not len(committees) else False
```
cmc333333: Would this work as

```python
no_results = not candidates and not committees
```

openfecwebapp/views.py
```diff
@@ -0,0 +1,70 @@
...
+def render_table(data_type, results, params, url):
+    # if we didn't get data back from the API
+    if len(results) is 0:
+        abort(500)
+
+    results_table = {}
+    results_table[data_type] = []
+    heading = "Browse " + data_type
+
+    results_table['pagination'] = generate_pagination_values(results, params, url, data_type)
```
cmc333333: I really like how you've separated concerns. Each of your functions is short and legible even though there's a lot of complexity under the covers.

openfecwebapp/views.py
```diff
@@ -0,0 +1,70 @@
...
+
+    for r in results['results']:
+        results_table[data_type].append(type_map[data_type](r))
+
+    return render_template(data_type + '.html', **results_table)
+
+def render_page(data_type, c_data):
+    # not handling error at api module because sometimes its ok to 
+    # not get data back - like with search results
+    if len(c_data['results']) is 0:
```
cmc333333: Maybe a little more Pythonic:

```python
if c_data['results']:
```

templates/candidates-single.html
```diff
@@ -0,0 +1,93 @@
...
+              <dt>Principal Campaign Committee</dt>
+              <dd><a href="{{ primary_committee.url }}">{{ primary_committee.name }}</a></dd>
+            </dl>
+            {% endif %}
+
+            {% if authorized_committees %}
+            <dl>
+              <dt>Authorized Committees</dt>
+              {% for committee in affiliated_committees %}
+              <dd><a href="{{ committee.nameURL }}">{{ committee.name }}</a></dd>
```
cmc333333: Noted this before, but here's the consequence: `committee.nameURL` might point to an invalid address, no?

templates/partials/candidate-filings-table.html
```diff
@@ -0,0 +1,28 @@
...
+      </tr>
+      <tr>
+        <td><a href="#">Filing name</a></td>
+        <td>Statement of candidacy</td>
+        <td>1/1/2012</td>
+      </tr>
+      <tr>
+        <td><a href="#">Filing name</a></td>
+        <td>Statement of candidacy</td>
+        <td>1/1/201-</td>
```
cmc333333: Just highlighting the dummy data.

cmc333333:
It's amazing how easy you make porting this over seem, @theresaanna . If you all want to keep mustache/handlebar/etc. templates, you can; you'll just need to find the Python library that implements them instead of using jinja.

I made a bunch of style comments, but they are all about tightening things up. In general, try to add docstrings to your functions rather than `#` comments. Also, remember that you can iterate over the elements of a collection rather than the index. If you really need the index, you can 
```python
for idx, value in enumerate(your_list)
```

__init__.py
cmc333333: I don't have enough knowledge of your application to speak to where content should live, it just seems wrong that identical content is present at two urls (`/` and `/search`). One should be canonical and the other should redirect to it, right?

openfecwebapp/views.py
theresaanna: So I specifically want to check if `c_data['results']` is empty. Would this suggestion return false if it was? I read it as "does c_data have a key results".

theresaanna:
@cmc333333 I looked at the available Flask + Handlebars libraries and they were sparse. I know I could use the regular Python Handlebars library, but I wanted to keep things very clean and Flask, and I didn't want to put in the effort to figure it out, tbh. I think Jinja is nice, and I'm glad to have the concept of template blocks now. 

Thank you for your review. I suspected many things could be more pythonic, so this is great feedback.

__init__.py
theresaanna: @noahmanger I'd like your input on this.

Right now, going to either http://site/ or http://site/search gets you the home page. The search disambiguation page is at /search?search=thing. 

There is duplication because I feel like its weird to send people to /search when they come into /, and weird to have the search disambiguation at /?search=thing.

Do you have a different idea? Thanks.

openfecwebapp/data_mappings.py
theresaanna: Good call. AFAIK, the *only* thing we can be sure will exist is the ID, which is why I was lax. You're probably right and I should tighten this up.

openfecwebapp/data_mappings.py
theresaanna: I didn't even know that was possible. Thanks!

openfecwebapp/api_caller.py
theresaanna: Good call, this is a leftover from when I was loading data in the view. I can remove it.

openfecwebapp/views.py
cmc333333:
```python
c_data['results']
``` will throw an exception if "results" is not present in the dict. Are you thinking of `get`?

File: openfecwebapp/data_mappings.py (hunk 11 above)
Comment (theresaanna) [inline openfecwebapp/data_mappings.py]: Hah. Hahahaha. Hahahahahahaha. I was so in the mode of.... ok. This never happened.

File: openfecwebapp/views.py (hunk 16 above)
Comment (theresaanna) [inline openfecwebapp/views.py]: I can have a situation where:

```
c_data = {
    'results': []
}
```
and I don't want to continue if results has no contents. Am I missing something?

__init__.py
noahmanger: I agree that the `/search` should redirect to `/` rather than present the same content ant two different URLs.

PR #25, opened 2015-02-05
Hopefully worthwhile readability refactor of render_page

theresaanna:


templates/candidates-single.html
```diff
@@ -58,8 +58,8 @@ <h4 class=>Party</h4>
             {% if leadership_committees %}
             <dl>
               <dt>Leadership PACs</dt>
-              {% for committee in leadership_committees %}
-              <dd><a {% if committee.name_url %}href="{{ committee.name_url }}"{% endif %}>{{ committee.name }}</a></dd>
+              {% for c_id in leadership_committees %}
+              <dd><a {% if leadership_committees[c_id].url %}href="{{ leadership_committees[c_id].url }}"{% endif %}>{{ leadership_committees[c_id].name }}</a></dd>
```
cmc333333: Should the link be here at all? This'll generate an <a> with no href if no url is present, right? Perhaps that's a super rare case

openfecwebapp/views.py
```diff
@@ -56,16 +56,18 @@ def render_page(data_type, c_data):
...
-                            results)
+            for cmte_id in tmpl_vars['affiliated_committees']:
+                if cmte['designation_code'] in committee_type_map: 
+                    committee_ids.append(cmte['id'])
+                    cmte_type = committee_type_map[
+                        cmte['designation_code']]
+
+            results = load_totals(",".join(committee_ids))
+            if results['results']:
+                for r in results['results']:
```
cmc333333: Style thing: you might think this is cleaner than the if-for combo:
```python
for r in results.get('results', []):
```

openfecwebapp/data_mappings.py
```diff
@@ -174,21 +174,22 @@ def map_candidate_page_values(c):
...
-            candidate['authorized_committees'] = []
-            candidate['leadership_committees'] = []
-            candidate['joint_committees'] = []
+            }
+            candidate['authorized_committees'] = {}
+            candidate['leadership_committees'] = {}
+            candidate['joint_committees'] = {}
 
-            for cmte in candidate['affiliated_committees']:
+            for cmte_id in candidate['affiliated_committees']:
```
cmc333333: Since you're not using `cmte_id` here, you could just do 
```python
for cmte in candidate['affiliated_committees'].values():
```

openfecwebapp/views.py
```diff
@@ -56,16 +56,18 @@ def render_page(data_type, c_data):
...
-
-                    results = load_totals(",".join(committee_ids))
-                    if results['results']:
-                        tmpl_vars[c_type][i]['totals'] = map_totals(
-                            results)
+            for cmte_id in tmpl_vars['affiliated_committees']:
+                if cmte['designation_code'] in committee_type_map: 
+                    committee_ids.append(cmte['id'])
+                    cmte_type = committee_type_map[
+                        cmte['designation_code']]
```
cmc333333: I don't understand this loop. It looks like you're getting `cmte_id` but never using it?

cmc333333:
Yay reducing API calls and yay using meaningful field names instead of the magical zero index.

I've noted some questions on these loops. In both this and the API, there's a ton of iterating through loops and accessing "special" dictionary keys. This makes it very, very hard to verify that the data makes sense and is very error prone.

Might I suggest that a long term strategy is to create a library of data models which is shared between the API and the front end? This doesn't get you static safety, but at least makes it clear what you _expect_ each piece of data to look like.

templates/candidates-single.html
theresaanna: What do you propose as an alternative?

theresaanna:
I'm going to close this pull request without merging. I'm incorporating the feedback into the code, but its going to be part of a larger pull request.

templates/candidates-single.html
cmc333333: Sorry to miss this -- how should this appear if the committee has no url? Perhaps, the the `a` be a `span`. Perhaps there shouldn't be a `dd` at all. Perhaps there's enough confidence that the url will always be present, in which case, there's no need for the "if" at all. I'm not sure; just noting that this implies the corner case is addressed when it's hasn't been.

theresaanna: I just commented belatedly, you didn't miss anything.

Thanks. I see your point. I think, as we simplify the connection between the templates and the data from the API, we'll just build this URL in the template using the ID. And, worst case scenario, we have an A with no href. I personally don't think that's the biggest deal, could be wrong.

PR #26, opened 2015-02-06
adding fields into totals calls and fixing disbursement field implementation

theresaanna:
...tion


LindsayYoung:
This looks right to me

PR #31, opened 2015-02-11
Fixing candidate pages

theresaanna:
This request primarily fixes candidate pages. As of now, they are not rendering properly when using the api_performance branches.

It also incorporates some rework of render_page, which made sense in the context of some of this work.

I am under no illusions that these are the most elegant solutions ever. These are patches so that people can use the candidate page. I figure that things are going to change a lot once the API work is done, so I am not fretting the solutions for the moment.

Right now, totals data for committees that are not primary aren't showing. I thought, since these pages are broken, I'd get this in regardless. Unsure if its an API issue, front end issue or both.


openfecwebapp/data_mappings.py
```diff
@@ -94,6 +94,9 @@ def map_totals(t):
     from the 'totals' endpoint for use on candidat
     and committee pages
     """
+    if not t['results'][0].get('reports') and not t['results'][0].get('totals'):
```
cmc333333: Should this be an `or`?

cmc333333:
:+1: Python stuff at the top -- there's a lot of diff associated with the merges, though

PR #36, opened 2015-02-13
Adding messages if there's no financial data

noahmanger:
Will need attention to the exact copy, but the html / conditionals are
there.


rjmajma:
That was fast.

PR #37, opened 2015-02-19
add --cached flag to cache API requests

shawnbot:
This PR adds support for a `--cached` flag that enables caching of API calls, facilitating a much faster frontend development workflow. You can run it like so:

```sh
python __init__.py --cached
```

I *almost* put the cache installation code in `__init__.py`, but ended up moving it to `api_caller.py` so that all of the request-specific stuff lived there, which makes it simpler to modify how caching is enabled if we were to ever move away from using `requests` for some reason. Happy to entertain another way of doing this, or to use `argparse` instead of just sniffing `sys.argv`.


theresaanna:
Great call. Thanks!

PR #40, opened 2015-02-23
New selenium tests

micahsaul:
Beginnings of a test suite for browser testing. Currently just verifying that pages load, but we should add more tests as we go along. Note that I'm not mocking anything, as I think it's important to know when API changes break the site. I used the @attr('selenium') decorator to mark these tests so we can skip them in Travis.


micahsaul:
18F/openFEC/issues/404

PR #53, opened 2015-03-06
Glossary

noahmanger:
This is the start of the glossary. It works well enough to start testing it with people, but still lacks many important definitions and not many terms have actually been marked up as necessary. 

My sense is that creating the definitions and applying them throughout will be more of an ongoing content task.

@theresaanna definitely welcoming js best practice input on this. I tried to follow the conventions we've been using, but it wouldn't surprise me if I missed something.


static/js/modules/glossary.js
```diff
@@ -0,0 +1,124 @@
...
+    setDefinition,
+    showGlossary,
+    hideGlossary,
+    clearTerm;
+
+// Indexing the terms and then lowercasing the indices
+indexToLowercase = function(arrayOfObjects, index){
+    var key, 
+        newKey,
+        object = _.indexBy(arrayOfObjects, index),
```
theresaanna: I would just say we should rename things like `object` and `newObject` since they don't describe the values within. Maye `terms` and `lowercaseTerms` respectively?

noahmanger: So I had that, then I made it generic thinking that this function might be useful in other circumstances. But otherwise, yeah, I agree.

theresaanna: I can see that. My suggestion would be that we should make it describe its current use and refactor if/when we use it elsewhere.

theresaanna:
@noahmanger I'm stepping away from the computer. I know its easiest if this is merged before you do the homepage, so you're good to merge whenever. I didn't want to go ahead and do it in case you wanted to change those var names.

PR #56, opened 2015-03-10
Homepage redesign

noahmanger:
This is the newly styled homepage and also a restyled glossary that is fixed positioned over all the content.

Homepage copy is still a work in progress, as is copy throughout the site.

This completes https://github.com/18F/openFEC/issues/433


static/js/init.js
```diff
@@ -48,15 +48,15 @@ $(document).ready(function() {
...
 
     // Forcing the .main-container to fit the entire height of the page
-    var setMainSize = function() {
-        var headerHeight = $('.site-header').height(),
-            footerHeight = $('footer').height(),
-            pageHeight = $('.page-wrap').height(),
-            mainHeight = pageHeight - headerHeight;
-            $('.main-container').height(mainHeight);
-    }
+    // var setMainSize = function() {
```
theresaanna: Do we know that we'll need these commented out bits in the future?

noahmanger: I'm just not sure. I'm still figuring out how I'm doing the slide-out sides and this might be necessary. I'll remove it if we end up not needing it.

theresaanna: That's perfect. 

I'm just picky about lots of commented code sitting around. I feel like it gets confusing.

PR #58, opened 2015-03-13
Adding http auth

theresaanna:


__init__.py
```diff
@@ -11,8 +13,14 @@
 
 app = Flask(__name__)
 
+app.config['BASIC_AUTH_USERNAME'] = username
+app.config['BASIC_AUTH_PASSWORD'] = password
+app.config['BASIC_AUTH_FORCE'] = True
+
+basic_auth = BasicAuth(app)
+
+""" move from immutablemultidict -> multidict -> dict """
```
arowla: Docstrings should come after the function definition.

theresaanna: Got myself confused there then. Thanks.

PR #63, opened 2015-03-17
Making web app Cloud Foundry-ready and creating deploy script

theresaanna:
**So that you don't break your local**, you will need to define the HTTP username and password as environment vars `FEC_WEB_USERNAME` and `FEC_WEB_PASSWORD`. Locally, you can set these to whatever you want.

local_config.py is no longer needed or supported. Set environment variables if necessary for debug, api location, port and host. You don't need to, though, because the app sets defaults.

Defaults:
debug = False
api_location = http://localhost:5000
host = 0.0.0.0
port = 3000
username and password that should be used for deploy are the ones currently used for the staging site.

Environment vars (app var = environment var):
port = VCAP_APP_PORT
debug = FEC_WEB_DEBUG
api_location = FEC_WEB_API_URL
host = FEC_WEB_HOST
username = FEC_WEB_USERNAME
password = FEC_WEB_PASSWORD

If you want to deploy to Cloud Foundry, use `deploy.sh`, which builds JS and pushes to CF. It takes four args: CF space, CF app, HTTP auth username and password. 

Notes:
- I'm not happy with the debug var handling, but its the best I could come up with given that CF can only receive a string, not a boolean value.
- I know the default clause in setting the API location in the deploy script will cause the API requests to be unsuccessful. I think this is ok for now until we actually have a prod API. These are the only boxes we're deploying to.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,17 +1,18 @@
 #!/bin/bash
 
 # usage:
-# ./deploy.sh space-name app-name
+# ./deploy.sh space-name app-name web-username web-password
 # You must be logged into Cloud Foundry
 # You must also follow the npm installation instructions here:
 # https://github.com/18F/openFEC-web-app#installing
 # if you haven't already set this up in your environment
 
 api_url=""
+debug=""
 
 if [ -z "$1" ] || [ -z "$2" ]; then
-  echo "Please pass in the name of the Cloud Foundry app you want to"
-  echo "and the space it is in"
+  echo "Please pass in the name of the Cloud Foundry app you want,"
+  echo "the space it is in, and the HTTP username and password"
   echo "Example: './deploy.sh dev-web fec-web-dev'"
   echo "Run 'cf spaces' to see possible spaces."
   echo "Run 'cf apps' to see possible apps within your current space."
@@ -19,9 +20,10 @@ if [ -z "$1" ] || [ -z "$2" ]; then
 fi
 
 if [ $2 = "fec-web-dev" ]; then
-  api_url="http://fec-api-dev.cf.18f.us/"
+  api_url="http://fec-api-dev.cf.18f.us"
+  debug="true"
 elif [ $2 = "fec-web-stage" ]; then
-  api_url="http://fec-api-stage.cf.18f.us/"
+  api_url="http://fec-api-stage.cf.18f.us"
 else
   api_url="prod"
 fi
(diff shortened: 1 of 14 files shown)


manifest.yml
```diff
@@ -3,3 +3,7 @@ applications:
 - name: fec-web
   host: fec
   buildpack: python_buildpack
+env:
+  FEC_WEB_DEBUG: True
+  FEC_WEB_API_URL: 'http://fec-api-dev.cf.18f.us/'
```
theresaanna: Need input: should the deploy script change this URL to match the app we're deploying?

deploy.sh
```diff
@@ -0,0 +1,27 @@
...
+  echo "Please pass in the name of the Cloud Foundry app you want to"
+  echo "and the space it is in"
+  echo "Example: './deploy.sh dev-web fec-web-dev'"
+  echo "Run 'cf spaces' to see possible spaces."
+  echo "Run 'cf apps' to see possible apps within your current space."
+  exit 1
+fi
+
+echo "Building JS..."
+npm run build
```
cmc333333: Might be worth checking the return code here -- if build fails for whatever reason, you're still pushing

theresaanna: Good call.

manifest.yml
cmc333333: It makes sense to me for the deploy script to set the `DEBUG` and apis based on the second parameter (assuming you have enough information from "fec-web-dev" to convert to "fec-api-dev" etc.).

I see two ways to do this. 
1. Leave the manifest without these env vars, `cf push` then `cf set-env` and restart the app
2. Instead of `cf push`ing with this manifest, do something like: 
```bash
cp manifest.yml /tmp/
echo "env:" >> /tmp/manifest.yml
echo "    FEC_WEB_DEBUG: Some Determined Value" >>  /tmp/manifest.yml
echo "    FEC_WEB_API_URL: Some Other Value" >>  /tmp/manifest.yml
cf push "$2" -f /tmp/manifest.yml
```

There might also be a way to set env vars at `push` time, but I didn't see one.

theresaanna: I didn't think of using `set-env`. I will experiment with this.

cmc333333:
Makes sense to me.

cmc333333:
@LindsayYoung and @arowla also :+1:'d

PR #72, opened 2015-04-14
560 implement report charts

msecret:
Completed most of the bug, still have to reconfigure chart UI to put debts on the bottom.

Read commit messages for explanations.

I want to add tests but was unsure of the structure of them.


noahmanger:
What do you mean putting the debts on the bottom? I'm wondering if you're looking at an older mock. They debts used to go below the x-axis, but now should just go above.

noahmanger:
Oops, I see what I did now. The one in InVision was old, sorry. I just updated it:

http://invis.io/WY266IM4F

msecret:
OK that's good, it makes my life easier

msecret:
screen shot 2015-04-14 at 5 30 34 pm

noahmanger:
Sorry, just got around to reviewing this. I can't speak to the Python, but the HTML / Sass looks good to me. @LindsayYoung or @theresaanna can you review the other stuff?

openfecwebapp/api_caller.py
```diff
@@ -37,8 +37,9 @@ def load_single_type(data_type, c_id):
 
 def load_cmte_financials(committee_id):
     r_url = '/committee/' + committee_id + '/reports'
+    limited_r_url = limit_by_amount(r_url, 4)
     t_url = '/committee/' + committee_id + '/totals'
-    reports = _call_api(r_url, {})
+    reports = _call_api(limited_r_url, {})
```
jmcarp: Why limit results from candidates but not committees?

openfecwebapp/api_caller.py
```diff
@@ -48,3 +49,6 @@ def load_cmte_financials(committee_id):
 def install_cache():
     import requests_cache
     requests_cache.install_cache()
+
+def limit_by_amount(curr_url, amount):
+    return curr_url + '?page=1&per_page=' + str(amount * 2)
```
jmcarp: Prefer URL libraries over string concatenation for building URLs, e.g.
```python
from urllib.parse import urlencode

def limit_by_amount(curr_url, amount):
    query = urlencode({'page': 1, 'per_page': amount * 2})
    return '{0}?{1}'.format(curr_url, query)
```

Which takes care of type-casting and joining for you.

LindsayYoung:
I am looking at Obama for America since it has a lot of info associated with it `/committee/C00431445`

I am not seeing the items in the charts and tables that come from the most recent report:
https://api.open.fec.gov/committee/C00431445/reports?api_key=xxx
 -  `cash_on_hand_end_period `, (this should be charted at the bottom)
 - `cash_on_hand_beginning_period` (that is listed above)
 - `coverage_end_date` (this should used as a label for the last chart)

The candidate page has the same cash and debt problem
/candidate/P80003338

On the Receipts and Disbursements charts, it looks like the underling dates are wrong the beginning and end period are the same. I will make a separate ticket for that- we should check again once we have the golden gate data.

Also, @noahmanger if we are only listing one cash on hand, I think cash on hand end of period would be better since it is more recent.

noahmanger:
@LindsayYoung I agree that we should just be `cash_on_hand_end_of_period`

arowla:
Should this PR be tagged '[WIP]' or is it done?

PR #79, opened 2015-04-20
[wip] Newrelic

LindsayYoung:
I got new relic working on the test instance. Still need to up the security settings.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -7,6 +7,7 @@ from openfecwebapp.views import (render_search_results, render_table,
 from openfecwebapp.api_caller import (load_search_results,
     load_single_type, load_single_type_summary,
     install_cache)
+from newrelic import agent as nr_agent
 
 import sys
 import locale
@@ -84,6 +85,7 @@ def currency_filter(num, grouping=True):
     return locale.currency(num, grouping=grouping)
 
 if __name__ == '__main__':
+    nr_agent.initialize()
     if '--cached' in sys.argv:
         install_cache()
     app.run(host=host, port=int(port), debug=debug)
--- a/__init__.py
+++ b/__init__.py
@@ -7,7 +7,6 @@ from openfecwebapp.views import (render_search_results, render_table,
 from openfecwebapp.api_caller import (load_search_results,
     load_single_type, load_single_type_summary,
     install_cache)
-from newrelic import agent as nr_agent
 
 import sys
 import locale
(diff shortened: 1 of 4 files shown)


__init__.py
```diff
@@ -84,6 +85,7 @@ def currency_filter(num, grouping=True):
     return locale.currency(num, grouping=grouping)
 
 if __name__ == '__main__':
+    nr_agent.initialize()
```
jmcarp: My impression from the docs is that we can either run our app prefixed with `newrelic-admin run-program` or import and initialize the agent within the app. One or the other should be enough--I don't think we need both.

https://docs.newrelic.com/docs/agents/python-agent/installation-configuration/python-agent-integration

arowla: Good catch, @jmcarp . Yes, let's standardize on the wrapper script and take this import/initialize out.

arowla: cc @LindsayYoung

newrelic.ini
```diff
@@ -0,0 +1,210 @@
...
+
+#
+# The application environments. These are specific settings which
+# override the common environment settings. The settings related to a
+# specific environment will be used when the environment argument to the
+# newrelic.agent.initialize() function has been defined to be either
+# "development", "test", "staging" or "production".
+#
+
+[FEC_ENVIRONMENT:development]
```
jmcarp: Should this be `NEW_RELIC_ENV`?

newrelic.ini
```diff
@@ -0,0 +1,210 @@
...
+# High Security Mode enforces certain security settings, and
+# prevents them from being overridden, so that no sensitive data
+# is sent to New Relic. Enabling High Security Mode means that
+# SSL is turned on, request parameters are not collected, and SQL
+# can not be sent to New Relic in its raw form. To activate High
+# Security Mode, it must be set to 'true' in this local .ini
+# configuration file AND be set to 'true' in the server-side
+# configuration in the New Relic user interface. For details, see
+# https://docs.newrelic.com/docs/subscriptions/high-security
+high_security = false
```
LindsayYoung: We need this to be true, right?

newrelic.ini
arowla: Actually that should all be taken care of in the CF manifests, I think... We can probably just take out this part. At least the newrelic.ini that @dlapiduz submitted for the API had monitor_mode set to true globally.

newrelic.ini
arowla: Right. Changing.

PR #82, opened 2015-04-21
Cloudfoundry Tweaks

arowla:
* Manifests and deploy script lives in the API repo now, so removing the old ones
* Symlinking .gitignore to .cfignore, as recommended in https://docs.cf.18f.us/apps/deployment/


arowla:
Please review/merge, cc @LindsayYoung or @jmcarp .

PR #83, opened 2015-04-22
More responsive components: Entity headers

noahmanger:
Right now just the candidate header will look good and I'll do the committee header after @msecret 's done with that page.


static/styles/_components/_entity.scss
```diff
@@ -75,6 +86,47 @@
...
+  padding: 1rem 0;
+  label {
+    width: 50%;
+    margin-right: 1rem;
+    display: inline-block;
+  }
+
+  select,
+  .chosen-container {
+    width: 50% !important;
```
msecret: I'm a little lost on what this UI looks like, but why are there so many important rules here?

noahmanger: Ugh I know :) It's kind of a temporary hack because of the way the ChosenJS is working. I was having a really hard time getting the size right without it.

noahmanger: Ok, I'll be able to clean this up with Chosenjs gone. I'm going to wait to style until that's all merged though.

noahmanger:
I'm making a few changes on this right now then it will be ready to merge.

PR #85, opened 2015-04-22
Run selenium tests with Travis

jmcarp:
Run selenium tests in multiple browsers on Travis builds through Sauce Labs. See `TESTING.md` for proposed usage.

Note: Travis doesn't run Sauce Connect on pull request builds, so this build is likely going to fail.

PR #86, opened 2015-04-23
Sauce feedback

jmcarp:
Log test suites as passed or failed in Sauce Labs.

* Add minimal Sauce Labs API client
* Notify Sauce Labs of suite success / failure in `tearDownClass`

This is a follow-up to #85, but makes the additional change of switching from nose to py.test as the test runner, so I wanted to submit separately for discussion. Py.test is fully compatible with nose and unittest tests, so all existing tests behave as previously. But it's easier to add hooks and generally customize py.test, which is helpful for integration with Sauce Labs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,26 @@
 python: "3.4"
+
+sudo: false
+
+addons:
+  sauce_connect: true
+
+env:
+  global:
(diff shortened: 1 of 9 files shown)


arowla:
Needs to target `develop`

jmcarp:
Replaced by #110.

PR #90, opened 2015-04-28
Accordion

msecret:


noahmanger:
Looks good to me. We might want to try it with starting them all out collapsed, but we can make that change later if we want.

PR #97, opened 2015-04-28
Vendor jquery, eliminate a third party ping

konklone:
This increases visitor privacy, and reduces the site's attack surface, by moving jQuery into the repository instead of hitting `ajax.googleapis.com`.


konklone:
Also, I did set up the dev server locally and test this before submitting -- the setup instructions were clear and worked like a charm!

jmcarp:
Is it preferable to vendorize instead of adding to `package.json`?

konklone:
You'd know better than me -- I'm not as familiar with that approach, but if that's how you're handling vendor dependencies, it shouldn't impact the issues above.

PR #99, opened 2015-04-28
Removing tablesorting

arowla:
From @noahmanger :
* Removing the tablesort plugin because we're going to reimplement eventually once you can sort all results, not just those on the page
* Removing old chosenjs files because I forgot to earlier

Addresses 18F/openFEC#625


arowla:
Originally #98, but this new one is for merging into `develop` for our changeover to git-flow.

PR #100, opened 2015-04-28
Feature/browserify ify

jmcarp:
Solves the same problem as #97, but moves more dependencies (jquery and typeahead) into browserify and out of source control / CDNs. I tested this manually and everything looks fine to me, but could use review from @msecret or @noahmanger.


static/js/modules/accordion.js
```diff
@@ -119,4 +121,4 @@ var accordion = {
   }
 };
 
-module.exports = accordion;
\ No newline at end of file
+module.exports = accordion;
```
msecret: Why was the export removed?

jmcarp: I'm a little confused about this one, but it looks like GitHub is displaying the diff wrong. I didn't delete that line, but it's showing up as deleted anyway. Check out the modified file at https://github.com/jmcarp/openFEC-web-app/blob/feature/browserify-ify/static/js/modules/accordion.js. It has that line intact.

static/js/modules/typeahead.js
```diff
@@ -1,5 +1,11 @@
 'use strict';
+
+/* global require, Bloodhound */
+
+var $ = require('jquery');
+require('typeahead.js/dist/typeahead.bundle.js');
```
msecret: Just `require('typeahead');` doesn't work? I was reading in various issues that it should potentially.

jmcarp: You're right--will change this.

PR #104, opened 2015-04-29
Unembedded resources (from API refactor)

arowla:
Known issues:
- Candidate pages may show multiple primary committees
- Financials (totals/reports) are not ready


openfecwebapp/data_prep/committees.py
```diff
@@ -1,37 +1,22 @@
...
-    cands = []
-    for cand in c['candidates']:
-        candidate = {}
-        candidate['candidate_id'] = cand['candidate_id']
-        candidate['name'] = cand['candidate_name']
-        if cand['candidate_id'] not in cands:
-            committee['candidates'].append(candidate)
-            # we should get rid of this after the refactor
-            cands.append(cand['candidate_id'])
+    url = fields.Function(lambda x: url_for('committee_page', c_id=x['committee_id']))
```
jmcarp: If we use flask-marshmallow, this can be slightly cleaner--something like
```python
url = fields.URLFor('committee_page', c_id='<committee_id>')
```

openfecwebapp/data_prep/financial_summaries.py
```diff
@@ -1,11 +1,43 @@
...
+    total_disbursements = Currency(attribute='disbursements')
+    total_cash = Currency(attribute='cash_on_hand_end_period')
+    total_debt = Currency(attribute='debts_owed_by_committee')
+    report_year = fields.Integer()
+    years_totals = fields.Method("format_years_totals")
+
+    def format_years_totals(self, obj):
+        return "{} - {}".format(obj.cycle - 1, obj.cycle)
+
+    @pre_load
```
jmcarp: This is nitpicky, but we could skip the aliasing and temporary variable here:
```python
def map_totals_and_reports(self, data, many):
    return {
        'reports': data['reports'][0],
        'totals': data['totals'][0],
    }
```

openfecwebapp/data_prep/financial_summaries.py
```diff
@@ -1,11 +1,43 @@
...
 
+class Currency(fields.Decimal):
+    def _serialize(self, value, attr, obj):
+        if value is None:
+            return fields.Decimal('0')
+
+        return locale.currency(value, grouping=True)
+
+class CommitteeFinancials(Schema):
+    total_receipts = Currency(attribute='receipts')
```
jmcarp: Possibly beyond the scope of this PR, but I'm not sure why we're going to the effort of remapping these keys. If `total_cash` is a better name than `cash_on_hand_end_period`, why not have the API call this field `total_cash`? I really like the direction this patch is taking, but it could be even simpler if we could harmonize variable names across repos.

openfecwebapp/api_caller.py
```diff
@@ -9,11 +9,13 @@
 MAX_FINANCIALS_COUNT = 4
 
 
-def _call_api(path, filters):
+def _call_api(path, filters={}):
```
jmcarp: It's unsafe to use mutable values as defaults, since one function call can change the value that the next call sees. How about a signature of `def _call_api(path, **filters)`?

openfecwebapp/models/candidates.py
```diff
@@ -0,0 +1,24 @@
...
+def handle_errors(schema, errors, obj):
+    raise ValueError('An error occurred:\n{} \n\n...while marshalling the data:\n{}\n\n...for schema: {}'.format(errors, obj, schema))
+
+
+class CandidateSchema(Schema):
+    """ Defines a schema for candidate.
+    Only fields which need special treatment (renaming, etc.)
+    need be defined separately. All others may go in Meta.fields.
+    """
+    incumbent_challenge = fields.Str(allow_none=True)
```
jmcarp: The `allow_none` option only has an effect for deserializing, which I don't think we're doing. Can these fields move into `Meta#fields` below?

openfecwebapp/models/committees.py
```diff
@@ -0,0 +1,29 @@
...
+        additional = (
+            'committee_id', 'name', 'committee_type', 'committee_type_full', 
+            'treasurer_name', 'designation', 'designation_full',
+            'treasurer_name', 'street_1', 'city', 'state', 'zip'
+        )
+
+
+@CommitteeSchema.error_handler
+def handle_committee_errors(schema, errors, obj):
+    raise ValueError('An error occurred:\n{} \n\n...while marshalling the data:\n{}\n\n...for schema: {}'.format(errors, obj, schema))
```
jmcarp: Not sure what raising a `ValueError` instead of the default `ValidationError` buys us. Both will raise a 500 in the app and throw a reasonable-looking traceback. If we want to throw a non-500 error on invalid data, we could call `abort` here instead.

openfecwebapp/models/committees.py
```diff
@@ -0,0 +1,29 @@
+from marshmallow import Schema, fields
+from flask import url_for
+
+
+class CommitteeSchema(Schema):
```
jmcarp: Again, I like this way better than the code it's replacing, but do we even need a schema here? We're just passing most fields through unaltered, and IMO the fields that we do change aren't essential. For example, why not pass through `designation` exactly as we get it from the API and use code like `{% if committee.designation == 'P' %}` in the templates?

openfecwebapp/models/committees.py
arowla: Originally I wasn't using `strict`, and it was failing silently. But this
probably isn't needed now that `strict` is turned on.

On Tue, May 5, 2015 at 5:10 PM, Joshua Carp <notifications@github.com>
wrote:

> In openfecwebapp/models/committees.py
> <https://github.com/18F/openFEC-web-app/pull/104#discussion_r29713320>:
>
> > +    organization_type = fields.Str(allow_none=True)
> > +    organization_type_full = fields.Str(allow_none=True)
> > +    party = fields.Str(allow_none=True)
> > +    party_full = fields.Str(allow_none=True)
> > +
> > +    class Meta:
> > +        additional = (
> > +            'committee_id', 'name', 'committee_type', 'committee_type_full',
> > +            'treasurer_name', 'designation', 'designation_full',
> > +            'treasurer_name', 'street_1', 'city', 'state', 'zip'
> > +        )
> > +
> > +
> > +@CommitteeSchema.error_handler
> > +def handle_committee_errors(schema, errors, obj):
> > +    raise ValueError('An error occurred:\n{} \n\n...while marshalling the data:\n{}\n\n...for schema: {}'.format(errors, obj, schema))
>
> Not sure what raising a ValueError instead of the default ValidationError
> buys us. Both will raise a 500 in the app and throw a reasonable-looking
> traceback. If we want to throw a non-500 error on invalid data, we could
> call abort here instead.
>
> —
> <https://github.com/18F/openFEC-web-app/pull/104/files#r29713320>.
>



-- 
Alison Rowland
Technical Lead, https://fbopen.gsa.gov | 18F
202-317-0124

openfecwebapp/views.py
```diff
@@ -46,19 +37,75 @@ def render_table(data_type, results, params):
...
+    'candidate': lambda x: CandidateSchema().dump(x).data,
     'committees': lambda x: x,
-    'committee': map_committee_page_values
+    'committee': lambda x: CommitteeSchema().dump(x).data
 }
 
-def render_page(data_type, c_data):
-    # not handling error at api module because sometimes its ok to 
+
+def render_page(data_type, *args, **kwargs):
```
jmcarp: We accept `*args` here, but only ever use `args[0]`. The `**kwargs` splat also seems to accept `candidates` and `committees` keys. A clearer signature might be `def render_page(data_type, data, committees=None, candidates=None)`.

More broadly, the two branches this function can take don't share a lot of code, and putting them into a single function is IMO harder to read / test / debug than using two separate functions, `render_committee` and `render_candidate`.

openfecwebapp/views.py
```diff
@@ -46,19 +37,75 @@ def render_table(data_type, results, params):
...
+        candidate_schema = CandidateSchema()
+        # candidate fields will be top-level in the template
+        tmpl_vars = candidate_schema.dump(data).data
+        tmpl_vars['election_years'] = load_election_years(data['candidate_id'])
+
+        # process related committees
+        committee_fields = ['committee_id', 'name', 'designation_full', 'designation',
+                            'committee_type_full', 'committee_type', 'url',
+                            'is_primary', 'is_authorized', 'is_joint', 'is_leadership']
+        committees = CommitteeSchema(only=committee_fields, many=True, skip_missing=True, strict=True) \
```
jmcarp: Why use `only` here? This data is getting sent to jinja for template rendering, but it's not going to be sent out in the response--is there any cost to sending jinja the entire dictionary?

templates/partials/committee-charts.html
```diff
@@ -1,3 +1,6 @@
+{% with %}
```
jmcarp: This `with` block breaks the page--`reports` and `totals` are added directly to `tmpl_vars` in `views.py`, lines 98-99. They aren't attached to a `financials` dict, so this lookup fails. The page seems to work fine if we delete the `with`, `set`, and `endwith` statements.

arowla: This partial is included in both candidate and committee pages-- it works for one, but not the other. The two need to be reconciled. I don't think just removing the `with` will do it. (I don't have the answer yet, though.)

jmcarp: Ah, you're right. Part of the problem is that we're using different data structures for the candidate and committee data--there's an extra `financials` key in the committee data when it's fetched as part of the candidate endpoint
* Fetch a committee directly: `committee.reports`
* Fetch a committee nested within a candidate: `committee.financials.reports`

I think if we use consistent data structures for nesting financial statements within committees, and reintroduce the `{% with committee = context() %} in `committees-single.html` above, this template would work in both uses.

arowla: Fixed in 07d531c

jmcarp: :+1:

openfecwebapp/views.py
```diff
@@ -33,32 +22,65 @@ def render_table(data_type, results, params):
...
+        elif committee['designation'] == 'D':  # Leadership
+            tmpl_vars['has_leadership_cmtes'] = True
+
+    tmpl_vars['committees'] = committees
+    from marshmallow import pprint; pprint(tmpl_vars)
+
+    return render_template('candidates-single.html', **tmpl_vars)
+
+
+def get_first_result_or_raise_500(data):
```
jmcarp: This behavior predates this patch, so we can defer discussion until later, but I'm not sure we should be throwing a 500 here. This behavior doesn't give much feedback to the user or to the developer, and it can potentially give incorrect feedback: for example, browsing to `/candidate/fakecandidate` should throw a 404, but this logic will throw a 500 instead.

openfecwebapp/views.py
```diff
@@ -33,32 +22,65 @@ def render_table(data_type, results, params):
...
+            tmpl_vars['has_authorized_cmtes'] = True
+
+        elif committee['designation'] == 'J':  # (J)oint
+            tmpl_vars['has_joint_cmtes'] = True
+
+        elif committee['designation'] == 'D':  # Leadership
+            tmpl_vars['has_leadership_cmtes'] = True
+
+    tmpl_vars['committees'] = committees
+    from marshmallow import pprint; pprint(tmpl_vars)
```
jmcarp: It looks like we're now only using marshmallow for its `pprint` helper, which only has a benefit over the standard library `pprint` for `OrderedDict`s, which we also aren't using. I think we can use the standard library and remove marshmallow from the requirements (even though I generally want to use it at every opportunity).

openfecwebapp/api_caller.py
```diff
@@ -9,11 +9,16 @@
...
-def _call_api(path, filters):
+def _call_api(*path_parts, **filters):
     if api_key:
         filters['api_key'] = api_key
-    path = os.path.join(api_version, path.strip('/'))
+
+    path = os.path.join(api_version, *[x.strip('/') for x in path_parts])
     url = parse.urljoin(api_location, path)
+
+    print(url)
```
jmcarp: Prefer to remove print statements before merge. I think using the `logging` module could be helpful here. That way we can write something like `logger.debug(url)`, and set the log level to `DEBUG` for development, and maybe `WARN` in production.

jmcarp:
Added a few very minor comments inline, but LGTM :+1:
Also, I ran the Selenium tests locally, and the only failures I noticed were issues that should be fixed after merging with develop and/or merging https://github.com/18F/openFEC/pull/719 and https://github.com/18F/openFEC-web-app/pull/156.

openfecwebapp/views.py
arowla: Agreed, let's talk about this at some point.

PR #106, opened 2015-04-29
Use consistent limit for reports and totals charts.

jmcarp:
[Resolves https://github.com/18F/openFEC/issues/641]

@LindsayYoung I just saw that you marked this as not a problem, but it's a tiny change anyway. We can close the PR if it's better to wait for quarterly results.

### DIFF
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -4,6 +4,9 @@ from urllib.parse import urlencode
 from openfecwebapp.config import api_location, api_key
 
 
+MAX_FINANCIALS_COUNT = 4
+
+
 def _call_api(path, filters):
     if api_key:
         filters['api_key'] = api_key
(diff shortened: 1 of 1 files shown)


LindsayYoung:
If you already started, you can go for it. We will need it eventually. 

When I said it was not a problem, I just meant it is not as important as the things that are broken or wrong.

Also what would need to scale the charts to work for a wide range of committees. Presidential are going to raise way more than a random committee. We want a scale to look good on both.

noahmanger:
Yeah I think this is good to include now fwiw.

noahmanger:
I'm running this locally and not seeing the results limited. 
screen shot 2015-04-29 at 11 31 33 am

jmcarp:
@noahmanger what are you using for `FEC_WEB_API_URL`? I'm using https://api.open.fec.gov, and it seems to be working locally.

PR #109, opened 2015-04-30
Feature/responsive entities

noahmanger:
Here's a number of changes that make _most_ of the entity pages responsive. 

The charts still aren't there, but the rest are looking good to my eye.

Addresses https://github.com/18F/openFEC/issues/559

### DIFF
--- a/static/styles/_base/_tables.scss
+++ b/static/styles/_base/_tables.scss
@@ -110,7 +110,6 @@ td {
 
 .table__row--note {
   @include font-size(1.0);
-  padding: 0 .5rem;
   width: 100%;
   padding: .5rem 0 1rem 0;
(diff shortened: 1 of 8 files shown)


static/styles/_base/_tables.scss
```diff
@@ -109,12 +109,17 @@ td {
 }    
 
 .table__row--note {
-  @include font-size(1.4);
-  @include rem(padding, 0 .5rem);
+  @include font-size(1.0);
+  padding: 0 .5rem;
```
msecret: I think this padding should be removed as there is one below.

noahmanger: Oops. Fixed.

PR #110, opened 2015-04-30
Feature/travis sauce

jmcarp:
Replacement for #85 and #86.


.travis.yml
```diff
@@ -6,10 +6,32 @@ python:
...
-    - cd openfecwebapp
-install:
-    - pip install -r requirements.txt
+  # Install web app and run in background
+  - pip install -r requirements.txt
+  - npm install
+  - npm install -g browserify
+  - npm run build
+  - python __init__.py &
+  - if [[ $TRAVIS_BRANCH = 'master' ]]; then export FEC_WEB_API_URL="https://api.open.fec.gov"; fi
```
arowla: Not going to hold up the merge for this, but it might make sense to run off of staging instead of production in this case.

PR #111, opened 2015-04-30
Adding the support email address to the "contact support" buttons

noahmanger:
Addresses https://github.com/18F/openFEC/issues/652


jmcarp:
Very minor, but could we make the contact email a global setting instead of pasting it into individual templates? Would make it easier to change in the future. For example `app.jinja_env.globals['contact_email'] = '18F-FEC@gsa.gov'`.

noahmanger:
Great idea. Does that just go in __init__.py?

jmcarp:
I think so--that's the only place we have template globals so far.

PR #116, opened 2015-05-01
Making browse pages responsive

noahmanger:
Ok, this is looking pretty good to me, but want feedback.

The major changes:
- Made the results table and filters each full width at a certain screen size, so they stack
- Hooked up a simple js plugin so that on small screen sizes the first column of a table is fixed while the others scroll underneath
- Changed the layout slightly of the page-results / pagination controls and the filter toggle
- Added a message if no candidate or committee results are returned

screen shot 2015-05-01 at 4 37 57 pm

screen shot 2015-05-01 at 4 38 08 pm

screen shot 2015-05-01 at 4 38 21 pm


static/js/vendor/responsive-tables.js
```diff
@@ -0,0 +1,67 @@
...
+		original.wrap("<div class='scrollable' />");
+
+    setCellHeights(original, copy);
+	}
+	
+	function unsplitTable(original) {
+    original.closest(".table-wrapper").find(".pinned").remove();
+    original.unwrap();
+    original.unwrap();
+	}
```
msecret: The tabbing this code is off. I think it should be 2 spaces per tab and `{` on the same line as the function statement.

static/js/vendor/responsive-tables.js
```diff
@@ -0,0 +1,67 @@
...
+    }
+    else if (switched && ($(window).width() > 767)) {
+      switched = false;
+      $("table.responsive").each(function(i, element) {
+        unsplitTable($(element));
+      });
+    }
+  };
+   
+  $(window).load(updateTables);
```
msecret: we're already in document.ready, why doesn't this need to be called in updateTables?

static/js/vendor/responsive-tables.js
```diff
@@ -0,0 +1,67 @@
...
+  $(window).on("redraw",function(){switched=false;updateTables();}); // An event to listen for
```
msecret: Will this work in all browsers? I actually can't really find any documentation on it.

static/js/vendor/responsive-tables.js
noahmanger: This is just from a vendor, I didn't write it.

static/js/vendor/responsive-tables.js
noahmanger: Here's what they say on their site:
> Device Compatibility

This technique was tested on iOS, Android and Windows Phone 7 (as well as Chrome, Safari and Firefox). One caveat is that because Android 2.3 doesn't support scrollable divs on the page, the scrolling won't work on those devices. Android 3 and up does support this, so it's not an issue on newer Android phones.

http://zurb.com/playground/responsive-tables

msecret:
The only thing I see is the page navigation at the tops looks a little weird at medium and small breakpoints when you move past the first page. I'm still fine with merging at this point though and filing a separate issue for it because it's relatively small.

medium is strange because it overlaps and is pushing down the hide button.
screen shot 2015-05-05 at 10 15 00 am

small I feel like the next button is off center, which makes the navigation confusing.
screen shot 2015-05-05 at 10 15 57 am

noahmanger:
Ah yeah. Good eye. I'll fix that real quick.

On Tue, May 5, 2015 at 10:20 AM, Marco Segreto <notifications@github.com>
wrote:

> The only thing I see is the page navigation at the tops looks a little
> weird at medium and small breakpoints when you move past the first page.
> I'm still fine with merging at this point though and filing a separate
> issue for it because it's relatively small.
>
> medium is strange because it overlaps and is pushing down the hide button.
> [image: screen shot 2015-05-05 at 10 15 00 am]
> <https://cloud.githubusercontent.com/assets/1701077/7478505/1eb357a6-f310-11e4-91d9-96c0701d126e.png>
>
> small I feel like the next button is off center, which makes the
> navigation confusing.
> [image: screen shot 2015-05-05 at 10 15 57 am]
> <https://cloud.githubusercontent.com/assets/1701077/7478509/2516a792-f310-11e4-81d7-00ab29c4ec20.png>
>
> —
> <https://github.com/18F/openFEC-web-app/pull/116#issuecomment-99147016>.
>



-- 
Noah Manger

PR #117, opened 2015-05-01
Force HTTPS and HSTS when configured to do so

konklone:
This uses [`flask-sslify`](https://github.com/kennethreitz/flask-sslify) to force HTTPS and add HSTS, when an environment variable -- `FEC_FORCE_HTTPS` -- is set.

As configured, this sets HSTS for 1 year, and includes subdomains. It should only be used on domains where you can guarantee that all subdomains are also available over HTTPS. This could be made configurable with additional variables, but it seemed like the smallest amount of overhead.

The HSTS header will only be sent over HTTPS, and will only take effect if the browser successfully connects over HTTPS. So there's no risk of bricking `localhost` or whatever.

Fixes https://github.com/18F/openFEC/issues/619.


konklone:
This also contains a commit that ignores `.env`, used by one of Kenneth Reitz's _other_ popular tools, [`autoenv`](https://github.com/kennethreitz/autoenv).

konklone:
And, I've updated the branch to resolve the merge conflict. This should be good for review and merge any time.

.gitignore
```diff
@@ -8,3 +8,4 @@ static/js/app.js
 openfecwebapp/local_config.py
 static/styles/styles.css.map
 app/
+/.env
```
jmcarp: No complaints here, but kind of thing can also go in a global .gitignore, since not everyone uses autoenv. Even though it's great.

konklone: :+1: Good call for future stuff, I honestly had not thought of that.

PR #121, opened 2015-05-04
[WIP] Reenable cycle select widget on candidate detail page.

jmcarp:
* Remove `disabled` flag from selector
* Fix URL building

Now that https://github.com/18F/openFEC/issues/649 is resolved, we should be able to reenable cycle filtering on the candidate detail page. There are some deeper issues to resolve on the webapp side before this URL will actually work (see https://github.com/18F/openFEC/issues/692), so I'm marking this as WIP for now--not ready to merge!


LindsayYoung:
I don't think this works yet, you need to used candidate/id/history to get time-responsive results

so in api caller we would need something like - 
```
@app.route('/candidate/<c_id>/<cycle>')
def candidate_page_with_cycle(c_id, cycle):
    if c_id != 0:
        data = load_single_type('candidate/{0}/history/{1}'.format(c_id, 'recent'), c_id)
    else:
        data = load_single_type('candidate/{0}/history/{1}'.format(c_id, cycle), c_id)
    return render_page('candidate', data)
```
And then, we would have to look up related committees  using `candidate/id/committees`

LindsayYoung:
This shouldn't be hard to do once we have things un-nested

LindsayYoung:
Also, I am working on the underlying issue of c_id = 0 so that might not be necessary.

LindsayYoung:
This adds back in the year filter dropdown, which is not functioning correctly. It is not accurately filtering the results by year. Are we sure we want to do this rather than hiding the dropdown completely until the functionality is working?- @noahmanger @msecret

jmcarp:
I think we should revert this. I should have been more clear in the description for the PR, but this wasn't ready for merge for the reasons @LindsayYoung mentioned. Agreed that we should hide the dropdown entirely until we have time to fix the underlying issues.

PR #123, opened 2015-05-04
More small cross-browser visual fixes.

msecret:
- Extra line on page subsection on candidate page.
- More accurate alignment on accordion icon.
- Ensure there's always a px fallback for rem.

### DIFF
--- a/static/styles/_base/_forms.scss
+++ b/static/styles/_base/_forms.scss
@@ -49,7 +49,7 @@ select[multiple=multiple] {
   @include box-sizing(border-box);
   @include transition(border-color);
   @include font-size($form-font-size);
-  @include rem(padding, 1rem);
+  @include rem(padding, 0 1rem);
   margin: 1em 0;
   background-color: white;
   border-radius: $form-border-radius;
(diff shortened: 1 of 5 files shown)


templates/partials/financial-summary.html
```diff
@@ -6,7 +6,7 @@ <h2 id="section-1-header" tabindex="0">Financial Summary</h2>
       </div>
 
       {% if primary_committee or authorized_committees %}
-      <div class="page-subsection">
+      <div class="page-subsection page-subsection--noline">
```
noahmanger: I actually want the line here between the authorized committee section and the joint fundraising committees. Fine to keep the css class, but let's remove it from here.

static/styles/_base/_forms.scss
```diff
@@ -49,13 +49,13 @@ select[multiple=multiple] {
   @include box-sizing(border-box);
   @include transition(border-color);
   @include font-size($form-font-size);
+  @include rem(padding, 1rem);
```
noahmanger: Should actually be `padding: 0 1rem` but I fix this in a forthcoming pr.

templates/partials/financial-summary.html
msecret: So this line is between primary and other committee:

screen shot 2015-05-04 at 2 33 58 pm

Is that what you meant? I wanted to check before I re-added it.

noahmanger: Yep!

PR #125, opened 2015-05-04
Add territories; make it DRY.

jmcarp:
* Add territories to state dropdown
* Move reused dropdown code into partials
* Use loops instead of manually enumerating options
* Make casing consistent in filter bar

[Resolves #683]


templates/candidates.html
```diff
@@ -30,285 +30,12 @@
...
-                <option value="94">94</option>
-                <option value="95">95</option>
-                <option value="96">96</option>
-                <option value="97">97</option>
-                <option value="98">98</option>
-                <option value="99">99</option>
-            </select>
-           <button class="button--remove" data-removes="district"><i class="ti-close"></i></button>
-        </div>
+        {% include 'partials/election-years.html' %}
```
noahmanger: Great idea! Can we prefix these file names with `filters_` or something? Will keep things tidier.

PR #130, opened 2015-05-05
Feature/select styling

noahmanger:
Simple fix for overriding default browser select styling.


static/styles/_base/_forms.scss
```diff
@@ -90,7 +90,24 @@ input[type="file"] {
 select {
   width: auto;
   max-width: 100%;
-  background: #fff;
+  -webkit-appearance: none;
```
msecret: Is there a bourbon mixin to use for this rather then defining manually multiple times?

static/styles/_base/_forms.scss
```diff
@@ -90,7 +90,24 @@ input[type="file"] {
...
+  -moz-appearance: none;
+  background-image: url('../img/angle-down.png'); // Fallback incase no svg background
+  background-image: url('../img/angle-down.svg');
+  background-repeat: no-repeat;
+  background-position: 95% 50%;
+  background-size: 2rem;
```
msecret: Does this need a px fallback? Or maybe not because background-size isn't always implemented?

static/styles/_base/_forms.scss
noahmanger: Oh yep. Added it.

static/styles/_base/_forms.scss
noahmanger: I've just been defining in rems with no fallback under the thought that we'll use a polyfill once we need to support older browsers.

PR #131, opened 2015-05-05
Feature/alpha text

noahmanger:
Hiding the "Learn more" link if the banner is open in order to reduce confusion. Addresses: https://github.com/18F/openFEC/issues/670


static/styles/_components/_states.scss
```diff
@@ -3,6 +3,10 @@
   display: none;
 }
 
+.visible {
```
msecret: Theres a .hidden selector in base/_utilities. Should that and others be moved here to this states file?

PR #132, opened 2015-05-05
Added a skip nav link for accessibility

msecret:
Part of openFEC#659.

### DIFF
--- a/static/styles/_base/_utilities.scss
+++ b/static/styles/_base/_utilities.scss
@@ -21,7 +21,13 @@
 }
 
 .u-hidden {
-	display: none;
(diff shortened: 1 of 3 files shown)


noahmanger:
Do we want this to become visible on `:focus`? Maybe it's not necessary.

msecret:
I haven't before. I'm not sure it matters because if they're tabbing to it on a screen reader, they won't see where it is anyway. Also, it's in a pretty random location on the page now so might be jarring

noahmanger:
Cool. I'm fine as is. I just know some place (like github) do it that way. I'll merge.

PR #133, opened 2015-05-05
Feature/659 various fixes

msecret:
Various fixes for 508, all defined by Nick Bristow here: https://github.com/18F/openFEC/issues/702

- Fixed glossary terms on enter keypress
- Add ids to all filters
- Added aria-described by to glossary

(more details in commit messages)

### DIFF
--- a/static/js/modules/glossary.js
+++ b/static/js/modules/glossary.js
@@ -101,6 +101,7 @@ clearTerm = function() {
 module.exports = {
   init: function(){
     glossaryLink.on('click keypress', function(e){
+        console.log('here1');
         if (e.which === 13 || e.type === 'click') {
             var dataTerm = $(this).data('term'),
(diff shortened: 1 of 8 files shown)


static/js/modules/glossary.js
```diff
@@ -101,6 +101,7 @@ clearTerm = function() {
 module.exports = {
   init: function(){
     glossaryLink.on('click keypress', function(e){
+        console.log('here1');
```
noahmanger: Can we remove `this console.log`?

noahmanger:
Looks good other than that note. There's a separate thing though I kept forgetting about, which is if you're using the glossary search, arrow down to a suggestion and hit enter, it doesn't actually pull up the definition. You could tackle that in this one or we could make a separate issue.

PR #136, opened 2015-05-06
Removing negative z-index

noahmanger:
The chart axes disappeared for some reason. Removing the negative
z-index and putting a positive one on the bars solved it.


msecret:
So what we should really do with z-indexes, it set them up in the variables file. I've done it on peace corps. Something like:

@z-index-1: 100;
@z-index-2: 200;

@background: @z-index-1;
@foregroud: @z-index-2;

Can we create an issue or TODO for this if you agree?

noahmanger:
Good call. I've been meaning to do that. I'll make an issue.

PR #138, opened 2015-05-06
Fix errors on committee profile pages

LindsayYoung:
#695
 - fixed mislabeled transfers row
 - removed rows that were not producing real data
 - fixed spelling errors
 - Expanded the explanation when no data is shown

### DIFF
--- a/templates/committees-single.html
+++ b/templates/committees-single.html
@@ -140,13 +140,14 @@
           <h2 id="section-1-header" tabindex="0">Financial Summary</h2>
           <p class="text--lead">Get the full picture of all of the money received and spent by this committee.</p>
           <div class="row js-accordion">
+
             <h3 class="section-header">Detailed Summary</h3>
             {% with committee=context() %}
 
             {% set table_data = [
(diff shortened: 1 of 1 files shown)


templates/committees-single.html
```diff
@@ -140,13 +140,14 @@ <h5 class="entity__term__label">Address</h5>
...
           <div class="row js-accordion">
+
             <h3 class="section-header">Detailed Summary</h3>
             {% with committee=context() %}
 
             {% set table_data = [
                  (committee.totals.0.individual_itemized_contributions, 'Itemized Individual Contributions'),
                  (committee.totals.0.individual_unitemized_contributions, 'Unitemized Individual Contributions'),
-                 (committee.totals.0.transfers_to_affiliated_committee, 'Transfer from a Committee'),
+                 (committee.totals.0.transfers_to_affiliated_committee, 'Transfer to a Committee'),
```
arowla: Shouldn't this field go under disbursements, rather than income? (Actually, it is duplicated in the table now.)

In the nesting refactor branch I changed this to `transfers_from_other_authorized_committee`, but we need to find an appropriate analogous field to pull from the PacOrParty data for this.

PR #139, opened 2015-05-06
Restyling accordion buttons

noahmanger:
Now they will look good at all screen sizes.

Fixes: https://github.com/18F/openFEC/issues/713


static/styles/styles.css
```diff
@@ -0,0 +1,2587 @@
+@charset "UTF-8";
```
msecret: I think this file isn't supposed to be checked in anymore. @jmcarp would know

noahmanger: Oh great. Hope so.

jmcarp: Right, we don't need to check in compiled assets (styles.css, app.js) anymore.

noahmanger: Great. I added them to my .gitignore. Ok to still merge this in though?

msecret: @noahmanger I think you have to run this to remove the actually file now that its being ignore:

```
git rm --cached <file>
```

PR #141, opened 2015-05-06
Restore glossary search on typeahead select.

jmcarp:
When the user selects an option in the glossary typeahead, the
associated text should appear below. A previous bug fix prevented this
behavior, such that selecting an option did not show the glossary text.

This also uses the `typeahead:selected` custom event, which will hopefully be more robust and easier to understand than listening for particular keypress buttons. @msecret could you verify that the bug you fixed in d013e08137e7fb22ae9dea2761375dbe9c645130 is still fixed with this patch? I think it is, but I'm not sure exactly how the glossary was behaving before you fixed it.

### DIFF
--- a/static/js/modules/glossary.js
+++ b/static/js/modules/glossary.js
@@ -1,5 +1,7 @@
 'use strict';
 
+/* global require */
+
 var $ = require('jquery');
 var _ = require('underscore');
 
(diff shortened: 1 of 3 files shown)


noahmanger:
I totally just did this randomly here https://github.com/18F/openFEC-web-app/commit/09664f32ada4464dedcf49fdc5e975d4e8694c1e . Happy to go with either method.

noahmanger:
I like you're approach. Can you just change the main typeahead to be bound to the correct search bar and remove the `if ( datasetName === 'Definitions' )`?

jmcarp:
Updated the PR @noahmanger. I moved both `typeahead:selected` listeners into separate functions to (hopefully) make it clearer what each typeahead is listening for. Sorry about the conflict!

PR #145, opened 2015-05-07
Feature/multi select year

jmcarp:
Use multi-select for election cycle picker.

* Incorporate style updates from @noahmanger
* Show default cycles unless otherwise specified
* Convert repeated query parameters to string (for now)
* Simplify query string handling

[Resolves https://github.com/18F/openFEC/issues/626]

This patch also includes some drive-by formatting fixes because jshint keeps yelling at me.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -23,15 +23,23 @@ app = Flask(__name__)
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.INFO)
 
+
 @jinja2.contextfunction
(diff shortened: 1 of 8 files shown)


noahmanger:
Yessss. Well done. Works for me, but did you mean to not include it on committees.html?

jmcarp:
@noahmanger that's weird, it looks like the cycle filter hasn't been on the committees search page for a while. I'll put it back.

PR #149, opened 2015-05-07
[WIP]Feature/candidate dates

LindsayYoung:
This is similar to the committee dates pr.

 I took out the broken select and replaced it with a list of elections. I called them recent elections, because we are planning on limiting the dates that we are showing for the first release. 

I corrected the dates in the committee summaries.

I added a title over the charts to say that we are looking at recent reports, since they don't always apply to the cycle we are displaying.

Feel free to tweak it. I just want people to understand which sections apply to which dates.


noahmanger:
Will this show all elections they've been in or just the most recent?

LindsayYoung:
I think this is just moving to Noah's branch

LindsayYoung:
This is in #163

PR #153, opened 2015-05-09
Add optional hooks for dependency management.

jmcarp:
Add optional post-checkout and post-merge hooks to update dependencies and
compiled assets when underlying dependencies and code change.

After running `invoke add_hooks`, git will automatically check for changes in `requirements.txt`, `package.json`, and all static files on checkout or merge, and install dependencies and/or rebuild compiled assets if the relevant source files are different. The hooks are optional, so nobody would have to use them if they didn't want to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,3 +91,17 @@ If the development API is down or for testing with feature branches of the API,
 a local API server can be used:
 
     $ unset FEC_WEB_API_URL
+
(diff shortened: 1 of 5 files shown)


arowla:
We should have something similar for requirements.txt in the API app too, right, @jmcarp ?

PR #156, opened 2015-05-11
Update with API changes.

jmcarp:
Update queries to match changes to API in https://github.com/18F/openFEC/pull/719.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -29,9 +29,10 @@ def get_context(c):
     return c
 
 
-def _get_default_years():
+def _get_default_cycles():
     now = datetime.datetime.now().year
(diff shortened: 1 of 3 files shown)


msecret:
I tried to test your changes but met up with some errors that probably have more to do with the API not being in sync.

When picking some years in the cycle selector and picking filter I was getting the generic error screen and the API was giving a 500:

```INFO:werkzeug:127.0.0.1 - - [11/May/2015 13:46:12] "GET /candidates?name=&cycle=2016&cycle=2014&cycle=2012&party=&state=&district=&office= HTTP/1.1" 500 -
```

If this is expected at this point, let me know and I'll merge.

PR #159, opened 2015-05-11
Feature/conflict resolution

jmcarp:
Fix miscellaneous issues that came up after merging parallel changes to API and webapp. :see_no_evil:


LindsayYoung:
Peace prize goes to @jmcarp

PR #160, opened 2015-05-11
Bundle vendor assets.

jmcarp:
Prefer to include all assets in bundle for cache-busting (and eventually
minification). And it's nice to have all the assets managed in the same way :smiley:


noahmanger:
:+1:

PR #161, opened 2015-05-12
Fix types of remove buttons.

jmcarp:
By default, `<button>` elements have type `"submit"`, such that pressing
Enter on a form element triggers that button. This means that pressing
Enter when focusing on the candidate or committee name input triggered
the associated remove button, clearing the input. This patch adds the
correct button type of `"button"` to buttons that should not trigger a
form submit.

[Resolves https://github.com/18F/openFEC/issues/669]


noahmanger:
Good lookin' out.

PR #163, opened 2015-05-12
Feature/committee-dates-noah

noahmanger:


__init__.py
```diff
@@ -133,8 +133,21 @@ def currency_filter(num, grouping=True):
 def date_filter_sm(date_str):
     if not date_str:
         return ''
-    return parse_date(date_str).strftime('%m%y')
+    return parse_date(date_str).strftime('%m/%y')
 
+@app.template_filter('date_md')
```
jmcarp: Both newly added filters need unit tests. Tests can be added to `openfecwebapp/tests/test_utils.py` after #167 is merged.

__init__.py
```diff
@@ -133,8 +133,21 @@ def currency_filter(num, grouping=True):
...
+def date_filter_md(date_str):
+    if not date_str:
+        return ''
+    return parse_date(date_str).strftime('%b %Y')
+
+@app.template_filter('last_3_characters')
+def last_3_characters(int):
```
jmcarp: How about a more concise implementation, e.g.
```python
def last_n_characters(value, nchar=3):
    if type(value) == int:
        return int % (10 ** nchar)
    return ''
```
If `nchar` is parameterized (I defaulted it to `3` above), then we can ask for arbitrary numbers of digits if we ever want to.

static/js/vendor/responsive-tables.js
```diff
@@ -1,4 +1,4 @@
-$(document).ready(function() {
+jQuery(document).ready(function() {
```
msecret: Why was this changed?

noahmanger: Oh, I was confused because jQuery wasn't working but I forgot to run `npm install`. Reverting.

PR #178, opened 2015-05-14
Feature/fix typeahead sort

jmcarp:
Fix grouping in main typeahead search.

The actual issue here turned out to be that we were checking the
candidate and committee IDs against `null` but were getting back a
different falsy value of `''`. This patch uses a more general check,
ignoring results with all falsy ID values, and revises the filtering
logic to use a more explicit filter and map, rather than the implicit
filtering in `$.map`.

[Resolves https://github.com/18F/openFEC/issues/741]

### DIFF
--- a/openfecwebapp/tests/selenium/base_test_class.py
+++ b/openfecwebapp/tests/selenium/base_test_class.py
@@ -1,4 +1,5 @@
 import os
+import logging
 import unittest
 
 import pytest
@@ -9,6 +10,10 @@ from selenium.webdriver.common.keys import Keys
 from selenium.common.exceptions import NoSuchElementException
 
 from openfecwebapp.sauce import SauceClient
+
+# Silence Selenium logs
+remote_logger = logging.getLogger('selenium.webdriver.remote.remote_connection')
+remote_logger.setLevel(logging.WARN)
 
 
 sauce_url = 'http://{0}:{1}@ondemand.saucelabs.com:80/wd/hub'.format(
--- a/static/js/modules/typeahead.js
+++ b/static/js/modules/typeahead.js
@@ -3,6 +3,7 @@
 /* global require, module, window, Bloodhound, API_LOCATION, API_VERSION, API_KEY */
 
 var $ = require('jquery');
+var _ = require('underscore');
 require('typeahead.js');
 var URI = require('URIjs');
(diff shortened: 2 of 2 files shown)


openfecwebapp/tests/selenium/base_test_class.py
```diff
@@ -10,6 +11,10 @@
 
 from openfecwebapp.sauce import SauceClient
 
+# Silence Selenium logs
+remote_logger = logging.getLogger('selenium.webdriver.remote.remote_connection')
+remote_logger.setLevel(logging.WARN)
```
arowla: A++++++++++

noahmanger: Nice. Does this mean we'll just see warnings and failures?

jmcarp: For this logger, we'll only see messaged logged at `WARN`, `ERROR`, or `CRITICAL`. All other loggers are unaffected.

jmcarp:
By the way, in the spirit of disclosure, I now realize that I broke this in the first place on the API side. We're currently using the latest stable version of marshmallow for serialization, and null values are serialized as empty strings. I actually brought this up with the marshmallow developer a few weeks ago, and he fixed that behavior such that nulls serialize to nulls, but I didn't notice that we needed to upgrade to the dev version of marshmallow to get this change. I'll do that later today.

Shorter version: sorry @noahmanger!

noahmanger:
No worries! Thanks for the transparency!

On Thu, May 14, 2015 at 9:38 AM, Joshua Carp <notifications@github.com>
wrote:

> By the way, in the spirit of disclosure, I now realize that I broke this
> in the first place on the API side. We're currently using the latest stable
> version of marshmallow for serialization, and null values are serialized as
> empty strings. I actually brought this up with the marshmallow developer a
> few weeks ago, and he fixed that behavior such that nulls serialize to
> nulls, but I didn't notice that we needed to upgrade to the dev version of
> marshmallow to get this change. I'll do that later today.
>
> Shorter version: sorry @noahmanger <https://github.com/noahmanger>!
>
> —
> <https://github.com/18F/openFEC-web-app/pull/178#issuecomment-102096292>.
>



-- 
Noah Manger

PR #181, opened 2015-05-15
Use election cycles for cycle range in candidates table.

jmcarp:
The `election_years` and `cycles` columns aren't exactly the same. Since
the filter and table header refer to election cycles, we should use the
cycles data.

* Use election cycles rather than years to build election cycles column.
* Add min-max range filter to display cycle ranges.
* Update unit tests.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -135,11 +135,13 @@ def date_filter_sm(date_str):
         return ''
     return parse_date(date_str).strftime('%m/%y')
 
+
 @app.template_filter('date_md')
 def date_filter_md(date_str):
     if not date_str:
         return ''
     return parse_date(date_str).strftime('%b %Y')
+
 
 @app.template_filter('last_n_characters')
 def last_n_characters(value, nchar=3):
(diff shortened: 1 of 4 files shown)


LindsayYoung:
We were going to use election years that for - #744

I do think there should be cycle filtering too. but can we keep election years?

LindsayYoung:
Sorry, is this for filtering or will it give different results?

jmcarp:
Filtering already uses election cycle, and the column in the results table is labeled "Election Cycle", so I was expecting the values to be election cycles. But if you think showing years in that column is clearer, let's go with that.

LindsayYoung:
Let me take a closer look.

LindsayYoung:
I thought this was on the API side and that is why it wasn't making sense. My apologies for the confusion. :+1:

LindsayYoung:
@noahmanger 

I was confused by the label "Election Cycle" for that column in Candidates.

Perhaps we could use the election years and make a column that says "Elections"? Unless, you think that we would be confusing people with a lack of parity in the drop down. 

I think other people might be confused by the "Election Cycle" too because (people use the term differently)[http://www.ncsl.org/blog/2014/06/26/whats-an-election-cycle-depends-where-you-are.aspx], especially among academics. 

I think the label makes sense in a drop-down menu, but I think we might want to be more explicit by showing the ranges "2001-2002". Since currently, it doesn't look like there is a way to see elections that happen in an odd year.

LindsayYoung:
Moved this discussion to issue #771

LindsayYoung:
I talked to Noah about the labeling and I will make a pr for this pr if you don't mind.

jmcarp:
:+1:

arowla:
Is this ready?

jmcarp:
Given changes in #185, I don't think we need this one anymore.

PR #185, opened 2015-05-15
Feature/show years

LindsayYoung:
This works with the dev api. 

closes #771


noahmanger:
So I like this, but there's still some inconsistency because on the candidate-single page, we say "Election Years" but then just list the individual years when there was an election, not the ranges.

I think this should either be changed to show the two-year cycles, _or_ change the label to "Elections".

LindsayYoung:
OK, I'll change the label

PR #187, opened 2015-05-18
Use full-text search for committee and candidate search fields.

jmcarp:
This patch passes the name search value as "q" instead of "name",
meaning that we use the full-text search on candidate and committee
names instead of the stricter `ILIKE` used by "name". Practically, this
means that searching for "nancy pelosi", "pelosi nancy", or "pelosi,
nancy" will all yield Nancy Pelosi, which is what I think users will
expect.

One side effect of this proposal is that searching for a
candidate's FEC ID will also find that candidate, since the full-text
search includes both name and ID. If this is a problem, we can fix it on
the backend, but I don't think it's a cause for concern.

[Resolves https://github.com/18F/openFEC/issues/777]

### DIFF
--- a/static/js/modules/filters.js
+++ b/static/js/modules/filters.js
@@ -69,7 +69,7 @@ var selectedFilters = {};
 
 // all of the filters we use on candidates and committees
 var fieldMap = [
-    'name',
+    'q',
     'cycle',
     'party',
     'state',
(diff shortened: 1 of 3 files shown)


LindsayYoung:
I think having id is better. Power users will want things searchable by ID.

PR #188, opened 2015-05-18
Feature/selenium fix

jmcarp:
* Update Selenium version so that Firefox tests can run locally
* Update cycle test to handle year ranges


jmcarp:
Paging @arowla for review, time permitting.

PR #189, opened 2015-05-18
Point users to the right data and set appropriate expectations

LindsayYoung:
Some things are still wonky so I would rather send users to the real information. I don't think we are at at a point where if we don't show something, it doesn't exist.


jmcarp:
:+1:

PR #191, opened 2015-05-18
Add blue-green deploys.

jmcarp:
Code borrowed from API.

### DIFF
--- a/manifest_prod.yml
+++ b/manifest_prod.yml
@@ -11,7 +11,12 @@ applications:
   buildpack: "https://github.com/ddollar/heroku-buildpack-multi.git"
   host: fec-prod-web
   services:
-    # Must provide FEC_WEB_USERNAME, FEC_WEB_PASSWORD, NEW_RELIC_LICENSE_KEY
+    # Must provide:
+    # * FEC_WEB_USERNAME
+    # * FEC_WEB_PASSWORD
+    # * FEC_WEB_API_KEY
+    # * FEC_WEB_API_KEY_PUBLIC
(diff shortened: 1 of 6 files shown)


jmcarp:
Important: this deploy uses two applications, "web-a" and "web-b". Environment variables must be set for these new apps in order for them to work--otherwise, we'll get errors. @arowla let's talk about how to proceed before deploying.

arowla:
Let's set up the user-provided service. I've seen enough of hand-set ENV vars disappearing already, even without this change.

arowla:
Issue 18F/openFEC#800

arowla:
See 18F/openFEC#796

manifest_prod.yml
```diff
@@ -6,10 +6,25 @@ env:
...
   path: .
   buildpack: "https://github.com/ddollar/heroku-buildpack-multi.git"
   host: fec-prod-web
+  services:
+    - fec-creds-prod
+  env:
+    NEW_RELIC_APP_NAME: OpenFEC Web (production)
+    FEC_WEB_API_URL: https://api.open.fec.gov/
+    FEC_FORCE_HTTPS: true
+    # These must be set manually via `cf set-env`
```
arowla: Would it make sense to note somewhere (or just right here) all the additional varnames we now need to make sure are manually set on the creds service?

```
# FEC_WEB_API_KEY
# FEC_WEB_API_KEY_PUBLIC
# NEW_RELIC_LICENSE_KEY
```

... And any others I might have missed.

jmcarp: :+1:

PR #193, opened 2015-05-21
Hide duplicate dates in charts.

jmcarp:
For some monthly filings, report dates currently display as e.g. "01/15
- 01/15", which can be confusing (why does the date repeat?). This patch
checks for duplicate dates and only shows one date in that case. Logic
is also moved from templates to a filter, which is also tested.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -129,6 +129,29 @@ def currency_filter(num, grouping=True):
         return locale.currency(num, grouping=grouping)
 
 
+def _unique(values):
+    ret = []
+    for value in values:
+        if value not in ret:
+            ret.append(value)
+    return ret
+
+
+def _fmt_chart_tick(value):
+    return parse_date(value).strftime('%m/%y')
+
+
+@app.template_filter('fmt_chart_ticks')
+def fmt_chart_ticks(group, keys):
+    if not group or not keys:
+        return ''
+    if isinstance(keys, (list, tuple)):
+        values = [_fmt_chart_tick(group[key]) for key in keys]
+        values = _unique(values)
+        return ' - '.join(values)
+    return _fmt_chart_tick(group[keys])
+
+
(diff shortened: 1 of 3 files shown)


arowla:
Are we still doing this? Didn't the issue for this get closed?

__init__.py
```diff
@@ -129,6 +129,29 @@ def currency_filter(num, grouping=True):
         return locale.currency(num, grouping=grouping)
 
 
+def _unique(values):
+    ret = []
+    for value in values:
+        if value not in ret:
+            ret.append(value)
+    return ret
```
arowla: Can we just do the below?

    return list(set(values))

__init__.py
```diff
@@ -129,6 +129,29 @@ def currency_filter(num, grouping=True):
...
+
+
+def _fmt_chart_tick(value):
+    return parse_date(value).strftime('%m/%y')
+
+
+@app.template_filter('fmt_chart_ticks')
```
arowla: Nitpick, but we don't need to specify the filter name if it's the same as the method name... Though I could see one arguing for just always specifying it for consistency. What do you think?

__init__.py
jmcarp: No--since python sets are unordered, casting to `set` (and then back to `list`) discards order information, which we want to keep here.

__init__.py
jmcarp: Agreed. I think it would be best to pick a convention and stick to it (most of the time, anyway). Speaking of which, especially as the number of filters grows, it would be nice to keep them in their own module, which would make it less necessary to prefix function names with "filter_". It's a little annoying to do, since the filters have to know about the app, but still manageable.

__init__.py
arowla: :+1:

PR #194, opened 2015-05-21
Changed FEC PDF links to be built by a filter

arowla:
- Added check in template to make sure we don't try building a
  non-existent link

### DIFF
--- a/templates/partials/committee-summary.html
+++ b/templates/partials/committee-summary.html
@@ -28,7 +28,7 @@
     </div>
     <div class="table__row--note">
       {% if report %}
-      Source: <a href="{{ report|url_to_fec_pdf }}">{{report.report_type_full|fmt_report_desc|default("unavailable")}} {{ report.report_year }}</a>
+      Source: <a href="{{ report|url_to_fec_pdf }}">{{report.report_type_full|fmt_report_desc|default("Report")}} {{ report.report_year }}</a>
       {% else %}
       Source: None
       {% endif %}
--- a/templates/partials/committee-summary.html
+++ b/templates/partials/committee-summary.html
@@ -1,5 +1,5 @@
 {% with %}
-{% set reports = committee.reports.0 if committee.reports else {} %}
+{% set report = committee.reports.0 if committee.reports else {} %}
 {% set totals = committee.totals.0 if committee.totals else {} %}
 
 <figure class="table--simple reports-table">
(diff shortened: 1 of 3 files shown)


templates/partials/committee-summary.html
```diff
@@ -20,14 +20,18 @@ <h4 class="figure__title">Totals for {{totals.cycle|fmt_year_range|default("unav
...
     </div>
     <div class="table__row">
       <div class="table__cell"><span class="term" data-term="Debt">Debt</span></div>
-      <div class="table__cell">{{reports.debts_owed_by_committee|currency|default("unavailable")}}</div>
+      <div class="table__cell">{{report.debts_owed_by_committee|currency|default("unavailable")}}</div>
     </div>
     <div class="table__row--note">
-      Source: <a href="http://docquery.fec.gov/pdf/{{ reports.beginning_image_number|last_n_characters }}/{{ reports.beginning_image_number }}/{{ reports.beginning_image_number }}.pdf">{{reports.report_type_full|fmt_report_desc|default("unavailable")}} {{ reports.report_year }}</a>
+      {% if report %}
+      Source: <a href="{{ report|url_to_fec_pdf }}">{{report.report_type_full|fmt_report_desc|default("unavailable")}} {{ report.report_year }}</a>
```
jmcarp: If the report is unavailable, should we use the `<a>` at all? I would think we would just show the text in that case.

jmcarp: Can this code lead to a case where we generate HTML like `<a href="unavailable">...</a>`? Disregard previous comment--this is the question I meant to ask :smiley:

arowla: I checked the behavior on /candidate/H2OH06086... It takes the "else" route and says "Source: None".

jmcarp: Good. But if that's the case, then is the `default("unavailable")` logic on this line unreachable? If so, seems like we don't need it, and if not, this could lead to an edge case where we get "unavailable" in the URL.

arowla: Well in this case, the "unavailable" is only in the case that we don't have a full report description. We _ought_ to have that if we have a report, but I'm not 100% certain if that will always be the case. So what if we replace the default link text "unavailable" with something like `default("Committee Report")`?

templates/partials/committee-summary.html
```diff
@@ -20,14 +20,18 @@ <h4 class="figure__title">Totals for {{totals.cycle|fmt_year_range|default("unav
...
     </div>
     <div class="table__row">
       <div class="table__cell"><span class="term" data-term="Debt">Debt</span></div>
-      <div class="table__cell">{{reports.debts_owed_by_committee|currency|default("unavailable")}}</div>
+      <div class="table__cell">{{report.debts_owed_by_committee|currency|default("unavailable")}}</div>
     </div>
     <div class="table__row--note">
-      Source: <a href="http://docquery.fec.gov/pdf/{{ reports.beginning_image_number|last_n_characters }}/{{ reports.beginning_image_number }}/{{ reports.beginning_image_number }}.pdf">{{reports.report_type_full|fmt_report_desc|default("unavailable")}} {{ reports.report_year }}</a>
+      {% if report %}
+      Source: <a href="{{ report|url_to_fec_pdf }}">{{report.report_type_full|fmt_report_desc|default("Report")}} {{ report.report_year }}</a>
```
arowla: @jmcarp I went ahead and changed the default report text to read just "Report" in the case that we have a report but for some reason the `report_type_full` is not filled in. Does that work, do you think?

jmcarp: :+1:

PR #202, opened 2015-05-22
Feature/ogc copy edits

noahmanger:
Addresses the minor text edits that OGC passed along as well as some of the things FEC staff noted in the spreadsheet.


LindsayYoung:
looks good it looks like the tests are failing just because there are too many builds.

PR #203, opened 2015-05-26
Do HTTPS redirect before basic auth check.

jmcarp:
[Resolves https://github.com/18F/openFEC/issues/831]


konklone:
Nice -- sorry about not catching that.

PR #205, opened 2015-05-26
Feature/better bar labels

noahmanger:
Changed the dates on the bar chart labels to be `dd/mm/yy` to address https://github.com/18F/openFEC/issues/821 . They make so much more sense now.

screen shot 2015-05-26 at 4 40 44 pm

Also restyled the typeahead suggestions to make names easier to read (addresses https://github.com/18F/openFEC/issues/833)

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -162,7 +162,7 @@ def _unique(values):
 
 
 def _fmt_chart_tick(value):
-    return parse_date(value).strftime('%m/%y')
+    return parse_date(value).strftime('%m/%d/%y')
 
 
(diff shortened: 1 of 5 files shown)


msecret:
Just to check: did you look at this on different device sizes like mobile and tablet? I'm worried about the width of them labels now.

noahmanger:
Yep. They break to two lines on smaller screens. Things will get tricky once we have more than 4 groups, but we're ok for now.

noahmanger:
Oops. Didn't update the tests...working on that now.

LindsayYoung:
LGTM

PR #207, opened 2015-05-27
Generate correct history parameter when next cycle is not available.

jmcarp:
We use the `next_cycle` filter to generate the `history` query parameter
for the candidate detail page. The filter is meant to get the next
active cycle greater or equal to the specified cycle. This patch handles
the edge case in which no such cycle is found, in which case we should
simply use the latest active cycle. This fixes some errors on Jeff
Flake's candidate page.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -216,10 +216,14 @@ def restrict_cycles(value):
 
 @app.template_filter()
 def next_cycle(value, cycles):
-    """Get the earliest election cycle greater than or equal to `value`."""
+    """Get the earliest election cycle greater than or equal to `value`. If no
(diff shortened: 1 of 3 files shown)


LindsayYoung:
:+1:

LindsayYoung:
Now 2017-2018 is showing up in the election cycle picker, can we add a filter for that in this function too?

http://localhost:3000/candidate/S2AZ00141?cycle=2012

jmcarp:
@LindsayYoung the 2017-2018 cycle shouldn't be showing up anymore with this patch. At least I don't see it anymore. Let me know if you're still seeing this.

LindsayYoung:
LGTM will merge when tests pass

PR #215, opened 2015-05-29
Show glossary options on focus.

jmcarp:
* Upgrade to latest typeahead.js.
* Show default glossary options on focus.

NB: Still needs work on updating styles for latest typeahead.js.

@noahmanger: Help with styles much appreciated!

### DIFF
--- a/static/styles/_components/_search-bar.scss
+++ b/static/styles/_components/_search-bar.scss
@@ -10,14 +10,14 @@
   label {
     margin-bottom: 30px;
   }
-  
+
   @include media($medium) {
     .input-button-combo {
       height: 6rem;
       font-size: 2.0rem;
     }
   }
-  
+
   @include media($large) {
     @include span-columns(18 of 22);
     @include shift(2 of 22);
@@ -46,7 +46,7 @@
   width: 100%;
 }
 
-.tt-dropdown-menu {
+.tt-menu {
   @include font-size(1.2);
   background: #fff;
   text-align: left;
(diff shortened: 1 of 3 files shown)


static/styles/_components/_search-bar.scss
```diff
@@ -56,36 +56,33 @@
...
-  display: block;
-  color: #fff;
-  padding: 5px 20px;
-  background: $dark-gray;
-}
-
-.tt-suggestions {
-  color: #000;
-}
+// .tt-menu:hover {
```
msecret: Is there a reason this is commented out rather then removed?

static/styles/_components/_search-bar.scss
```diff
@@ -56,36 +56,33 @@
...
+//   display: block;
+//   color: #fff;
+//   padding: 5px 20px;
+//   background: $dark-gray;
+// }
 
 .tt-suggestion {
-  @include clearfix();
+  // @include clearfix();
```
msecret: Same here

jmcarp:
@msecret: This is an unfinished patch (JS updates should be good to go, but styles still need work). It sounds like we might stop using typeaheads for the glossary, but if we stick with them, I was hoping to get some help with styling from you or @noahmanger.

jmcarp:
@noahmanger: What are your thoughts about typeahead compared to something like list.js for the glossary? If we're going to stick with typeahead, showing the first N results and filtering down as the user types, I'll get this patch ready to merge. If we're going to use a different solution, I think we can close this PR.

noahmanger:
Personally, I think I'd rather use Listjs (or something comparable), and
show _all_ results by default in a way that people can scroll through, or
start typing to filter down. Seems like a more natural interaction model
(filtering, not searching), especially with such a finite list of terms.
But open to arguments to the contrary.

On Tue, Jun 2, 2015 at 8:58 AM, Joshua Carp <notifications@github.com>
wrote:

> @noahmanger <https://github.com/noahmanger>: What are your thoughts about
> typeahead compared to something like list.js for the glossary? If we're
> going to stick with typeahead, showing the first N results and filtering
> down as the user types, I'll get this patch ready to merge. If we're going
> to use a different solution, I think we can close this PR.
>
> —
> <https://github.com/18F/openFEC-web-app/pull/215#issuecomment-108000191>.
>



-- 
Noah Manger

jmcarp:
Sounds good to me. Closing this for now. We may still want to upgrade typeahead versions down the line, but I think that can be deferred.

PR #218, opened 2015-05-31
Adding a browser warning for old versions of IE

noahmanger:
Adds a modal alert for old versions of IE.

Work in progress -- needs:
- Copy and links. Anyone have good boilerplate to use?
- Browser test on old IE version to make sure it works.

screen shot 2015-05-30 at 9 16 01 pm


msecret:
@noahmanger can you confirm that this won't break the tests? I'll merge when it's ready

noahmanger:
Oh weird. Looks like it failed because of a bad sauce connection. I restarted. I wouldn't think it would affect the tests at all.

PR #220, opened 2015-06-01
Feature/datatables

jmcarp:
Render candidate and committee tables with DataTables.

* Add sorting on all columns.
* Add more flexible pagination controls.
* Fetch table data via AJAX.

This patch is using 100% DataTables defaults style-wise and will definitely need some review on that front. @noahmanger or @msecret, any suggestions or sub-PRs on style would be much appreciated, since I'm not very familiar with the webapp styles.


jmcarp:
This is looking much better with updates from @noahmanger. Ready for review @msecret.

static/js/modules/tables.js
```diff
@@ -0,0 +1,163 @@
...
+  } else {
+    return '';
+  }
+}
+
+var candidateColumns = [
+  {
+    data: 'name',
+    render: function(data, type, row, meta) {
+      var anchor = $('<a>');
```
msecret: Can we use `$anchor` as the var name to specify that it's a jquery element

static/js/modules/tables.js
```diff
@@ -0,0 +1,163 @@
...
+  });
+  form.submit(function(event) {
+    event.preventDefault();
+    api.ajax.reload();
+  });
+}
+
+module.exports = {
+  init: function() {
+    var table = $('#results');
```
msecret: can we use js-prefixed class names as selectors? So something like `js-results`?

static/styles/_base/_tables.scss
```diff
@@ -13,9 +15,8 @@ table {
   }
 
   th {
-    padding: 1rem;
-    height: 4rem; 
-    border-bottom: 1px solid $light-gray;
+    padding: 1rem !important;
```
msecret: I think this should be restructured so `!important` isn't required, although I think Noah is helping with this.

static/styles/_layout/_data-view.scss
```diff
@@ -68,52 +57,97 @@
...
-  padding: 1rem 0 0 0;
-  text-align: center;
-  border-top: 1px solid $light-gray;
-  
-  @include media($medium) {
-    width: auto;
-    float: right;
-    padding: 0;
-    border-top: 0;
+    .dataTables_length {
```
msecret: This looks like too many levels of nesting.

static/js/modules/tables.js
jmcarp: Your call. I tried not to change any of the markup in this patch--`results` happened to be the ID given to the tables previously. I'd be in favor of something more explicit, like `js-results` or `datatable-results`.

templates/candidates.html
```diff
@@ -43,23 +38,11 @@
...
 </section>
-  <div class="page-controls">
-    {% if pagination %}
-      {% include 'partials/pagination.html' %}
-    {% endif %}
-  </div>
+{% endblock %}
+
+{% block scripts %}
+<link rel="stylesheet" href="//cdn.datatables.net/1.10.7/css/jquery.dataTables.css" />
```
msecret: Can we move this into the main codebase somehow? Is it possible to include it as an npm dependency? Also, should the tables be viewable as fast as possible on the page, or could we move this style sheet to the bottom?

static/js/modules/tables.js
jmcarp: Sure, I'll do a pass through the patch and make sure jquery elements are prefixed with `$`.

templates/candidates.html
jmcarp: Definitely possible to include in the main bundle--I was experimenting with this yesterday, and it works fine if you include the css from the npm install in `styles.scss`. I didn't move the file there because the css uses relative paths for images, and it wasn't immediately clear how best to handle that. I can take another look and make it work. The other issue is that we're using a single css bundle for every page at the moment, but we only need the datatables css for a few pages. Do we want to load the datatables assets on every page in the app?

jmcarp: If we want to bundle all assets in the app but also not load all assets on every page, we should be able to handle it with the factor-bundle plugin for browserify.

static/styles/_base/_tables.scss
noahmanger: Ugh I know. These table styles are all to override the datatables css, which is hyper-specific and includes lots of `!important`s . Is there a smarter way to modify the styles of a node module that I'm missing?

static/styles/_layout/_data-view.scss
noahmanger: Same thing as my comment above.

static/styles/_base/_tables.scss
noahmanger: Ooooh, well I now see that we're not even using the npm module yet. So let's refactor this whenever we figure out how we're loading the built in styles.

noahmanger:
Also, this seems to be causing some troubles with the way the tables adjust to smaller screens. I think we may want to look into using the datatables responsive extension: https://www.datatables.net/extensions/responsive/

jmcarp:
@noahmanger @msecret: updated to include datatables css in bundle.

noahmanger:
I'd merge this but there's some conflicts in view.py and I'm not sure what to do there.

jmcarp:
I'll fix conflicts tonight.

jmcarp:
Conflicts fixed @noahmanger. One test is still expected to fail--I submitted https://github.com/18F/openFEC-web-app/pull/237 to fix that.

PR #222, opened 2015-06-02
Added a results check to the load_nested_type.

LindsayYoung:
Committees might not have f3 reports so there won't be results for them.
This keep the pages form breaking.
I think it fixes the remaining openfec#879 bugs

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -128,7 +128,7 @@ def candidate_page(c_id, cycle=None, history=None):
     data = load_single_type('candidate', c_id, *path)
     cycle = cycle or min(current_cycle(), max(data['results'][0]['cycles']))
     path = ('history', str(cycle))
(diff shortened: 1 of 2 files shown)


jmcarp:
Even for API responses with no results, we return `"results": []`, so the `if 'results` in data` check should *always* be true, and the `else` block should never run. So I'm not sure how this patch could change the behavior of the app. Are there pages that work after the patch that didn't before?

LindsayYoung:
This fixed a lot of results for me locally. But that is a good point, it should still have results even if there are not results.

LindsayYoung:
I am closing the request, because I am thinking that the problem is actually something different.

Though, I am still having the problem locally with a lot of committees. For example: 
http://localhost:3000/committee/C00374058?cycle=2014

That is running with the dev api on sample data. It looks ok on the dev server.

PR #223, opened 2015-06-02
Getting pdf urls from the API.

LindsayYoung:
Removing the template filter for building the pdf links.

### DIFF
--- a/templates/partials/committee-summary.html
+++ b/templates/partials/committee-summary.html
@@ -29,8 +29,10 @@
     <div class="table__row--note">
       {% if report.pdf_url %}
         Source: <a href="{{ report.pdf_url }}">{{report.report_type_full|fmt_report_desc|default("Report")}} {{ report.report_year }}</a>
+      {% elif report %}
+        Source: {{report.report_type_full|fmt_report_desc|default("Report")}} {{ report.report_year }}
       {% else %}
-        Source: None
+        Source: Unknown
       {% endif %}
     </div>
   </div>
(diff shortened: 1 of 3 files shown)


LindsayYoung:
will close 18F/openFEC#870 18F/openFEC#888

templates/partials/committee-summary.html
```diff
@@ -27,10 +27,10 @@ <h4 class="figure__title">Totals for {{totals.cycle|fmt_year_range|default("unav
       <div class="table__cell">{{report.debts_owed_by_committee|currency|default("unavailable")}}</div>
     </div>
     <div class="table__row--note">
-      {% if report %}
-      Source: <a href="{{ report|url_to_fec_pdf }}">{{report.report_type_full|fmt_report_desc|default("Report")}} {{ report.report_year }}</a>
+      {% if report.pdf_url %}
+        Source: <a href="{{ report.pdf_url }}">{{report.report_type_full|fmt_report_desc|default("Report")}} {{ report.report_year }}</a>
       {% else %}
-      Source: None
+        Source: None
```
jmcarp: What do you think about giving a little more information here? The source isn't actually "None"--it's just that we don't have access to the original document. How about including the report type, but indicating that the source document isn't currently available?

```
Source: 
{% if report.pdf_url %}
<a href="{{ report.pdf_url }}">{{report.report_type_full|fmt_report_desc|default("Report")}} {{ report.report_year }}</a>
{% else %}
{{report.report_type_full|fmt_report_desc|default("Report")}} [Not available]
{% endif %}
```

PR #225, opened 2015-06-02
Feature/flexbox fix

noahmanger:
I discovered some problems with the way the search field was rendering on IE 10 (and a couple other small inconsistencies). This should fix it all.

### DIFF
--- a/static/js/modules/typeahead.js
+++ b/static/js/modules/typeahead.js
@@ -116,8 +116,7 @@ module.exports = {
       window.location = window.location.origin + '/' + datasetName + '/' + datum.id;
     });
 
-    $('.twitter-typeahead').addClass('flex-container');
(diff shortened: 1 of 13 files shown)


static/js/modules/typeahead.js
```diff
@@ -116,9 +116,7 @@ module.exports = {
       window.location = window.location.origin + '/' + datasetName + '/' + datum.id;
     });
 
-    $('.twitter-typeahead').css({
-        display: 'flex',
-    });
+    $('.twitter-typeahead').addClass('flex-container').css('display','');
```
noahmanger: FYI need to a class to take advantage of Bourbon's browser prefixing for the `display` property. And then the css method is to override the fact that the library sets an inline style of `display: inline-block`.

PR #231, opened 2015-06-02
Feature/glossary improvements

noahmanger:
This replaces Typeahead in the glossary with [ListJS](http://www.listjs.com/) for search/filtering. 

screen shot 2015-06-02 at 4 23 36 pm

It's *almost* there, but needs some finessing. The way it does search right now is to search through both the terms and definitions for any instance of the word, which means that search results can be a little confusing:

screen shot 2015-06-02 at 4 23 48 pm

Also, when you click on a term, like "committee" it just runs a search for that term, returning multiple results.

Would be nice if someone could take a look and make any modifications.


noahmanger:
Oh, and the tests need to be updated. Would love a quick lesson on these so I can get better at maintining them.

static/js/modules/terms.json
```diff
@@ -1,74 +1,78 @@
...
-    "term": "Principal Campaign Committee",
-    "definition": "An authorized committee designated by a candidate as the principal committee to raise contributions and make expenditures for his or her campaign for a federal office."
+    "glossary-term": "Principal Campaign Committee",
+    "glossary-definition": "An authorized committee designated by a candidate as the principal committee to raise contributions and make expenditures for his or her campaign for a federal office."
   },
   {
-    "term":"Candidate",
-    "definition":"An individual seeking nomination for election, or election, to a federal office becomes a candidate when he or she (or agents working on his or her behalf) raises contributions or makes expenditures that exceed $5,000."
+    "glossary-term":"Candidate",
+    "glossary-definition":"An individual seeking nomination for election, or election, to a federal office becomes a candidate when he or she (or agents working on his or her behalf) raises contributions or makes expenditures that exceed $5,000."
```
LindsayYoung: This description still doesn't make sense.

PR #232, opened 2015-06-03
Feature/glossary improvements

jmcarp:
A few proposed changes to the new glossary:

* When the user clicks a term on the page, show *only* that term in the glossary
* When the user clicks a term on the page, expand the term's definition
* Install list.js with npm
* Fix merge conflicts

@noahmanger: I'm happy to update the tests and/or talk through how they work. I can also work on filtering items by title and not by contents (agreed that this is potentially confusing) if you think that's the correct behavior.


noahmanger:
Great! Those changes def help. I'm really not sure if it's better to just search the term or also the definition. How much work do you think that'd take? 

Also, I'm pretty slammed today so if you're able to just update the tests that'd be great. I'll look at what you do and ask any questions.

jmcarp:
@noahmanger I'll update the tests and look into modifying the search today.

PR #236, opened 2015-06-03
Feature/glossary improvements

jmcarp:
* Only search on glossary term names
* Fix selenium tests


noahmanger:
Tests are failing, but UI-wise it works for me great.

PR #237, opened 2015-06-03
Move HTML markup for "None" to macro.

jmcarp:
Putting HTML in a string literal in the `currency` filter was breaking
tests. It's also not ideal, since it requires using the `safe` filter,
and unsafe content could potentially leak through. This patch restores
the behavior of the `currency` filter and moves the markup for the "None"
element to a macro. This also applies the new `null` macro to other uses
of the `currency` filter.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -176,8 +176,7 @@ def server_error(e):
 def currency_filter(num, grouping=True):
     if isinstance(num, (int, float)):
         return locale.currency(num, grouping=grouping)
-    else:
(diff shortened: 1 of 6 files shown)


noahmanger:
Ah good lookin' out. Sorry, I should've just asked the best way to do that. 

Did you not check in the null macro though? I don't see it.

jmcarp:
Derp, forgot to check in the new file. Fixed.

PR #242, opened 2015-06-04
Feature/perf bar

msecret:


templates/partials/hero.html
```diff
@@ -14,5 +14,8 @@ <h1 id="hero-heading" class="text--display">Explore Campaign Finance Data</h1>
           </div>
       </form>
     </div>
+  <script type="application/javascript">
```
jmcarp: Should this be wrapped in `{% if debug %}...{% endif %}`?

templates/layouts/main.html
```diff
@@ -15,9 +15,23 @@
...
     {% if api_location %}
     <script>
-        API_LOCATION = "{{api_location}}"
-        API_VERSION = "{{api_version}}"
-        API_KEY = "{{api_key}}"
+        API_LOCATION = "{{api_location}}";
+        API_VERSION = "{{api_version}}";
+        API_KEY = "{{api_key}}";
+
+        function trackMetric(name, imagesSelector) {
```
jmcarp: Ditto below: Only for development mode, right?

templates/partials/hero.html
msecret: I actually want these performance marks to be present all the time. I can explain more in a meeting, but essentially these are custom metrics that we could track for various pages. If we keep these marks, it means that performance tracking tools like new relic, or google analytics will automatically pick them up.

templates/layouts/main.html
msecret: Same as above

templates/partials/hero.html
jmcarp: Oh, that's cool. Didn't know `window.performance.mark` was a thing!

__init__.py
```diff
@@ -54,6 +54,7 @@ def _get_default_cycles():
 app.jinja_env.globals['api_location'] = config.api_location
 app.jinja_env.globals['api_version'] = config.api_version
 app.jinja_env.globals['api_key'] = config.api_key_public
+app.jinja_env.globals['debug'] = config.debug
```
jmcarp: Is this used in the jinja templates?

msecret: No, would it be useful at all?

jmcarp: Doesn't seem like we have a use for it right now. Might as well take it out of the jinja globals until we need it.

jmcarp:
@msecret do we have to do anything to get the extra performance metrics into New Relic?

PR #244, opened 2015-06-08
Principal committees are in committees_authorized so they weren't showing

LindsayYoung:
I'm bringing Principal Committees back.

fixes 18F/openFEC#917

### DIFF
--- a/templates/candidates-single.html
+++ b/templates/candidates-single.html
@@ -35,7 +35,7 @@
             <h5 class="entity__term__label">
               <a class="term" data-term="Principal Campaign Committee">Principal Committee</a>
             </h5>
-            {% for c in committees if c.designation == 'P' %}
+            {% for c in committees_authorized if c.designation == 'P' %}
             <p class="entity__term__data">
               <a href="{{ url_for('committee_page', c_id=c.committee_id, cycle=cycle) }}">{{ c.name }}</a>
             </p>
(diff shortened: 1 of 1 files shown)


templates/candidates-single.html
```diff
@@ -35,7 +35,7 @@ <h5 class="entity__term__label">District</h5>
             <h5 class="entity__term__label">
               <a class="term" data-term="Principal Campaign Committee">Principal Committee</a>
             </h5>
-            {% for c in committees if c.designation == 'P' %}
+            {% for c in committees_authorized if c.designation == 'P' %}
```
jmcarp: To loop over all principal committees, use `for c in committee_groups['P']`. I noticed that we were looping over all committees several times, so I moved that logic into a `groupby` in `views.py`.

LindsayYoung: Sounds good. I will make the change

arowla:
This regression ought to have been caught by the test here (or at least a test like it-- tied to the principal committee link specifically)... but it seems it hasn't been run by Travis/caught by the test runner. We should make sure that this test actually does pass with this fix.

https://github.com/18F/openFEC-web-app/blob/develop/tests/selenium/single_candidate_page.py#L15

To invoke just that test:

    $ py.test --selenium tests/selenium/single_candidate_page.py::SingleCandidatePageTests::testCommitteeLink

arowla:
Oh and I have a fix for the syntax errors in that file here: https://github.com/18F/openFEC-web-app/pull/246

PR #245, opened 2015-06-08
Feature/contextual search box

jmcarp:
Set search type to current page type.

When a user browses to the committee search page or a committee detail
page, set the typeahead search type to committees; else set to
candidates. Update tests accordingly.

[Resolves https://github.com/18F/openFEC/issues/918]

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -140,12 +140,12 @@ def committee_page(c_id, cycle=None):
 
 @app.route('/candidates')
 def candidates():
-    return render_template('candidates.html')
+    return render_template('candidates.html', result_type='candidates')
 
 
 @app.route('/committees')
 def committees():
-    return render_template('committees.html')
+    return render_template('committees.html', result_type='committees')
 
 
 @app.errorhandler(404)
--- a/openfecwebapp/views.py
+++ b/openfecwebapp/views.py
@@ -7,14 +7,12 @@ from openfecwebapp.api_caller import load_cmte_financials
 
 
 def render_search_results(results, query, result_type):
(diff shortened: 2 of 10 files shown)


msecret:
@noahmanger do we also want it so when you visit the committees/candidates page, `/committees` the  search will be the page type? Or is that not necessary/expected?

noahmanger:
Oh good question. I don't actually know. We'll have to ask users. In
general I think best to default to candidates.

On Mon, Jun 8, 2015 at 4:14 PM, Marco Segreto <notifications@github.com>
wrote:

> @noahmanger <https://github.com/noahmanger> do we also want it so when
> you visit the committees/candidates page, /committees the search will be
> the page type? Or is that not necessary/expected?
>
> —
> <https://github.com/18F/openFEC-web-app/pull/245#issuecomment-110168824>.
>



-- 
Noah Manger

jmcarp:
Based on discussion above, we might want to revert a few lines of this patch. Here's the current behavior after merging--@arowla, @noahmanger, @msecret, let me know if anything should be changed.

* Candidate list page (`/candidates`): highlight candidates
* Candidate search page (`/?search_type=candidates`): highlight candidates
* Candidate detail page (`/candidate/<candidate_id>`): highlight candidates
* Committee list page (`/committees`): highlight committees
* Committee search page (`/?search_type=committees`): highlight committees
* Committee detail page (`/committee/<committee_id>`): highlight committees
* Everything else: highlight candidates

That's what I would expect as a user, but if you have different intuitions, or find otherwise in testing, this is all trivial to change.

PR #246, opened 2015-06-08
Fixed syntax errors

arowla:
Fixing the syntax in the amazing test that flew under the radar with Python syntax errors.

I make no claims on actual test results...


arowla:
AFAICT, this only had a failure due to a transient issue with sauce connect. To potential reviewers: merge with confidence.

jmcarp:
:+1:

PR #248, opened 2015-06-08
Add state and district links to candidate detail page.

jmcarp:
[Resolves https://github.com/18F/openFEC/issues/666]


noahmanger:
I poked at this just because I was interested. I think that instead of separately linking to the state and district it should just link to the combined state + district filters for that two-year period, so that you can see who else ran there.

jmcarp:
Good catch--just a mistake on my end :confounded:. Should be fixed in 606abe2.

PR #249, opened 2015-06-08
[WIP] The production API needs to be updated before this is merged.

LindsayYoung:
- Linking to the right github repo
 - adding a link to API in footer

### DIFF
--- a/templates/layouts/main.html
+++ b/templates/layouts/main.html
@@ -110,7 +110,9 @@
     <img src="/static/img/18f.png" alt="18F Logo" class="u-float-left">
     <ul class="u-float-left">
       <li>A project of <a href="https://18f.gsa.gov">18F</a></li>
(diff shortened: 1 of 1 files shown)


templates/layouts/main.html
```diff
@@ -110,7 +110,9 @@
     <img src="/static/img/18f.png" alt="18F Logo" class="u-float-left">
     <ul class="u-float-left">
       <li>A project of <a href="https://18f.gsa.gov">18F</a></li>
-      <li><a href="https://github.com/18F/openFEC/">View project source code</a></li>
+      <li><a href="https://api.open.fec.gov/developers">View API</a></li>
```
jmcarp: We already have a redirect from the webapp to the API developers page at `/developers`. You can use `<a href="{{ url_for('developers') }}">` here, and you'll get a link to the appropriate API docs (dev, staging, or prod).

PR #251, opened 2015-06-09
Update Develop

LindsayYoung:
All of the work since May 14


msecret:
@LindsayYoung , the tests are all failing for this, is that expected?

PR #253, opened 2015-06-09
Close open navigation items with escape.

### LINKED ISSUE #872: bug: Election page "Contributions by size" should be "Individual contributions by size"
The aggregates we show, come from contributions [from individuals](https://github.com/18F/openFEC/blob/develop/data/sql_incremental_aggregates/prepare_schedule_a_aggregate_size.sql#L25).

~~update links to add `contributor_type=individual` on election pages~~
~~update links to add `contributor_type=individual` on committee pages~~
 - [ ] update words on election pages

screen shot 2015-10-21 at 12 49 32 pm

### LINKED ISSUE #934: Build version of 2/4/6 cycle select for usability testing
Based on design direction determined by #933 , build and implement cycle select on staging for user testing.

Story map card: [2/4/6 cycle select](https://18f.storiesonboard.com/m/fec/!card/qv-XXBL-Je7ikjnTowJqCg)

jmcarp:
[Resolves #872]
[Resolves #934]

### DIFF
--- a/static/js/init.js
+++ b/static/js/init.js
@@ -3,6 +3,7 @@
 /* global require, window, document */
 
 var $ = require('jquery');
+var keyboard = require('keyboardjs');
 
 // Hack: Append jQuery to `window` for use by legacy libraries
 window.$ = window.jQuery = $;
@@ -49,7 +50,7 @@ $(document).ready(function() {
             $body.removeClass('controls--fixed');
             $body.css('padding-top', 0);
           }
-        })
+        });
     }
 
     // Reveal containers
@@ -65,7 +66,7 @@ $(document).ready(function() {
                 $revealContent.addClass('u-hidden').attr('aria-hidden', 'true');
                 $(this).html('View charts');
             }
-        })
+        });
     }
 
     // General reveal / disclosure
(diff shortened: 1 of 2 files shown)


static/js/init.js
```diff
@@ -3,6 +3,7 @@
 /* global require, window, document */
 
 var $ = require('jquery');
+var keyboard = require('keyboardjs');
```
msecret: Do we really need a whole library if we're just going to use a mapping of keys to key codes? Are we going to be using this library elsewhere for more in depth functionality?

jmcarp: We don't need it, but copying and pasting key codes from the internet is extra work and error-prone. We're also using the library in the typeahead module, but not in depth--really, all it does is key mapping lookups and a more concise event listener. It's about 10k minified. This would be easy to revert if you want--I kind of like it, but no strong preferences.

noahmanger: FWIW, I could see doing more with keyboard shortcuts down the road but
don't feel too strongly about it or whether we should include a library.

On Tue, Jun 9, 2015 at 4:32 PM, Joshua Carp <notifications@github.com>
wrote:

> In static/js/init.js
> <https://github.com/18F/openFEC-web-app/pull/253#discussion_r32076153>:
>
> > @@ -3,6 +3,7 @@
> >  /* global require, window, document */
> >
> >  var $ = require('jquery');
> > +var keyboard = require('keyboardjs');
>
> We don't need it, but copying and pasting key codes from the internet is
> extra work and error-prone. We're also using the library in the typeahead
> module, but not in depth--really, all it does is key mapping lookups and a
> more concise event listener. It's about 10k minified. This would be easy to
> revert if you want--I kind of like it, but no strong preferences.
>
> —
> <https://github.com/18F/openFEC-web-app/pull/253/files#r32076153>.
>



-- 
Noah Manger

arowla: If we don't have trouble with our assets being too heavy, then I'm :+1: to including a library for convenience.

PR #255, opened 2015-06-10
Feature/add subtotals

LindsayYoung:
People thought the reports were wrong when the subtotals were not there. This adds the subtotals back in. 

Fixes 18F/openFEC#937


jmcarp:
LGTM. Just need to remove `templates/.DS_Store` from source control.

LindsayYoung:
Thanks for catching that!

PR #256, opened 2015-06-10
Fix typos in Presidential report template.

jmcarp:
[Resolves https://github.com/18F/openFEC/issues/950]


jmcarp:
Handled in #255.

PR #260, opened 2015-06-10
Javascript disabled warning

msecret:
When a user doesn't have javascript enabled, they should see a warning
that some functionality will not work, and be informed how to enable
javascript.

I used the site http://enable-javascript.com/ as a link for explaining
how to enable it, @emileighoutlaw, please check the content.

### DIFF
--- a/templates/partials/javascript-warning.html
+++ b/templates/partials/javascript-warning.html
@@ -1,7 +1,7 @@
 <noscript>
   <div style="background-color: #DB4857; padding: 20px;">
     <h2 style="color: #ffffff">Javascript Required</h2>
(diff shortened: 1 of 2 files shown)


templates/partials/javascript-warning.html
```diff
@@ -0,0 +1,12 @@
+<noscript>
+  <div style="background-color: #DB4857; padding: 20px;">
+    <h2 style="color: #ffffff">Javascript Required</h2>
+    <p style="color: #ffffff">OpenFEC uses Javascript to provide the best possible user experience. Please enable Javascript in your browser or visit fec.gov.</p>
```
noahmanger: Let's make fec.gov a link

PR #262, opened 2015-06-11
Feature/column widths

noahmanger:
Makes the name columns on candidate and committee tables wider so you can see more of the name.

Also includes a quick fix that I think makes the beta banner look nicer:

screen shot 2015-06-10 at 5 38 02 pm


leahbannon:
ooh I like the fix

PR #263, opened 2015-06-11
Feature/datatables history

jmcarp:
Push filters to query parameters.

* Push filters to query parameters on loading data
* Update data on back/forward navigation

[Resolves https://github.com/18F/openFEC/issues/957]

Note: This patch makes use of `window.history.pushState`, which won't work in IE9 and earlier.


LindsayYoung:
This looks like after they were turning in f3 forms

http://localhost:3000/committee/C00347583?cycle=2000

LindsayYoung:
This looks good. Sorry it took so long. I got distracted finding other bugs.

PR #266, opened 2015-06-12
Hide pre-1980 cycles in election cycle pickers.

jmcarp:
Since we aren't exposing data before 1979.


jmcarp:
Tests passing @LindsayYoung!

PR #267, opened 2015-06-12
Handle null values in all charts.

jmcarp:
* Hide bars with null value.
* Hide series with all null values.
* Hide group series with all null series.

This is a follow up on https://github.com/18F/openFEC/issues/896, which
was partially resolved by #229.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -53,6 +53,27 @@ def _get_default_cycles():
     return list(range(cycle - 4, cycle + 2, 2))
 
 
+def series_has_data(values, key):
+    return next(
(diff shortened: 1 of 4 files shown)


jmcarp:
See /committee/C00005462?cycle=1980 before and after the patch. Before:
screen shot 2015-06-12 at 12 54 39 pm


After:
screen shot 2015-06-12 at 12 55 32 pm

In this case, the entire "Cash and Debt" chart is hidden, since all the values are `null`. But the patch is a little more nuanced than that--it will also hide pairs of bars that are all `null`, as well as individual `null` bars.

PR #268, opened 2015-06-15
Hide beta banner on close button.

### LINKED ISSUE #963: Update report templates to include all fields
Per @LindsayYoung 's comment here https://github.com/18F/openFEC-web-app/pull/952#issuecomment-158114173 , we should come through the report filing templates and make sure that all report fields are included in them. 

This work was originally done in jinja templates, so those would be a good place to start as a reference.

Filing this as an open issue for now and then we can figure out when we want to tackle it.

jmcarp:
[Resolves #963]

@noahmanger not sure why we were overriding the `display: none` associated with `[aria-hidden="true"]` in the first place.

### DIFF
--- a/static/styles/_components/_alerts.scss
+++ b/static/styles/_components/_alerts.scss
@@ -20,9 +20,4 @@
   .link--icon {
     text-decoration: none;
   }
-
-  &[aria-hidden="true"] {
-    display: block;
-    top: -50px;
-  }
 }


noahmanger:
Oh weird. Yeah I have no idea why that code is in there but noticed this wasn't working. Thanks for fixing!  Will merge once tests pass.

PR #276, opened 2015-06-16
Deploy using nginx.

jmcarp:
We're currently serving the webapp to the web directly using gunicorn.
Gunicorn is only designed to be used behind a proxy; their docs
recommend nginx. This patch puts gunicorn behind an nginx proxy.

Related to https://github.com/18F/openFEC/issues/977.


jmcarp:
If we go with this approach, we'll also be able to use nginx to serve static files, which will scale much better than the current approach of static serving with Flask. We can also push compression from Flask (slow) to nginx (fast).

boot.sh
```diff
@@ -0,0 +1,35 @@
...
+export LD_LIBRARY_PATH=$APP_ROOT/nginx/lib:$LD_LIBRARY_PATH
+
+conf_file=$APP_ROOT/nginx/conf/nginx.conf
+if [ -f $APP_ROOT/public/nginx.conf ]
+then
+  conf_file=$APP_ROOT/public/nginx.conf
+fi
+
+mv $conf_file $APP_ROOT/nginx/conf/orig.conf
+erb $APP_ROOT/nginx/conf/orig.conf > $APP_ROOT/nginx/conf/nginx.conf
```
arowla: Niiiice. Never seen `erb` used this way before.

boot.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# ------------------------------------------------------------------------------------------------
+
+touch $APP_ROOT/nginx/logs/access.log
+touch $APP_ROOT/nginx/logs/error.log
```
arowla: Are we going to be able to access these logs via CF?

openfecwebapp/config.py
```diff
@@ -5,7 +5,7 @@
 api_location_public = os.getenv('FEC_WEB_API_URL_PUBLIC', api_location)
 api_version = os.getenv('FEC_WEB_API_VERSION', 'v1')
 host = os.getenv('FEC_WEB_HOST', '0.0.0.0')
-port = os.getenv('VCAP_APP_PORT', '3000')
+port = os.getenv('FEC_WEB_PORT', '3000')
```
arowla: Are we sure changing this isn't going to mess up CF? I recall `VCAP_APP_PORT` being a magical var name...

boot.sh
jmcarp: Yeah, this script is `tail`-ing them, and they seem to show up in `cf logs`.

arowla: :+1:

openfecwebapp/config.py
jmcarp: It turns out that this code never gets touched on CF, which is confusing. This is only used when we hit the `if __name__ == '__main__'` block in `__init__.py`, which doesn't happen when running the app with gunicorn.

In other words, whenever `VCAP_APP_PORT` is defined on the environment, this variable isn't used. Unless we want to use it for local development, which I haven't been doing so far.

boot.sh
jmcarp: Seems like this is how lots of deployment / orchestration tools work (Ansible does the same thing, but with Jinja). Anyway, it wasn't my idea--this is copied exactly from https://github.com/cloudfoundry/staticfile-buildpack. Will note that in the file.

PR #277, opened 2015-06-16
updated glossary icons

noahmanger:
This uses Jen's idea of a book icon in the header with little "i" (info) icons inline. I _think_ the combo works well and is more clear than the question mark. That said, we won't know until we can get it in front of people, but was an easy change to make. 

screen shot 2015-06-16 at 12 06 09 pm

This partially addresses https://github.com/18F/openFEC/issues/910 in so much as it attempts to make the glossary more discoverable. I still think a label could be a good addition, but might not be necessary.


jmcarp:
This looks good to me. @noahmanger, do you want to get more feedback on this, or is it ready to merge?

noahmanger:
Anyone have objections? My $.02 is we should try it out.

rdpiazza:
Really like the way the glossary icon looks, but I'd like to wait until the next iteration, since there could be things we'd overlook.  For example, if we change the image on the top of the bar, we'll also need to change the text and image for "Need Help Understanding the Terms?" on the home page.  Want to limit changes this close to go time.

arowla:
I think this is a big improvement, but agree that waiting is probably best.

noahmanger:
+1 . Sounds good.

On Wed, Jun 17, 2015 at 7:28 AM, Alison Rowland <notifications@github.com>
wrote:

> I think this is a big improvement, but agree that waiting is probably best.
>
> —
> <https://github.com/18F/openFEC-web-app/pull/277#issuecomment-112824107>.
>



-- 
Noah Manger

PR #287, opened 2015-06-22
Feature/ie only

LindsayYoung:
This is not working. I think it might be breaking because it is not really an accordion and it only has 2 things on it. 

I am using C90013103 and an example Independent Expenditure only committee.


LindsayYoung:
Fixes 18F/openFEC#939

PR #301, opened 2015-06-26
Feature/redesign search box

noahmanger:
Dropdown is replaced by radio buttons on small screens. We should test in usability testing if it's better to even have the radios on desktop. Might be.

radio

Addresses https://github.com/18F/openFEC/issues/976


noahmanger:
Thanks @msecret for helping me trouble shoot the tests.

LindsayYoung:
This is looking weird to me locally. I tried to do npm install but that didn't help.
screen shot 2015-06-29 at 12 15 12 pm

noahmanger:
Make sure you rebuild Sass. The media query hiding the radio buttons on larger displays just isn't working.

Issue #303, opened 2015-06-26
Implement new committee type / designation filters


noahmanger:
Based on the conversation [here](https://github.com/18F/openFEC/issues/519), reimplement the way the filters for designation, committee type and org type show up on the browse committee page following these mocks:

![option 2](https://cloud.githubusercontent.com/assets/1696495/8383827/940485e8-1bf1-11e5-9221-62a3f702cf3c.png)

The three fields should be condensed into a single column as `Designation - type - org type`. Not sure what the column should be called. Maybe just "Committee type"?

noahmanger:
Resolved in #304

PR #304, opened 2015-06-26
Feature/new committee type filters

noahmanger:
First step at re-organizing the committee filters. It differs from the mocks in #303 in that whereas the mocks consolidated certain options (like "Party - Nonqualified" "Party - Qualified" into just "Party"), this keeps them separate. Not sure if we want to go through the trouble of modifying whatever needs to be modified in order to have one option contain two separate values, but my guess is it's not worth it.

Also, this doesn't consolidate the three columns into one in the table. Is it possible for DataTables to do that?


msecret:
One thing I noticed while testing this: you can add the same filter multiple times by clicking it again and again in the dropdown. Not sure if this was expected at this point.

msecret:
Wait ignore that comment above^, I was mistaken because two party names were very close so didn't realize the were different.

PR #306, opened 2015-06-29
First pass at data tables for filings on committee pages and /filings

LindsayYoung:
URL is not being constructed properly yet.


static/js/modules/tables.js
```diff
@@ -94,6 +110,51 @@ var committeeColumns = [
...
+    orderable: false,
+    render: function(data, type, row, meta) {
+      var anchor = document.createElement('a');
+      anchor.textContent = 'View filing';
+      anchor.setAttribute('href', data);
+      anchor.setAttribute('target', '_blank');
+      return anchor.outerHTML;
+    }
+  },
+  {data: 'committee_name', className: 'min-desktop', orderable: false},
```
jmcarp: Do we need the `committee_name` column here? Won't this be the same for every row?

LindsayYoung: For the committee page, we don't need committee name, but for the filings query builder we do since you could query across committees. That's why there are two tables. You can see the filings query builder on the front end at `/filings`.

jmcarp: Agreed, but both tables include the `committee_name` column, and the committee page table looks like this:
screen shot 2015-07-20 at 10 18 58 am

LindsayYoung: Oh, I agree- we don't need committee name on the committee page! I can fix that.

LindsayYoung:
@noahmanger the basics are here but it is still looking a bit off.

Here are my notes on the progress of this branch:
 - Right now it just filters by committee_id, need to add name
 - I just added filter by candidate_id to the dev api, when that goes through we can add the registration forms to the candidate pages
 - Need to add other document types on the API
 - Filings should only load if you go to the tab, or this will
 slow down the site
 - the two year period should be hidden on the upper right.

noahmanger:
Looking pretty good. A few thoughts:
- Is there any way to have a more descriptive name field? It would probably be something like "Report Type" + something. I'm thinking like "April Quarterly 2015" or something.
- And then I'd love for this descriptive name to replace the "View filing" link
- Do we need the "Form type" column on this page? 
- For amendment indicator can we just show something if it is amended and if not keep it blank?
- How much work is it make the nested amendment rows like we had in the mocks?
- We still need to implement the details sidebar

LindsayYoung:
These are great comments

 - Yes, I will make something that looks better in the API
 - We can add the link to in the name
 - Form type isn't listed there anymore- I think you were looking at an older version of the branch
 - The amendment indicator can be "A", "N" or null, so I was trying to be true to the underlying data. 
 - I don't know how to combine the docs and keep them sortable. It is not impossible, but it would take more thinking.
 - Sidebar is a whole separate task

Issue #293, opened 2015-06-25
Improve filter toggle button 


noahmanger:
Doesn't make sense in the light blue box. I'm playing with it.

noahmanger:
Something like this...

![](https://s3.amazonaws.com/f.cl.ly/items/0e1j1A0n3x1C2T1j1618/Screen%20Recording%202015-06-24%20at%2006.02%20PM.gif)

noahmanger:
And then on mobile it will slide down from the top.

noahmanger:
@jenniferthibault thoughts?

jenniferthibault:
The arrow directions are great! We may still need supporting text (show filters, hide filters), as I'm not sure it's obvious what will be happening. Is sideways text a bad idea?

Sorry for the hold-up for such a simple comment. Now that the meeting is done I'll be more on point

noahmanger:
Yeah I was wondering about that too. I don't know how I feel about sideways text. I don't mind it when I see it places. Two other ideas:

Include this sort of filter icon next to the arrow:
![](https://cdn0.iconfinder.com/data/icons/pixon-1/24/filter_list_order_sequence_sort_sorting_outline-512.png)

Or on hover, the tab could slide out a little and say "Filters".

jenniferthibault:
Oh yes! The slide out to say "Filter" on hover/tab focus (for 508) seems the most direct to me. I read that icon as "graphs," or "charts," more than filter.

PR #308, opened 2015-06-30
Improve filter button

msecret:
This moves the filter button to the right side of the filter bar. This
was done so the filter button UI was attached to the actual filter
UI, so the button affords closing and opening the side bar.

This iteration doesn't include any animation. Animation would require
some extra work.

Fixes https://github.com/18F/openFEC-web-app/issues/293


msecret:
This is how it's looking at various breakpoints. @jenniferthibault @noahmanger, the position of the filter buttons breaks at the smallest breakpoint when its minified. It's running into the header below it.

screen shot 2015-07-01 at 4 53 03 pm
screen shot 2015-07-01 at 4 53 20 pm
screen shot 2015-07-01 at 4 53 37 pm
screen shot 2015-07-01 at 4 53 52 pm
screen shot 2015-07-01 at 4 54 01 pm
screen shot 2015-07-01 at 4 54 11 pm
screen shot 2015-07-01 at 4 54 16 pm

jenniferthibault:
@msecret I'm working through some other tweaks to different elements on other screen sizes you've shown, but: 
A) Is this set of changes to the mobile view possible?; 
B) Is this kind of markup helpful?

artboard 20

noahmanger:
This looks way better. A couple other ideas:

1. Would love for this to default to closed on mobile so you see the data first.
2. If that's the case, we may need an indicator that filters are applied.

jenniferthibault:
Next up: Show filters on mobile screens. Same philosophy for the changes you see here as in "hide filter" above.
artboard 20

jenniferthibault:
Medium and XL screens:

artboard 23

artboard 22


Glad to talk through anything that throws a monkey wrench into something that I haven't noticed!

jenniferthibault:
I was confusing myself with all the descriptions of how the menu worked. I hope this helps to clarify!
artboard 24

noahmanger:
On small screens I'm worried that the icon switching sides with the text might be kind of a weird jump. Is there a reason you had it like that?

jenniferthibault:
THANK YOU @noahmanger ! It made sense in my head when I was making it, but I didn't pause to review it when finished. For small screens, the icon is now on the left so that people read it (and hopefully get the signal) before they have to read the text. For large screens, since the text is moving left-right, the icon is on the right so that the words disappear & reveal in-sync with the animation. Does that make sense, or should they all be on the same side?  cc @msecret — sorry for the confusion here.

artboard 24 copy

noahmanger:
Nice. Yeah I think it makes sense to be on the right on large screens. I like it!

msecret:
@noahmanger @jenniferthibault , I just updated this, here are some new screens.

screen shot 2015-07-07 at 3 27 11 pm
screen shot 2015-07-07 at 3 27 17 pm
screen shot 2015-07-07 at 3 27 29 pm
screen shot 2015-07-07 at 3 27 42 pm

msecret:
Still left to do:
- Mobile animation for filter slide up. I'm having trouble with this, not 100% we should do it right now.
- Fix staggering fields problem on tablet. I think I might want to put this in a different PR has it's pretty separate from this issue.

static/styles/_components/_side-panel.scss
```diff
@@ -23,24 +23,144 @@
 }
 
 .side-panel {
```
noahmanger: This is my fault for the way this was originally implemented, but the glossary also has the `.side-panel` class, so all of these style changes break it. We could either move them to `.side-panel--left` or remove the `.side-panel` class from the glossary. (They started out as similarly functioning things, so made sense for them to share some styles, but they're no pretty different so maybe it makes more sense to break them out).

static/styles/_components/_side-panel.scss
```diff
@@ -23,24 +23,144 @@
...
-  padding: 2rem;
-  top: 0;
+  background: $lightest-gray;
+  border-color: $light-gray;
+  border-style: solid;
+  border-bottom-width: 1px;
+  border-top-width: 0;
   bottom: 0;
+  //padding: 1rem;
```
noahmanger: Does this need to stay in here?

static/styles/_components/_side-panel.scss
```diff
@@ -23,24 +23,144 @@
...
+  position: absolute;
+  top: 0;
+  width: 100%;
+
+  h4 {
+    margin-top: 1em;
+  }
+
+  &.side-panel--open {
```
noahmanger: Does this need to be nested or can it be moved up a level?

static/styles/_components/_side-panel.scss
```diff
@@ -23,24 +23,144 @@
...
+    float: left;
+    margin-left: -17%;
+    position: static;
+    width: 17%;
+
+    .side-panel__box {
+      position: relative;
+    }
+
+    .side-panel__toggle.side-panel__toggle {
```
noahmanger: Is this a typo?

templates/candidates.html
```diff
@@ -11,27 +11,15 @@
...
-        {% include 'partials/filters/parties.html' %}
-        {% include 'partials/filters/states.html' %}
-        {% include 'partials/filters/districts.html' %}
-
-      <div class="field">
-        <button type="submit" class="primary">Apply filters</button>
-      </div>
-    </form>
-  </div>
+  {% set filters = [
```
noahmanger: Yessss. Love this.

static/styles/_components/_side-panel.scss
msecret: Unfortunately, no. It's a big of a specificity hack to get around having to use `!important`.

noahmanger: Oh cool. I had no idea.

static/js/modules/tables.js
```diff
@@ -165,6 +165,7 @@ function initTable($table, $form, baseUrl, columns) {
   var $paging = $(api.table().container()).find('.results-info--top');
   $paging.prepend($('#filter-toggle'));
```
noahmanger: This line needs to be removed. It's moving the button.

PR #302, opened 2015-06-26
Feature/itemized tables

jmcarp:
Still in progress; not ready for merge!

PR #312, opened 2015-07-02
Feature/committee file date

### LINKED ISSUE #289: Enable filtering and sorting by Form 1 Filing Date on /committees
**Goal:** Users should be able to filter for committees registered on (or after?) a particular date and also to sort by filing date.

**Completion Criteria:** Implemented on site.

**Steps:**
- [x] Define behavior
- [x] Design interface
- [ ] Implement

**Screen:**
This is referring to another filter and column to add to the committee browse page: 
screen shot 2015-06-29 at 12 47 25 pm

jmcarp:
Add file date filter and column.

[Resolves #289]

This is working as expected, but probably needs more discussion; see #289.


jmcarp:
This has been updated per Slack / GitHub discussion @noahmanger @LindsayYoung.

jmcarp:
Pinging @noahmanger @LindsayYoung for review. After we talked this through last week, the patch became very simple and should be quick to review--it basically just adds a sortable column on first file date, and a date formatter helper for the frontend. There's also some utility code for calculating the current month, quarter, and year, which isn't used here but will be in #302.

noahmanger:
Looks good, but the problem is that there's no way to see the most recent because there's a bunch of committees with blanks in that column. How should we handle this?

jmcarp:
We could change the sorting logic to filter out `NULL` values--if you're sorting on date, whether ascending or descending, you probably only want to see records that have a date.

jmcarp:
We could also have the database treat `NULL` values as small (by default, they're considered large). That would help with finding the most recent records, but then looking for the least recent would return a bunch of records with `NULL` dates.

jmcarp:
Any preferences @noahmanger @LindsayYoung? I think omitting records with null values on the sorted column is reasonable, but maybe there are better solutions.

jmcarp:
@noahmanger the point you brought up about empty rows will be resolved once #322 is merged.

msecret:
There's quite a few JS libs added in this code.

I know I hadn't completely set this up yet before I moved to less time on this project, but ideally, when theres JS libs added, I'd like to see what the impact would be. Like how many kbs added, and how that effects things like time to first paint, speed index, or custom metrics. Although I understand this is hard to do now, just wanted to comment as a good place to move to.

noahmanger:
Good looking out @msecret . What do you all think about hiding the "Treasurer Name" column? @LindsayYoung I think this was a suggestion you made a while back?

jmcarp:
The libraries added here should each be < 30 kB, minified and gzipped. If we're looking to optimize load times, I'm guessing we'd get the most traction from splitting the browserify bundle into pieces--loading d3,  handlebars, and datatables on the homepage (which uses none of the above).

noahmanger:
Should we make a separate issue for looking at performance and making these sorts of changes?

jmcarp:
We have an issue for splitting code bundles: https://github.com/18F/openFEC/issues/908. We also have performance tracking on dev, which we should be able to track with New Relic (is that right @msecret?)--maybe we need an issue for gathering / evaluating those performance numbers.

msecret:
I'd like to check out New Relic to see what performance tracking they have implemented.

So the thing with performance that I've gathered: the best metric to really get a sense of how your users are experiencing the site is SpeedIndex. It's also the hardest to collect, essentially requiring a webpagetest.org instance running to capture a video of the site loading. This means that often an easier to measure metric, that also comes closer emulates the SpeedIndex metric is a custom metric. For us on FEC, this would be different on each view/page, and could be things like, "time till the whole results table shows", "time till search bar shows", "time till search results appear", etc. I think defining these custom metrics and coming up with what our goal for them would be a good step forward for tracking performance.

LindsayYoung:
I have mixed feelings on Treasurer name- it is very important since these are the people legally responsible, but it is not necessary everywhere.

noahmanger:
Yeah, I share that. If we're going to keep it we should also have a filter
for it, which I just realized we're missing. Maybe we just leave it in for
now.

On Mon, Jul 13, 2015 at 10:58 AM, Lindsay Young <notifications@github.com>
wrote:

> I have mixed feelings on Treasurer name- it is very important since these
> are the people legally responsible, but it is not necessary everywhere.
>
> —
> <https://github.com/18F/openFEC-web-app/pull/312#issuecomment-121006804>.
>



-- 
Noah Manger

noahmanger:
This looks good to me, but @msecret should probably review the js / python if you haven't already.

jmcarp:
This should be good to merge @msecret / @noahmanger --it's just adding a column and updating some tests.

Issue #320, opened 2015-07-06
Implement Schedule A data on committee page


jmcarp:
Implementation of #316

* [x] Implement aggregates by contribution size
* [x] Implement aggregates by contributor
* [x] Make everything look nice

See also https://github.com/18F/openFEC/issues/1006 for API updates.

PR #324, opened 2015-07-09
Committee Schedule A Tab

jmcarp:
This implements the first part of #320, per-committee aggregates by contribution size.


jmcarp:
Note: Depends on https://github.com/18F/openFEC/pull/1043.

jmcarp:
This is incomplete (needs styling and table overlay), but I would be comfortable having it merged after code review @noahmanger @msecret.

static/js/modules/maps.js
```diff
@@ -0,0 +1,152 @@
...
+
+var compactRules = [
+  ['B', 9],
+  ['M', 6],
+  ['k', 3]
+];
+
+function chooseRule(value) {
+  var rule;
+  for (var i=0; i<compactRules.length; i++) {
```
msecret: can you add more space like `i < compactRules.length`

static/js/modules/maps.js
```diff
@@ -0,0 +1,152 @@
...
+  }
+  return null;
+}
+
+function compactNumber(value, rule) {
+  var divisor = Math.pow(10, rule[1]);
+  return d3.round(value / divisor, 1).toString() + rule[0];
+}
+
+var quantiles = 7;
```
msecret: Can this be moved to where the compactRules var is defined?

static/js/modules/tables.js
```diff
@@ -356,49 +422,134 @@ function initTable($table, $form, baseUrl, columns, callbacks, opts) {
...
+          if (year) {
+            query.year = year;
+          } else {
+            query.cycle = cycle;
           }
-        );
-        break;
-    }
+          initTable($table, $form, path, query, committeeContributorColumns, offsetCallbacks, {
+            dom: '<"results-info meta-box results-info--top"lfrip>t',
```
msecret: is there supposed to be a class= attribute here?

static/js/modules/maps.js
jmcarp: Even better, I'm replacing this with `_.first`. Who needs loops anyway.

static/js/modules/tables.js
jmcarp: No, because the DataTables API confusingly treats anything in quotes as a class in this context: https://datatables.net/reference/option/dom. Unless you prepend a hash, in which case it's an ID.

static/js/modules/maps.js
jmcarp: :+1:

jmcarp:
Updated based on inline comments from @msecret.

templates/partials/contributions-tab.html
```diff
@@ -0,0 +1,126 @@
...
+        </div>
+        <div class="table__cell">
+          <h5>Contribution Size</h5>
+        </div>
+      </div>
+      <div class="table js-chart-series chart-series--horizontal">
+        {% for value, limits, label in table_data %}
+          <div class="table__row table__row--nested js-accordion_item">
+            <div class="table__cell">
+              <a
```
noahmanger: This link should also include the date filter, right? So if you only see contributions pegged to this cycle.

noahmanger:
This is looking really good. I'm fine to merge as-is once Marco's comments are resolved. There's a few additional things we need though (which can come in a later PR): each of the aggregates (state, employer, occupation) should link to the transaction table for those, like the amount aggregates do.

templates/partials/contributions-tab.html
jmcarp: Good call, will add.

PR #333, opened 2015-07-15
Fix filter panel on /donations

msecret:
The filter panel wasn't working on this view because the html code
had to be updated to work with the new CSS and JS.

### DIFF
--- a/templates/donations.html
+++ b/templates/donations.html
@@ -14,60 +14,62 @@
 </header>
 
 <section class="results-content">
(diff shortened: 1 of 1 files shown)


noahmanger:
Hrm. Still looking wonky to me:

screen shot 2015-07-15 at 4 26 23 pm

msecret:
@noahmanger I'm going to fix that separately because it's a global bug on the whole filter panel.

noahmanger:
Ah got it.

PR #337, opened 2015-07-20
Use Jinja blocks for flexible generic filters.

jmcarp:
This is an implementation of reusable filter sidebars, as @msecret proposed, that doesn't require all filter fields to be template includes. This approach uses blocks, which are more flexible and easier to customize.

This will cause merge conflicts with #336, so if we merge this branch, I can cherry-pick the SCSS changes there, or we can resubmit #336.


msecret:
This is great. How do you propose we deal with the merge conflict? I could remove all template changes from #336 and then commit just the CSS changes? It *looks* like none of the changes I made in the templates were related to the styles.

jmcarp:
Agreed, I think reverting the non-CSS changes in #336 would work. Or we could make a new PR that only includes the CSS changes--whatever's easier.

msecret:
I make a new PR here:

https://github.com/18F/openFEC-web-app/pull/338

There were some template changes, unfortunately. I had to create a new wrapping div with a new class.

msecret:
I have another fix to do on the filter panel but this looks good.

PR #338, opened 2015-07-20
Fix the filter panel, not the template org

msecret:
This fixes the filter panel on mobile by creating a new component
element. It does not fix the template organization problem. This
was done to make the merge conflict in #337 easier as it's less code
that will be modified.


jmcarp:
@noahmanger: I merged this patch into #337, so merging that one (when ready) should take care of both. @msecret: Thanks for fixing conflicts.

PR #346, opened 2015-07-21
Feature/add purpose filter

jmcarp:
Add disbursement purpose filter and column to Schedule B data table. @noahmanger: the original issue only mentioned the filter, but I think it's helpful to include filtered columns in the table if possible so users can understand how the filters work. I can take out the extra column if it's distracting though.

### DIFF
--- a/static/js/modules/tables.js
+++ b/static/js/modules/tables.js
@@ -212,6 +212,7 @@ var expenditureColumns = [
       return Handlebars.helpers.datetime(data);
     }
   },
+  {data: 'disbursement_description', className: 'min-tablet', orderable: false},
   {
     data: 'committee',
     orderable: false,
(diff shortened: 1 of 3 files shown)


msecret:
I see the purpose filter on the filters tab and the correct purpose in the table, but when I type something into the purpose filter, it doesn't seem to actual filter anything. Is that expected at this point?

noahmanger:
Yep, you're right. Good catch adding the column.

On Tue, Jul 21, 2015 at 9:17 AM, Marco Segreto <notifications@github.com>
wrote:

> I see the purpose filter on the filters tab and the correct purpose in the
> table, but when I type something into the purpose filter, it doesn't seem
> to actual filter anything. Is that expected at this point?
>
> —
> <https://github.com/18F/openFEC-web-app/pull/346#issuecomment-123390119>.
>



-- 
Noah Manger

jmcarp:
@msecret this won't work using the dev API until https://github.com/18F/openFEC/pull/1076 is merged. If you're running your own API server and pointing at the production RDS, it should work. Sorry, I should've mentioned this before.

PR #348, opened 2015-07-21
Feature/committee schedule b

jmcarp:
Branched off #324.

This is just a new tab with a toggle between three aggregate tables, so much simpler than the Schedule A tab. Note: aggregation by purpose isn't working at the moment, since purpose categories are WIP on the API side.


jmcarp:
@msecret this is updated with develop and ready for review.

noahmanger:
Just caught a couple things and filed a PR against your branch to fix them while they were front of mind.

PR #349, opened 2015-07-22
Feature/styling contributions tab

noahmanger:
This is a first pass at cleaning up the contributions tab. There's more work to be done, such as getting the map right (the table header wants to be fixed at 100px wide for some reason), but I think we might as well merge these in. 

screen shot 2015-07-21 at 6 00 27 pm


jmcarp:
This looks great! I see that some tests are failing on Travis, but they're passing for me locally. I'm going to merge now and resolve Selenium issues later.

PR #362, opened 2015-07-23
Styling committee filing filters

noahmanger:
1. Tightens up the filters on the filings table on committee pages
screen shot 2015-07-23 at 3 33 48 pm

2. Moves the table page controls below tables on the committee page. @jenniferthibault this was the pattern you were using on election pages, correct?
screen shot 2015-07-23 at 3 34 34 pm

### DIFF
--- a/static/js/pages/committee-single.js
+++ b/static/js/pages/committee-single.js
@@ -9,7 +9,7 @@ var events = require('../modules/events');
 var tables = require('../modules/tables');
 var helpers = require('../modules/helpers');
 
-var singlePageTableDOM = 't<"results-info meta-box"frip>';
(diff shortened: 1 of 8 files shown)


noahmanger:
The tests passed locally for me. Not sure why they're not working here.

jenniferthibault:
I had the pagination box connected the bottom of the table, because I thought it helped more with visually unifying each element when there were multiple tables in a row. That could change though! 

Is it just me, or is the text in the pagination box larger than the text in the table? Could we make sure they are the same size universally? (that, or that the pagination text is the smaller of the two?)

noahmanger:
Nice. I like it:

screen shot 2015-07-23 at 6 33 49 pm

Issue #351, opened 2015-07-22
Figure out consistent behavior and location for cycle select


noahmanger:
We need to figure out of the cycle select on committee pages a) applies across all tabs and if not b) where it lives on the page.

If it applies across all tabs, it can be where it currently is. But this would mean making it filter the filings tab, which it currently doesn't do. Though after playing with filters, I think maybe it should, because otherwise there's no way to easily sort to the report with the most receipts in the current election period.

If not, then I think it should live either on the heading of each individual tab _or_ next to each data table and control each individually.

Wanted to log this here for discussion.

noahmanger:
Check out my proposed solution at #364

PR #364, opened 2015-07-24
Cycle select update

noahmanger:
Based on #351, I wanted to come up with a better, more consistent place for the cycle select. So this patch does just that. @jenniferthibault can you review to make sure this captures what we discussed yesterday?

There are now cycle selects on any tab where you can control the cycle (so everything except "Filings"), and changing one changes them all.

screenshot-02
screen shot 2015-07-23 at 5 44 26 pm
screen shot 2015-07-23 at 5 44 33 pm

*Also:* I moved the logic on the Candidates and Committees page so that even if there's no records, you still see the "financial summary" section heading and explanatory text.

screen shot 2015-07-23 at 5 45 38 pm

*To do:* Once we have persistent tab URLs, you'll no longer be bumped to the first tab upon changing the select on the second or third tab.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -262,7 +262,7 @@ def date_filter_md(date_str):
 @app.template_filter()
 def fmt_year_range(year):
     if type(year) == int:
-        return "{} - {}".format(year - 1, year)
+        return "{}–{}".format(year - 1, year)
     return None
 
 
--- a/static/js/modules/filters.js
+++ b/static/js/modules/filters.js
@@ -76,10 +76,14 @@ var activateFilter = function(opts) {
 };
 
 var bindFilters = function() {
-    var cycleSelect = $('#cycle').change(function() {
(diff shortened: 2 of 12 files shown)


jenniferthibault:
Yup! This looks like what we talked about to me!
One thing, and apologies if this is late in the game to bring it up, I'm not sure how much of a PITA it will be.

I'd like to be sure that any time we are showing a range of numbers or a range of dates, we use an en-dash without surrounding spaces to separate them instead of a hyphen. An en-dash is typed on a Mac by holding `option`+`-` and releasing. On Windows, it's `Alt`+ `0``1``5``1` and release. En dashes are used exclusively for date and numerical ranges. 
_________

**Examples**
*Hyphen with spaces:* 2012 - 2013, $100 - $500
*En dash without spaces:* 2012–2013, $100–$500

PR #366, opened 2015-07-24
fixing styles that were overridden

noahmanger:
Some changes apparently got merged over in my various pull requests so this gets the filings tab looking right.


jmcarp:
LGTM!

PR #377, opened 2015-07-28
Feature/panel layout mobile

msecret:
This includes completed styling of the mobile panel (which acts more like a modal) minus the links to link and the original image link.

Three fields: "Type", "Report", "Amended" were unknown, so I either used defaults or left off. I will need to know what this information is and looks like from the API.

Layouts larger then mobile look fine, and don't change much at this time.

Used this design:
https://files.slack.com/files-pri/T025AQGAN-F07MWATEG/mobile_2.png\

Here's what it looks like:
screen shot 2015-07-28 at 2 28 10 pm
screen shot 2015-07-28 at 2 28 31 pm

### DIFF
--- a/static/js/modules/tables.js
+++ b/static/js/modules/tables.js
@@ -135,16 +135,26 @@ function mapQuerySeek(api, data) {
 }
 
 function modalAfterRender(template, api, data, response) {
-  var $table = $(api.table().node());
-  $table.on('click', '.js-panel-toggle tr', function(e) {
-    var $row = $(e.target).closest('tr');
-    if ($row.is('a')) {
+  var $table = $(api.table().node()),
+      $modal = $('#datatable-modal');
+
+  // Move the modal to the results div.
+  $modal.appendTo($('#results'));
+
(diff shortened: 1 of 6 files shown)


msecret:
I've checked the tests, and I don't think the failure is relevant. It has to do with a database error:
`The command "pg_restore --dbname cfdm_test data/subset.dump --no-acl --no-owner" failed and exited with 1 during .`

noahmanger:
Aaaalmost there. Looks like the footer is covering the bottom part of the panel though:

noahmanger:
Gah. So it works better, but now the problem is that if you click on a row further down the page, the panel is all the way at the top and you don't realize it's there. Pretty sure this is all from my poorly thought out concept, but let me know if you want to think through it.

msecret:
@noahmanger I'm not experiencing that because the page height shrinks. But yea, lets look at this together, I'm a little unsure of the current design as the viewport units could cause problems.

noahmanger:
Oh weird. The height's not shrinking for me.

PR #386, opened 2015-07-29
Feature/fix filters

jmcarp:
Fix recipient committee and date filters.

Incidentally, moves page-specific logic from modules to pages.

### DIFF
--- a/static/js/modules/filters.js
+++ b/static/js/modules/filters.js
@@ -208,26 +208,11 @@ $('.field input[type="text"]').on('keypress', function(e) {
     }
 });
 
-function bindFileFilter() {
-  var $field = $('#file-date');
-  var $startDate = $field.find('[name="start_date"]');
-  var $endDate = $field.find('[name="end_date"]');
-  $field.on('click', '[name="_file_date"]', function(e) {
-    var $input = $(e.target);
-    if ($input.attr('data-start-date')) {
-      $startDate.val($input.attr('data-start-date'));
-      $endDate.val($input.attr('data-end-date'));
-    }
-    $startDate.focus();
-  });
-}
-
 module.exports = {
   init: function() {
     bindFilters();
     // if the page was loaded with filters set in the query string
     activateInitialFilters();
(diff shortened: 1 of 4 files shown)


static/js/modules/filters.js
```diff
@@ -208,26 +208,11 @@ $('.field input[type="text"]').on('keypress', function(e) {
     }
 });
 
-function bindFileFilter() {
```
noahmanger: Can we make this more general, like `bindDateFilter`? It will be used for contribution date, disbursement date, filing date, and who knows what else.

static/js/pages/receipts.js
```diff
@@ -65,4 +68,25 @@ $(document).ready(function() {
...
+  // Set up committee typeahead
+  var $committeeFilter = $('#committee-typeahead-filter');
+  var $committeeInput = $('input[name="committee_id"]');
+  $committeeFilter.typeahead({}, typeahead.committeeDataSet);
+  $committeeFilter.on('typeahead:selected', function(event, datum, datasetName) {
+    $committeeInput.val(datum.id);
+  });
+
+  // Set up date picker
+  var $field = $('#file-date');
```
noahmanger: Let's make this more general...maybe `#date-picker`? It will be used for more than just filing date.

noahmanger: Also, this needs to be used on the disbursements page as well.

static/js/pages/receipts.js
```diff
@@ -65,4 +68,25 @@ $(document).ready(function() {
       pagingType: 'simple'
     }
   );
+
```
noahmanger: Should this be moved to a different file? Will be used by /disbursements as well.

noahmanger:
This looks good to me other than the couple questions about where the code should live / what it should be called. Maybe it's useful to see my #387 for how I was thinking we should move the date picker into it's partial to be re-used for different filters.

jmcarp:
Incorporated into #387.

PR #387, opened 2015-07-29
Feature/filter fixes

noahmanger:
This does a few things to clean up the filters:

1. Groups /disbursements filters into accordions (like the /receipts filters)
2. Adds the date filters to /disbursements and moves those fields to a macro
3. Positions the "remove" button inside text inputs
4. Various other little style cleanups (the sidebar / filter code is getting gnarly and is probably due for a refactor soon)

I imagine the stuff I did to dates is going to conflict with #386, so let's merge that first and then I'll make the appropriate changes.


static/styles/_components/_filters.scss
```diff
@@ -76,6 +82,17 @@
   }
 }
 
+.filter__header {
+  padding: 0 1rem;
+}
+
+.accordion__item {
```
msecret: Is there any reason this wouldn't be put in accordion.scss. Putting it here makes it harder to find.

noahmanger: Good catch. Moved to `_side-panel.scss` along with other modifications to the accordion components from being within `.side-panel`.

templates/macros/filters/date.html
```diff
@@ -0,0 +1,20 @@
...
+  <ul>
+    <li><input name="{{ name }}" type="radio" data-start-date={{ dates['today'][0] | date }} data-end-date={{ dates['today'][1] | date }}>Today</li>
+    <li><input name="{{ name }}" type="radio" data-start-date={{ dates['month'][0] | date }} data-end-date={{ dates['month'][1] | date }}>This month</li>
+    <li><input name="{{ name }}" type="radio" data-start-date={{ dates['quarter'][0] | date }} data-end-date={{ dates['quarter'][1] | date }}>This quarter</li>
+    <li><input name="{{ name }}" type="radio" data-start-date={{ dates['year'][0] | date }} data-end-date={{ dates['year'][1] | date }}>This year</li>
+    <li><input name="{{ name }}" type="radio">Custom</li>
+  </ul>
+  <div class="date-range-input">
+    <label for="min_date">From</label>
+    <input type="text" id="min_date" name="min_date" data-inputmask="'alias': 'mm/dd/yyyy'">
```
msecret: Is there code that uses this inputmask?

When doing research for the pattern library, we found traditional input mask code can be problematic for screen readers: https://www.filamentgroup.com/lab/politespace.html

If we're using a traditional input mask, I suggest a discussion with Bristow to check, I would be fine merging this before that happens though.

noahmanger: Ah. Just saw that discussion. Yeah, we're currently using it, but I'll file an issue to investigate further.

templates/macros/filters/dates.html
```diff
@@ -0,0 +1,20 @@
+{% macro field(name, label, dates) %}
```
msecret: Is this file the same as macros/filters/date.html? Whats the reason for having both?

noahmanger: Arg. Re-naming files gone awry is all. Deleting.

jmcarp:
Added a few fixes / generalizations post-merge in cf0be5e2d085db7d9d2fbf4939033657b687b9c6 and 0c9cfa1fefb9bbdcc3ba44ead9383388fdca6840.

Issue #365, opened 2015-07-24
Submit all filter forms on change


jmcarp:
We currently auto-submit the filter form for the filings table on the committee detail page on change, but other filter forms have a submit button that the user has to click on to update results. @noahmanger mentioned that we should change all filter forms to auto-submit.

noahmanger:
Closed in #390

PR #390, opened 2015-07-30
reloading tables on form input change

noahmanger:
Makes it so that the datatables reload when a filter changes. No more submit button!

(Local tests are being grumpy, so we'll see if they pass on Travis.)


noahmanger:
Addresses #365

PR #392, opened 2015-08-02
Feature/panel desktop

msecret:
Check commit messages for explanations

### DIFF
--- a/static/styles/_components/_panel.scss
+++ b/static/styles/_components/_panel.scss
@@ -12,11 +12,25 @@
   &[aria-hidden="false"] {
     @include transform(translateX(0%));
   }
+
+  @include media($medium) {
+    &[aria-hidden="false"] {
+      @include transform(translateX(100%));
+    }
+
+    background: white;
+    border: 1px solid $light-gray;
+    width: 95%;
+  }
 }
 
 .panel {
   background: $lightest-gray;
   padding: 2rem 1rem;
+
+  @include media($medium) {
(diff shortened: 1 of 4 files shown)


noahmanger:
This is looking really good. Not sure if you're still working on it, but the only thing I see worth noting is that the active row should be highlighted.

msecret:
Yea I was still working on a few things like that. I also wanted to get the arrow on the rows done, move the close button on desktop.

What the status of those "View original image" button? Is there an existing link for that?

Also, for the perma-link, I'm guessing it will take about a day to implement that. Is it something I should do this next sprint, or are there other things that are more important?

noahmanger:
Yeah I'd say it'd be good to do that this sprint. The "original image" link should exist...it may need to be built using the image numbers. @LindsayYoung how are we doing this now?

noahmanger:
Regardless, I think we can merge this in whenever the active state is ready.

static/styles/_components/_panel.scss
```diff
@@ -29,11 +43,35 @@
...
+    margin-top: 0;
+
+    &:last-child {
+      border-bottom: none;
+    }
+  }
+}
+
+#datatable-modal .panel .panel__row {
+  // TODO this shit is broken.
```
noahmanger: What's broken?

msecret: Haha, I was supposed to go back to that and write a better comment, I'll do so now

Issue #397, opened 2015-08-03
Add committee typeahead to all browse pages


noahmanger:
- [ ] Disbursements
- [ ] Filings

noahmanger:
Happening in #406

PR #406, opened 2015-08-03
Feature/adding preview to disbursements

noahmanger:
@msecret already did all the work. I just had to do a couple things to get the preview pane to show up on `/disbursements`.


templates/disbursements.html
```diff
@@ -17,22 +17,23 @@
   </div>
 </section>
 
-<div id="datatable-modal" class="modal__overlay" aria-hidden="true">
-  <div class="modal">
-    <div class="modal__row">
-      <button title="Close menu" class="js-hide modal__close" data-hides="datatable-modal"><i class="ti-close"></i></button>
+<div id="datatable-modal" class="panel__overlay" aria-hidden="true">
```
msecret: We should probably make this a partial as it's quite a bit of duplicated code. Although, does it only appear in these two files, or is there a chance it will be in more files?

noahmanger:
Just added typeahead to committee filters on /filings and /disbursements to resolve #397

noahmanger:
I don't get why these tests are failing :(

msecret:
The test failures seem to indicate that the results in the tables aren't being populated. There's a bunch of failures to find `<tr>` elements, like this selector: `tbody tr` or even this `.tst-search_results h3 a`. Is it possible that the API wasn't working or was down while the tests ran? I'm not super familiar with the tests setup. @jmcarp would know more.

jmcarp:
I'm looking into this more in #414, hoping to have this figured out soon. Meanwhile, tests are passing for me locally, so I'm going to merge.

Issue #376, opened 2015-07-28
Add links to aggregated transactions in Receipts & Disbursements tables


noahmanger:
On the various aggregate tables, there should be a way to click through to view the itemized transactions that constitute that total as a pre-filtered set of results on /receipts or /disbursements.

noahmanger:
In #409

PR #409, opened 2015-08-04
Adding aggregate links to itemized tables

noahmanger:
I saw an opportunity to help with one of the dev tasks ( #376 ), so I went ahead and did it. This adds links to the amounts in the Receipts and Disbursements tab tables to the itemized transactions that make up the total. So when you click the Facebook total in this table:

screenshot-01

It shows the transactions that make up that total:

screen shot 2015-08-03 at 7 22 44 pm

I tried to code this up in a smart, reusable way, but please let me know if there's anything I should do to tighten it up (I imagine there is).

### DIFF
--- a/static/js/modules/helpers.js
+++ b/static/js/modules/helpers.js
@@ -24,7 +24,16 @@ function datetime(value) {
 }
 Handlebars.registerHelper('datetime', datetime);
 
+function cycleDates(value) {
+  var secondYear = value;
+  var firstYear = secondYear - 1;
+  var min_date = '01-01-' + firstYear;
+  var max_date = '12-31-' + secondYear;
+  return { min_date: min_date, max_date: max_date };
+}
+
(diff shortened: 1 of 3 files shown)


noahmanger:
Oh, I should add: for a few of the links it seems we need to implement corresponding filters on the browse view. I think the ones missing are:

- Receipts: the ID of the contributing committee
- Disbursements: the ID of the disbursing committee
- Disbursements: the ID of the receiving committee 

The last two confuse me because I thought we had those working, but maybe I'm doing something wrong. I know that #406 fixes these filters on the frontend.

LindsayYoung:
We will need to filter out non-individual line-items too or the not reported total will not match.  

That depends on making line number a parameter #18F/openFEC/1102

noahmanger:
Thanks@LindsayYoung . Just captured in #412

PR #410, opened 2015-08-04
Add the styles for the close button

msecret:
This ensure the close panel button on bigger then mobile shows up
on the right, with an "X" as "Close". This was from design.

Technique uses content selectors to ensure everything can be done
through CSS to keep code out of confusing JS.


static/styles/_components/_panel.scss
```diff
@@ -37,6 +37,7 @@
   background: white;
   margin-top: 3px;
   padding: .65rem 1.5rem;
+  text-align: right;
```
noahmanger: This is aligning everything on the mobile panel to the right.

screen shot 2015-08-04 at 9 42 13 am

noahmanger:
Looks good. Can you fix merge conflicts and then I'll merge?

PR #411, opened 2015-08-04
Fix check for URL change.

### LINKED ISSUE #368: Back button behavior broken on browse pages
The browser's back button isn't working properly on browse pages. If you come from a different page to a browse page, there's now way to get back. Probably something to do with `pushState()`?

jmcarp:
[Resolves #368]


noahmanger:
This fixes most of it but I caught one thing that still doesn't work: if you're on a committee page Receipts or Disbursements tab and click "View all Receipts / Disbursements" you can't go back.

Everything else looks good though.

jmcarp:
Good catch! Patch updated.

msecret:
The JS looks great

PR #413, opened 2015-08-04
Add persistent URLs for tab items.

### LINKED ISSUE #347: Persistent URLs for sub-pages
Pages with multiple tabs currently don't have persistent URLs for each tab. So, for example, if a user browses to a committee detail page, clicks on the Filings tab, and then refreshes, she won't return to the Filings tab automatically. I propose to add per-tab URLs using `pushState`, like we're currently doing for the data table URLs (e.g. `/committees?party=REP`).

jmcarp:
[Resolves #347]

Note: This patch adds persistent URLs to tabs on the candidate and committee detail pages (e.g. `/committee/C00431445?tab=filings`). It does *not* add URLs for more granular content, such as sub-tabs ("Compare by State", "Compare by Employer"), sort rules, map selections, etc. If we want more granular URLs, we can file additional issues.

Pinging @noahmanger for review, @onezerojeremy for thoughts on finer URL control (re.: would users like more granular controls?).


noahmanger:
The URLs stay persistent if you refresh, but not if you change the cycle select. The page still defaults to the first tab.

jmcarp:
Patch updated. There are additional fixes in #411 that need to be merged for this to work completely correctly.

templates/candidates-single.html
```diff
@@ -64,9 +64,9 @@ <h5 class="entity__term__label">Elections</h5>
     <div class="page-controls__bottom">
       <div class="container">
          <nav class="page-tabs">
-            <ul role="tablist">
-              <li role="presentation" class="page-tabs__item"><a role="tab" tabindex="0" aria-controls="panel1" aria-selected="true" href="#section-1">Financial Summary</a></li>
-              <li role="presentation" class="page-tabs__item"><a role="tab" tabindex="-1" aria-controls="panel2" href="#section-2">Filings</a></li>
+            <ul role="tablist" name="tabs">
```
noahmanger: `name` shouldn't be used on `<ul>` and `<li>`s . Probably should be `data-name=` ?

jmcarp:
@noahmanger thanks, fixed!

PR #421, opened 2015-08-04
Arrow button on rows for panel

msecret:
This UI is to make it easier to understand how to open the table
rows to show the details panel. A new table column was used to
contain the button so that everything fits within data tables
width interface. Additional styles were set from the table scss
as its generally safe to ensure these styles for any icon in the
table.

@noahmanger whats the best way to make the arrows as thick as in
the mockup? Right now they are thinner.

### DIFF
--- a/static/js/pages/receipts.js
+++ b/static/js/pages/receipts.js
@@ -13,7 +13,7 @@ var columns = [
     data: 'contributor',
     orderable: false,
     className: 'all',
-    width: '27%',
+    width: '30%',
     render: function(data, type, row, meta) {
       if (data) {
         return tables.buildEntityLink(data.name, '/committee/' + data.committee_id, 'committee');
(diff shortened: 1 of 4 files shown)


static/js/pages/receipts.js
```diff
@@ -40,10 +40,10 @@ var columns = [
     }
   },
   {
-    width: '5%',
+    width: '3%',
     orderable: false,
     render: function(data, type, row, meta) {
-      return '';
+      return '<i class="icon ti-angle-right"></icon>';
```
noahmanger: Wrong close tag

noahmanger:
Don't worry about that. We'll figure it out later with the other visual updates.

msecret:
The tests failed in this case because npm failed to download some dependencies:

```
-npm ERR! fetch failed https://registry.npmjs.org/commander/-/commander-2.8.1.tgz
/npm ERR! fetch failed https://registry.npmjs.org/browserify-aes/-/browserify-aes-1.0.2.tgz
```

noahmanger:
I can't figure out why, but they only show up if your browser is wide enough:

screen shot 2015-08-05 at 9 27 11 am

Maybe something to do with the annoying way datatables sets column widths?

msecret:
@noahmanger maybe. Do I also have to add a class to ensure they show up on all breakpoints?

noahmanger:
Oh, I think the issue is that the width property on line 43 was too small. If I set it at 5%, it's visible on desktop, but not on mobile. If I set it to 10%, it's visible on all, but then takes up too much space.

noahmanger:
And also you need to add `className: 'all'`

PR #422, opened 2015-08-05
Use DataTables for contribution size table.

### LINKED ISSUE #398: Use Datatables for contribution size aggregate table
Any objections to re-implementing this table...

screen shot 2015-08-03 at 2 52 29 pm

...to use Datatables (like so):
screen shot 2015-08-03 at 2 56 56 pm

This is the way it is rendered on election pages, and I think it makes sense to have all of our aggregate tables use the same component. (My apologies for giving it a different style in the first place. It was a style holdover before we started using Datatables for everything).

jmcarp:
[Resolves #398]

### DIFF
--- a/static/js/pages/committee-single.js
+++ b/static/js/pages/committee-single.js
@@ -34,18 +34,7 @@ var sizeColumns = [
     width: '50%',
     className: 'all',
     render: function(data, type, row, meta) {
-      var info = sizeInfo[data];
-      var anchor = document.createElement('a');
-      anchor.textContent = info.label;
-      var uri = URI('/receipts')
-        .query({
-          committee_id: row.committee_id,
-          min_amount: info.limits[0],
-          max_amount: info.limits[1]
-        });
-      anchor.setAttribute('href', tables.buildAggregateUrl(uri, row.cycle));
-      anchor.setAttribute('title', 'View individual transactions');
-      return anchor.outerHTML;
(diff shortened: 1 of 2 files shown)


static/js/pages/committee-single.js
```diff
@@ -12,6 +12,55 @@ var helpers = require('../modules/helpers');
...
+      var info = sizeInfo[data];
+      var anchor = document.createElement('a');
+      anchor.textContent = info.label;
+      var uri = URI('/receipts')
+        .query({
+          committee_id: row.committee_id,
+          min_amount: info.limits[0],
+          max_amount: info.limits[1]
+        });
+      anchor.setAttribute('href', tables.buildAggregateUrl(uri, row.cycle));
```
noahmanger: I didn't specify this before, but this should be on the `total` column, not the `size` column, to keep it consistent with the others tables.

noahmanger:
Looks good other than that comment and the merge conflicts.

jmcarp:
@noahmanger fixed.

noahmanger:
This works by and large, so I'll merge. But I caught at least one case where there's inconsistency: 

The ones from this page aren't working: http://localhost:3000/committee/C00413914?cycle=2016&tab=receipts

noahmanger:
It seems to be an issue with either the data or the filters rather than with this code.

PR #432, opened 2015-08-05
Feature/links to elections

noahmanger:
This adds links from candidate pages to their election pages
I also added the cycle to the election page heading.


noahmanger:
Looks like the same probabilistic failures :(

jmcarp:
@noahmanger I moved the logic for building election URLs into a plain Python function in b3188e75b093be9ec8cdef428a016ba909c4296c.

noahmanger:
:+1: I realized that after I wrote it.

PR #439, opened 2015-08-06
Feature/updating receipts tab

### LINKED ISSUE #339: Style Contributions tab
Once the functionality is implemented, make sure the tab is styled correctly.

noahmanger:
With receipt aggregates now separated by those from individuals and those from committees, this updates the layout of the receipts tab to reflect the division:

screen shot 2015-08-06 at 10 44 43 am

This will resolve #339


templates/partials/receipts-tab.html
```diff
@@ -5,9 +5,17 @@
 <section class="page-section" id="section-2" role="tabpanel" aria-hidden="true" aria-labelledby="section-2-heading">
   <div class="container">
     <div class="page-section__intro">
-      <h2 class="section-heading" id="section-2-heading">Analyze Receipts {% include 'partials/cycle-select.html' %}  </h2>
-      <div class="toggles section__intro__left">
-        <span class="label">Compare by:</span>
+      <div class="section__intro__left">
```
msecret: The "left" part sounds more like a modifier, so should probably be `--left`. If there are too many places this would have to change, I don't care that much.

msecret:
Some of these selenium failures *may* seem valid. Like this one:

```
 AssertionError: 'Oops: We messed up' != 'A WHOLE LOT OF PEOPLE FOR GRIJALVA CONGRESSIONAL COMMITTEE'
E       - Oops: We messed up
E       + A WHOLE LOT OF PEOPLE FOR GRIJALVA CONGRESSIONAL COMMITTEE
```

It seems like we're changing the text that appears for when something goes wrong.

Also, "WHOLE LOT OF PEOPLE FOR GRIJALVA CONGRESSIONAL COMMITTEE" is an awesome committee name.

But this ^ could also mean that search results aren't returning correctly because it can't find `.tst-search_results h5 a` elements or `tbody tr`.

templates/partials/receipts-tab.html
noahmanger: Hmmm. In this case I meant it as the left-side element on the section intro. I don't mind either way.

noahmanger:
I think the issue with that failure is that it's showing the error message ("Oops: we messed up") rather than the committee result. Based on all the other failures, it seems like it's just not getting any data.

msecret:
I'm fine merging then, but @jmcarp this is a potential example of the tests failing randomly.

PR #441, opened 2015-08-07
Load data for tabs on tab selection.

### LINKED ISSUE #345: Load API data lazily
At the moment, all the widgets on a given page either load data from the API server-side, or start loading on the client upon page load. This means we're often loading far more data than the user is likely to look at. We should load data only when necessary (e.g., when the user clicks to the relevant tab or pane). I'm planning to implement this once the latest batch of features is merged in, especially asset splitting, as described in https://github.com/18F/openFEC/issues/908.

### LINKED ISSUE #359: Fix table header bug
I can't figure out why the table header is being set to 100px wide:

screen shot 2015-07-23 at 10 17 54 am

jmcarp:
Add the `onShow` helper, which creates a one-time event listener to do
some work when the specified tab is selected. Also add the
`initTableDeferred` helper, which defers table initialization until the
parent tab is displayed.

@noahmanger this coincidentally fixed #359, which seems to be related to rendering tables in hidden tabs. Now that we don't render the table until the tab is displayed, that behavior goes away.

[Resolves #345]
[Resolves #359]

### DIFF
--- a/static/js/modules/tables.js
+++ b/static/js/modules/tables.js
@@ -6,6 +6,7 @@ var $ = require('jquery');
 var URI = require('URIjs');
 var _ = require('underscore');
 var moment = require('moment');
+var tabs = require('../vendor/tablist');
 
 require('datatables');
 require('drmonty-datatables-responsive');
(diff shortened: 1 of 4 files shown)


noahmanger:
Excellent. This works for me. Nice that it solves #359 -- so weird that Datatables calculates the column widths the way it does. 

I'll let @msecret review the JS but I like it.

PR #442, opened 2015-08-07
Show incumbent challenge and party in elections data table.

jmcarp:
Related to #1106. Requires merge of
https://github.com/18F/openFEC/pull/1108.


noahmanger:
Let's just sort by "Total Receipts" in the first table and then this will be good to go.

jmcarp:
Updated @noahmanger

PR #443, opened 2015-08-07
Avoid hard-coded dependencies on test data where possible.

jmcarp:
Several of our tests assume that certain candidates and committees will
exist, which can cause spurious failures when the test subset changes.
This patch removes a few of these hard-coded dependencies, although
there's more we could do to decouple these tests from the data.

### DIFF
--- a/tests/selenium/search_results_test.py
+++ b/tests/selenium/search_results_test.py
@@ -15,17 +15,17 @@ class SearchResultsPageTests(SearchPageTestCase):
         self.url = furl.furl(self.base_url)
 
     def test_search_results_page_load(self):
-        self.url.args.update({'search': 'boehner'})
+        self.url.args.update({'search': 'john'})
         self.driver.get(self.url.url)
         self.assertIn('Search results', self.driver.title)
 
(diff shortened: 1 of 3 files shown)


msecret:
I understand the part where you're ensuring a draw happens before checking something, but I don't completely understand why you're using "john" now for "boehner" and what that helps. What does that have to do with hard-coded dependencies?

jmcarp:
So the changes in `single_committee_page_tests.py` and `single_candidate_page_tests.py` actually get rid of dependencies on specific items in the test subset, since they're clicking on the first committee and candidate in the browse results, respectively. The other changes are basically hacks: there are lots of candidates whose names include "john" and lots of committees whose names include "america", so those terms will work most test subsets (including the current subset), whereas searching for "boehner" or "grijalva" will only work if the subsets include those specific entities. But now that you mention it, we can do better here. I'll update that part of the patch in a minute.

jmcarp:
Updated to remove more dependencies--we now test search by searching for candidates and committees that we pull from the browse views.

PR #450, opened 2015-08-10
Update query parameters on clearing a filter.

### LINKED ISSUE #438: Inconsistent address updating
Perhaps related to the recent fixes to the browser history stuff, but I'm seeing inconsistent behavior when changing filters. Sometimes the params in the address bar change and sometimes they don't, but in both cases the filters seem to work as expected.

jmcarp:
* Clear query parameters on removing search terms
* Add regression test; make tests more robust

[Resolves #438]

### DIFF
--- a/static/js/modules/filters.js
+++ b/static/js/modules/filters.js
@@ -88,20 +88,24 @@ var bindFilters = function() {
     });
 };
 
+function getFields() {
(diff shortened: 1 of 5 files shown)


noahmanger:
+1. Works for me. But @msecret should probably review the js.

Issue #354, opened 2015-07-22
Implement location look up tool


jenniferthibault:
Most recent mockups here, and discussion in #313 :
![district 1 copy 4](https://cloud.githubusercontent.com/assets/11636908/8836594/ee2b521a-308f-11e5-9c35-8b8626c79cd4.png)
![district 1 copy 5](https://cloud.githubusercontent.com/assets/11636908/8836595/ee2c5368-308f-11e5-84b8-e753bc00b0c2.png)

jmcarp:
@jenniferthibault @noahmanger do we want an option to select an election cycle here, or is the idea to restrict to upcoming elections?

noahmanger:
Good question. I think it's ok to leave it off. Because the results are actually Districts, I think it makes sense to not have the option. You can view past cycles by going to an election page and changing the cycle there. 

I also like the simplicity of having this just be a 1 or 2 question tool. We may learn that people want a cycle select, but I think it's ok to not have it for now.

jmcarp:
This exists.

PR #451, opened 2015-08-10
[WIP] Feature/election lookup

jmcarp:
Prototype of #354.


gulpfile.js
```diff
@@ -117,3 +118,9 @@ gulp.task('build-js', function() {
   .bundle()
   .pipe(callback('static/js/common.js'));
 });
+
+gulp.task('test', function() {
+  return mochify('./tests/unit/**.js')
```
msecret: I was thinking we'd keep dir organization here mirrored with the actual js. SO in this case, we'd have `tests/unit/modules` or `tests/unit/pages`. What do you think of that?

tests/unit/election-lookup.js
```diff
@@ -0,0 +1,87 @@
...
+var sinonChai = require('sinon-chai');
+var expect = chai.expect;
+chai.use(sinonChai);
+
+var $ = require('jquery');
+var _ = require('underscore');
+
+var ElectionLookup = require('../../static/js/pages/election-lookup').ElectionLookup;
+
+_.extend(window, {
```
msecret: This should eventually be moved into a location and then all tests will import it.

gulpfile.js
jmcarp: Agreed. I separated the election lookup logic into a resuable module (modules/election-lookup) and a page that imports and instantiates the module (pages/election-lookup), so the tests have moved to tests/unit/modules/election-lookup.

tests/unit/election-lookup.js
jmcarp: Agreed--I was thinking of deferring this until we have more tests.

jmcarp:
Updated with style improvements from @noahmanger. Want to merge now for testing tomorrow? If not, I can deploy this branch directly to the dev server.

noahmanger:
Looks great. Let me know when you push it up.

PR #452, opened 2015-08-10
Original image button on panel

msecret:
This adds the button for the pdf image of the current thing on the
panel. This uses special jquery code to set the button href to the
pdf from the receipt/whatever data because the button code is not
part of the actual template being rendered. There could be a lot of
unnecessary duplication if this part of the UI was moved into the
template, so I found it easier to keep it outside with this special
code.


noahmanger:
Looks good. Can you just remove the "Image" line from the body of the panel so it's not duplicated?

PR #456, opened 2015-08-11
Fix typo.

jmcarp:
Spelling.


msecret:
Thanks for that, I think that typo was my doing.

Any idea why the tests are failing here? I don't think it's related to your change. I was thinking we could try and figure out before merging.

PR #461, opened 2015-08-12
Set Sauce screen resolution.

jmcarp:
Attempt to fix a few more Sauce + Selenium tests.


jmcarp:
¡Wow! @noahmanger @msecret 
screen shot 2015-08-12 at 10 13 26 am

PR #464, opened 2015-08-12
Use consistent color bars for totals columns.

### LINKED ISSUE #460: Use color bars consistently for aggregate data tables
Right now, some of the aggregate data tables use color bars to indicate contribution totals, and some don't. I think all the aggregate tables should use bars.

With:
screen shot 2015-08-11 at 6 17 48 pm

Without:
screen shot 2015-07-15 at 11 15 54 am

Does that sound right @noahmanger? If so, I can knock this out in a few minutes.

jmcarp:
* Use color bars for all aggregate tables on committee page
* Move duplicated color bar code into helper

[Resolves #460]

### DIFF
--- a/static/js/pages/committee-single.js
+++ b/static/js/pages/committee-single.js
@@ -22,23 +22,6 @@ var tableOpts = {
   useHideNull: false
 };
 
-function buildTotalLink(getParams) {
-  return function(data, type, row, meta) {
-    var span = document.createElement('div');
-    span.setAttribute('data-value', data);
-    span.setAttribute('data-row', meta.row);
-    var link = document.createElement('a');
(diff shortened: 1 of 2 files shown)


static/js/pages/committee-single.js
```diff
@@ -22,6 +22,23 @@ var tableOpts = {
   useHideNull: false
 };
 
+function buildTotalLink(getParams) {
```
noahmanger: Should this be in tables.js? What if we want to use it on election pages?

jmcarp: Agreed, will move.

jmcarp:
Updated according to review @noahmanger

PR #465, opened 2015-08-12
Feature/pretty filings

LindsayYoung:
This works for committee pages and filings. 

Need to add candidate, the API accepts candidate ID so that should be a minor lift.

### DIFF
--- a/static/js/pages/committee-single.js
+++ b/static/js/pages/committee-single.js
@@ -5,12 +5,12 @@
 var $ = require('jquery');
 var URI = require('URIjs');
 var _ = require('underscore');
-var tabs = require('../vendor/tablist');
 
 var maps = require('../modules/maps');
 var events = require('../modules/events');
 var tables = require('../modules/tables');
 var helpers = require('../modules/helpers');
+var decoders = require('../modules/decoders');
 
 var singlePageTableDOM = 't<"results-info results-info--bottom meta-box"frip>';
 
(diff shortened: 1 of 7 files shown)


LindsayYoung:
This should work better with staging data there are some candidate_ids missing from the dev data. 

Here is an example of a candidate that works http://localhost:3000/candidate/S4MD00111?tabs=filings

static/js/pages/committee-single.js
```diff
@@ -159,20 +159,33 @@ var occupationColumns = [
...
       anchor.setAttribute('target', '_blank');
       return anchor.outerHTML;
     }
   },
-  {data: 'amendment_indicator', className: 'min-desktop'},
-  {data: 'report_type_full', className: 'min-desktop'},
+  {
+    data: 'amendment_indicator',
+    className: 'min-desktop',
+    render: function(data, type, row, meta) {
```
jmcarp: This function is repeated below. How about putting it in another file (maybe `helpers.js`) and writing a little more concisely:
```javascript
var amendmentMap = {
  A: 'Amended',
  N: 'New'
};
function(data) {
  return amendmentMap[data] || '';
}
```

templates/partials/filings-tab-candidate.html
```diff
@@ -0,0 +1,31 @@
...
+      <h2 class="section-heading" id="section-4-heading">Candidate Filings {% include 'partials/cycle-select.html' %}</h2>
+    <div id="filters" class="meta-box filters--horizontal section__intro__left">
+      <form id="category-filters">
+        {% include 'partials/filters/amendment-indicator.html' %}
+        {% include 'partials/filters/primary-general.html' %}
+        {% include 'partials/filters/report-type.html' %}
+      </form>
+    </div>
+    <div class="section__intro__right">
+      <!-- <span class="label">View the raw data:</span>
```
jmcarp: Can we delete this commented code?

PR #467, opened 2015-08-12
Election lookup patch

noahmanger:
Very light styling on the election look up tool and a couple fixes to make the interaction a little easier to understand.


jmcarp:
Merged into `feature/election-lookup`.

PR #474, opened 2015-08-16
Feature/new-styles

noahmanger:
Apologies for the enormous PR, but it's not as gnarly as it looks. This gets the web-app in shape to use the new styles being developed in fec-style, so we'll want to wait until we start pulling those in before we merge. Once that happens, this branch basically has every page styled with the new styles.

There's four main things this does:
- Tweaks certain classnames to accommodate the refactored css
- Changes the markup on checkboxes and radio buttons to use the new custom checkbox and radio style. This required editing almost every filter partial. 
- Adds several icons
- Changes up the homepage content

There's a few components that still need to be built out / styled (the glossary and navigation dropdown menus) and it will probably take some work to get the tests up to date. But maybe those happen in later PRs.

Also worth noting that it's still quite rough around the edges in certain places, but that's fine while we finish up the design work. 

Dependent on https://github.com/18F/fec-style/pull/6/files


msecret:
I almost feel like we should get fec-style incorporated and merge this in sooner rather then later. This PR will conflict with almost any other PR, and I don't want you to have to redo a bunch of work

jmcarp:
What's the binary file `static/css`?

msecret:
The code for this is looking good

noahmanger:
Oh weird. I'm not sure what the binary css file is.

msecret:
I think we might want to go with `test` rather then `tst` if that makes it more clear. I'm basing this on the fact that @jmcarp was confused what they were, and hoping naming them `test` would clear up confusion like that.

noahmanger:
Is there anything else this needs or can we merge in? We should hold tight until this is merged as it has all the styles: https://github.com/18F/fec-style/pull/6

PR #478, opened 2015-08-17
Use local styles conditionally.

jmcarp:
* Use styles from local fec-style URL if specified
* Add a place for vendor styles


jmcarp:
@noahmanger we should be able to start deleting a bunch of styles from `static/styles` here. Is there anything we want to keep, other than the two top-level scss files? How about themify-icons?

static/styles/fec.scss
```diff
@@ -0,0 +1 @@
+@import "fec-style/scss/styles.scss";
```
msecret: Why is this being done here rather the `/static/styles/styles.scss`.

Also, I'm getting confused when things are referencing the local styles or fec-style.

noahmanger:
Great. Yeah we shouldn't need any of that. And no need to keep themify-icons or the datatables responsive styles

msecret:
So is this so we can conditionally include fec-styles? And then also have local styles to override?

msecret:
@noahmanger are you adding datatables to fec-style?

noahmanger:
The conditional styles is so that you can be developing off fec-styles locally and this pulls it in from a local server, rather than having to merge styles in and pull down via npm.

And I think vendor styles like datatables that are only going to be used here can just be included here, separate from fec-style.

noahmanger:
This works for me. Can you just add a line to the README?

noahmanger:
Just caught something. It looks like maybe the datatables styles are being loaded in after fec-styles, because its now overriding several things that it wasn't before.

jmcarp:
@noahmanger try re-running `npm install` and `npm run build`. I'm wondering whether you installed the styles with npm before you merged your latest changes to fec-style.

noahmanger:
Ooooh, yeah, I was just running them locally because we haven't merged them
in yet.

On Mon, Aug 17, 2015 at 3:47 PM, Joshua Carp <notifications@github.com>
wrote:

> @noahmanger <https://github.com/noahmanger> try re-running npm install
> and npm run build. I'm wondering whether you installed the styles with
> npm before you merged your latest changes to fec-style.
>
> —
> <https://github.com/18F/openFEC-web-app/pull/478#issuecomment-131985297>.
>



-- 
Noah Manger

jmcarp:
Updated: lots of files deleted, and docs updated.

msecret:
@noahmanger would we need access to the sass variables from fec-style in the local scss files? Like lets say web app wants to override panel from the style guide but needs the $primary variable?

noahmanger:
I guess if you wanted to do that it would be helpful, but I think even then I'd say those changes should live in fec-style. But I could be convinced.

templates/layouts/main.html
```diff
@@ -7,8 +7,13 @@
 
     {% include 'partials/meta-tags.html' %}
 
-    <!-- <link rel="stylesheet" href="/{{ assets['dist/styles/styles.css'] }}" /> -->
-    <link rel="stylesheet" href="http://localhost:8080/css/styles.css" />
+    <link rel="stylesheet" href="/{{ assets['dist/styles/styles.css'] }}" />
```
noahmanger: Is this the same thing as line 15?

noahmanger:
I vote for merging this as-is and if we need to figure out a way to share sass variables we can figure it out then.

templates/layouts/main.html
jmcarp: No, this is the vendor / custom styles. Lines 12-16 adds fec-style. We can change the names if you want.

noahmanger: Oh got it. That works for me.

PR #480, opened 2015-08-18
More panels

### LINKED ISSUE #479: Implement details panels on all tables
The "Details Panel" is the component we're using to show data when columns are hidden on smaller screen sizes. This needs to be implemented for the following browse pages:
- [ ] Candidates
- [ ] Committees

noahmanger:
Adds panels to /candidates/ and /committees/ . They still need some work, but this gets them in place.

Merge in after #474 .

Uses the styles in https://github.com/18F/fec-style/pull/8 .

[Resolves #479]


msecret:
Did you check that this works with the existing panels? If so, I'm ready to merge.

PR #491, opened 2015-08-19
Feature/search select

noahmanger:
Adds classnames to dropdown components and also switches search toggle to search select.

(Note: I'm just commenting out the search radios for now because we may want to change them quickly back to radios)

Uses the styles from here: https://github.com/18F/fec-style/pull/17


templates/macros/search.html
```diff
@@ -3,26 +3,38 @@
   {% set size = 'combo--search--large' %}
 {% endif %}
 <form id="{{ location }}-search" action="/" autocomplete="off" class="search__container js-search">
-  <div class="search__radio toggles">
-    <label for="{{ location }}-radio--candidates">
+<!--   <div class="search__radio toggles">
```
jmcarp: Can this block be deleted?

noahmanger: Sure.

PR #500, opened 2015-08-20
Feature/accessibility improvements

noahmanger:
Fixes some classnames and makes the details panel keyboard accessible. 

The table stuff needs more work: now, if you tab past the last item in a panel, it sets focus below the table, rather than the next row.


static/js/modules/tables.js
```diff
@@ -222,39 +222,49 @@ function modalAfterRender(template, api, data, response) {
...
-    // TODO figure way to share these values with CSS.
-    if ($(document).width() < 980) {
-      api.columns('.hide-panel-tablet').visible(false);
-    }
+  $table.find('tr').each(function(){
+    $(this).attr('tabindex', 0);
+  })
+
+  $table.on('click keypress', '.js-panel-toggle tr', function(ev) {
+    if (ev.which === 13 || ev.type === 'click') {
```
msecret: This would be a little cleaner if the code for opening the modal was in it's own function and the code for handling the keypress and click events calls it.

noahmanger: Yeah...this seems like a good candidate for a refactor in general. Mind if I make a separate issue for that?

msecret: The modal could also be it's own class, which is initialized and has methods for various things like `show`, `hide`, `hideColumns`, `populatePdf`, `setFocus`, etc.

noahmanger: Yep. Let's do that.

noahmanger: Captured https://github.com/18F/openFEC-web-app/issues/501

PR #502, opened 2015-08-21
Refactoring dropdown code

### LINKED ISSUE #291: Make dropdowns accessible
From Nick:
- [ ] The name search field doesn’t have an id
 - Based on the label it should be id=‘q’
- Each group of checkboxes is in a field set 
 - [ ] These need legends, for example the legend for the first on should be “Election Years”
 - If you don’t want to use legend, everything will need to be associated with aria
- For the control of the selects
 - [ ] Aria instructions need to be added to the more buttons
 - [ ] Focus is lost when selecting an option in the list. Pressing TAB after selecting one sends you back to the top of the document

noahmanger:
- Refactored dropdown code and moved to a component in fec-style
- Made some small class name changes to accomodate

Depends on https://github.com/18F/fec-style/pull/27

Resolves #291


templates/partials/filters/report-type.html
```diff
@@ -1,211 +1,211 @@
 <div class="filter">
```
msecret: Is there a reason this isn't being done in a loop? Seems like hteres a lot of repetition in this file.

noahmanger: I know, it really should be. Not sure who originally set it up like this, but if someone could just build an array of the report types that would be great. Really all of the filter partials with long lists should be refactored this way.

PR #517, opened 2015-08-25
Feature/checkbox from typeahead

noahmanger:
Makes use of the new filter typeahead from https://github.com/18F/fec-style/pull/40

Creates a new macro for these fields.

### DIFF
--- a/static/js/init.js
+++ b/static/js/init.js
@@ -16,6 +16,7 @@ var glossary = require('fec-style/js/glossary');
 var accordion = require('fec-style/js/accordion');
 var dropdown = require('fec-style/js/dropdowns');
 var typeahead = require('fec-style/js/typeahead');
+var typeaheadFilter = require('fec-style/js/typeahead-filter');
 
 require('jquery.inputmask');
 require('jquery.inputmask/dist/inputmask/jquery.inputmask.date.extensions.js');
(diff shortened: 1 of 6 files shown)


static/js/init.js
```diff
@@ -37,6 +38,16 @@ $('.js-dropdown').each(function() {
   new dropdown.Dropdown(this);
 })
 
+$('.js-typeahead-filter').each(function() {
+    var dataKey = $(this).attr('data-dataset');
+    var datasetMap = {
+        candidates: typeahead.datasets.candidates,
+        committees: typeahead.datasets.committees
+    }
+    var dataset = datasetMap[dataKey];
```
jmcarp: Do we need the `datasetMap` variable? How about `var dataset = typeahead.datasets[dataKey];`?

jmcarp:
I was looking at this for a bit and realized we're going to have to handle this a little differently for performance reasons. For the big tables (receipts and disbursements), filtering on multiple committee IDs turns out to be prohibitively expensive. I think we're going to want to stick with a single filter and leave out the checkboxes. I'm going to do some more research and see if I can find a way to make multiple committees performant, but until then, let's hold off on merging.

noahmanger:
Ah, good looking out. That's fine.

Even if we can't do multiple committees it would be great to have it set up so if there's a committee id applied on page load, it shows up in the filters. (Currently the committee name filter stays empty).

noahmanger:
Closing this in the mean time.

PR #528, opened 2015-08-26
Add filter for unique receipts.

jmcarp:
Based on issue https://github.com/18F/openFEC/issues/1141 and discussion with @noahmanger.


noahmanger:
LGTM. We'll want to add help text / think about wording on this, but thanks for adding.

PR #536, opened 2015-08-28
Feature/election lookup map

jmcarp:
Add a map of congressional districts to the election lookup tool. This will surely require styling and UI improvements, but it works as-is. It looks like this:
screen shot 2015-08-28 at 11 15 12 am


noahmanger:
This works great for me. I think @msecret should do a closer reading of the js though just because there's so much new stuff in there.

PR #549, opened 2015-08-31
Show incumbent officeholders in election search.

jmcarp:
Depends on https://github.com/18F/openFEC/pull/1165.


noahmanger:
Yes! Looks good to me.

jmcarp:
Conflicts resolved. Ready for review @noahmanger @LindsayYoung.

LindsayYoung:
I can take a look

LindsayYoung:
Can we change the wording from "open seat" to "challenger"? That will make more sense in this context

LindsayYoung:
Technically, we should say challenger if there is an incumbent and open seat if there is no incumbent, but calling them all challengers is less awkward that listing "open seat" when it is not.

jmcarp:
@LindsayYoung sure, but that's not part of this patch, and it's a change on the API side--at the moment, we're displaying the candidate status values from FEC verbatim. This patch only touches the election search results:

LindsayYoung:
Sorry to get sidetracked.

jmcarp:
@LindsayYoung not at all--I think it's worth updating the wording. But in another patch :smiley:

PR #560, opened 2015-09-01
New header and footer

noahmanger:
Updates the header and footer with the new seals.

Requires https://github.com/18F/fec-style/pull/52


jmcarp:
The new images shouldn't be checked in here, since they're now part of fec-style. I'll delete and merge.

noahmanger:
Oh right, thanks. I put them there to test locally but forgot to not check them in.

PR #569, opened 2015-09-02
Feature/filings detail

### LINKED ISSUE #399: Add details pane to filings tab table
So it will look like this:

details panel

Note: the actual columns in the table may be diffferent.

### LINKED ISSUE #405: Add details panel to Browse Filings

jmcarp:
* Add detail pane for browse filings and committee filings tabs
* Fetch details for pane by AJAX request, falling back to existing data if necessary
* Only show detail pane for F3* filings

@noahmanger this needs some work on styles: the filings panel is more or less unstyled, and it wasn't immediately clear how to get the panel to display at the correct size on the committee filings page. We'll also have to to think about differences in variable names between Presidential, House/Senate, and PAC/Party filings, since the variable names from the mockup don't fit all three.

[Resolves #399]
[Resolves #405]

### DIFF
--- a/candidate.hbs
+++ b/candidate.hbs
@@ -1,5 +1,5 @@
 <div class="panel__row">
-  <h4>{{report_year}} {{report_type_full}}</h4>
+  <h4>{{document_description}}</h4>
   <h5>Report coverage: {{datetime coverage_start_date}} - {{datetime coverage_end_date}}</h5>
   <table>
     <tr>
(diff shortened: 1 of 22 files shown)


LindsayYoung:
If we are showing everything can we use the respective report partials we already have? The names are tricky and we already spent a lot of time checking those.

jmcarp:
@LindsayYoung we can't use them in their exact form, since we can't mix and match jinja and handlebars templates, but yeah, we can copy over the variable names. But the list of variables we're showing in the mockup doesn't quite match any of the three existing reports templates. @noahmanger, were those variables just placeholders, or are they authoritative?

jmcarp:
Ready for review @noahmanger. Note: this will work slightly better once https://github.com/18F/openFEC/pull/1175, since I harmonized a few variable names there. Working with the panel code reminds me that we should refactor and test it, since it's a little hairy right now, but that can happen in a separate patch, and after launch.

noahmanger:
Oops, sorry I meant to merge yesterday.

jmcarp:
That's cool, it's better now anyway.

PR #571, opened 2015-09-02
Feature/election styles

noahmanger:
Minor styling of election pages (uses https://github.com/18F/fec-style/pull/58)


noahmanger:
Just did some more work on this. The candidate comparison stuff is slightly different from mocks:

- I don't have the toggle to hide the candidate list yet, but I tend to think it's ok. Nice to avoid one more knob on this page, and by the list items horizontally and using the dropdowns we save space on long lists.
- I moved the map / table toggle to the right of that div and added a line of explanation

screen shot 2015-09-02 at 6 38 19 pm

PR #576, opened 2015-09-03
adding url

LindsayYoung:
The view all links on the election search page were not getting built.


LindsayYoung:
Nope, that is too aggressive. Don't merrge

jmcarp:
This doesn't actually change URLs other than in the search results--the URL issue is an unrelated bug that I'll fix while merging. Specifically, if `FEC_CMS_URL` isn't set, URLs that point to the CMS were relative to the current page, which is why you saw "elections" incorrectly prefixed to some URLs.

PR #581, opened 2015-09-03
Feature/committee page styles

noahmanger:
Updates markup slightly on committee and election pages to reflect updated styles. Notable changes:
- Datatables with bars now default to $0 when the value was undefined and the bar now is in a container with a different color:

screen shot 2015-09-03 at 3 46 02 pm

- Toggle buttons are now in the ivory "filter boxes" on election pages and committee tabs
- Moves the "Between committee" transaction totals to their own tab

Uses the styles from https://github.com/18F/fec-style/pull/62


msecret:
Code looks good.

noahmanger:
Retrying the tests. Looks like a fluke.

PR #585, opened 2015-09-04
Feature/election lookup styles

noahmanger:
Markup updates for the election lookup to make use of https://github.com/18F/fec-style/pull/64

There's still a few things on [the mockup](https://invis.io/9N43H4YBJ) that need to be implemented, but they're more in the JS so @jmcarp we should talk about these pieces.

screen shot 2015-09-03 at 9 47 31 pm

### DIFF
--- a/static/templates/electionResult.hbs
+++ b/static/templates/electionResult.hbs
@@ -1,12 +1,21 @@
 <div>
   {{#each this}}
-  <div class="election-result box">
-    <h4 style="color: {{this.color}}">{{this.electionName}}</h4>
-    <div>Next general election: {{this.electionDate}}</div>
+  <div class="result">
+    <h3 style="color: {{this.color}}"><a href="{{this.url}}">{{this.electionName}}</a></h3>
+    <div class="row">
+      <span class="result__term">Next general election:</span>
+      <span class="result__definition">{{this.electionDate}}</span>
+    </div>
     {{#if this.incumbent}}
-    <div>Current officeholder: <a href="{{this.incumbent.url}}">{{this.incumbent.name}}</a></div>
+    <div class="row">
+      <span class="result__term">Current officeholder:</span>
+      <a class="result__definition" href="{{this.incumbent.url}}">{{this.incumbent.name}}</a>
+      </div>
     {{/if}}
-    <div>Current candidates: <a href="{{this.url}}">View all &raquo;</a></div>
+    <div class="row">
+      <span class="result__term">Current candidates:</span>
+      <a class="result__definition" href="{{this.url}}">View all &raquo;</a>
+    </div>
   </div>
   {{/each}}
 </div>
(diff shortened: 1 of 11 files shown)


templates/election-lookup.html
```diff
@@ -1,68 +1,70 @@
...
+  <span class="page-header__title">Find candidates and elections by location</span>
+</header>
+<section class="main" id="election-lookup" >
+  <div class="container">
+    <h2 class="t-ruled--bottom t-ruled--bold">Search</h2>
+    <div class="search-controls">
+      <form>
+        <div class="search-controls__row">
+          <label class="label" for="cycle-select">Choose an election cycle</label>
+          {{ select.cycle_select(cycles, cycle) }}
```
jmcarp: The cycle select widget changes the URL and refreshes the page on change, which we don't want to do here. I'll fix during review.

templates/macros/cycle-select.html
```diff
@@ -1,6 +1,6 @@
-{% macro cycle_select(cycles, cycle, location='query') %}
+{% macro cycle_select(cycles, cycle, location='query', id='cycle-select') %}
 {% if cycles %}
-  <select class="cycle-select select--neutral js-cycle" aria-label="Two-Year Period" name="cycle" data-cycle-location="{{ location }}">
+  <select id="{{ id }}" class="cycle-select select--neutral js-cycle" aria-label="Two-Year Period" name="cycle" data-cycle-location="{{ location }}">
```
jmcarp: Do we need the `id` here for accessibility reasons? It doesn't look like we're using it in the code.

static/templates/electionResult.hbs
```diff
@@ -1,12 +1,21 @@
 <div>
   {{#each this}}
-  <div class="election-result box">
-    <h4 style="color: {{this.color}}">{{this.electionName}}</h4>
-    <div>Next general election: {{this.electionDate}}</div>
+  <div class="result">
+    <h3 style="color: {{this.color}}"><a href="{{this.url}}">{{this.electionName}}</a></h3>
```
jmcarp: After these changes, the per-district colors aren't showing up anymore. I think that's fine, since the design called for color icons instead of color text--I'll file a separate issue for that.

jmcarp:
I noticed that one of the javascript unit tests failed, but the Travis build didn't notice. Fixed the test and the Travis issue on merge.

templates/macros/cycle-select.html
noahmanger: I added it so we could explicitly assign a `<label>`. We don't always do this, but it's nice to have there if we need it, imo.

noahmanger: Though maybe a non-issue now if we're using a different cycle select on the election lookup page.

jmcarp: Cool, makes sense. I kept the same cycle select for the election lookup page, but disabled the URL change.

PR #593, opened 2015-09-04
Use list of election districts from 18F/1179.

### LINKED ISSUE #587: Deal with re-districting
Using Pelosi as an example: https://fec-dev-web.18f.gov/candidate/H8CA05035

It looks like her district had a different number until 2012, because clicking on any election year link in the header takes you to an election page without her, presumably because the 2010 re-districting drew her district different / with a new number.

I know we've talked about this in the past, but we should fix it here.

jmcarp:
Depends on https://github.com/18F/openFEC/pull/1179.

[Resolves #587]

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -86,12 +86,12 @@ def cycle_end(value):
     return datetime.datetime(value, 12, 31)
 
 
-def get_election_url(candidate, cycle):
+def get_election_url(candidate, cycle, district=None):
     return url_for(
         'elections',
         office=candidate['office_full'].lower(),
         state=candidate['state'] if candidate['state'] != 'US' else None,
-        district=candidate['district'],
+        district=district or candidate['district'],
         cycle=cycle,
     )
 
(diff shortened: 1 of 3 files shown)


jmcarp:
Now that https://github.com/18F/openFEC/pull/1179 is merged, this is ready for review @noahmanger.

noahmanger:
This works as expected, except it looks like the district maps still reference the current district boundaries, rather than what they were at the time. Two ways I could see to deal with this:

1. We simply hide the district maps on election pages for anything before 2012 (the first year of the current districting), or
2. Get all the past congressional district maps. 

Maybe do option 1 for launch and 2 for later?

jmcarp:
Right, this patch only touches the links on the candidate district page. Hiding maps for older cycles is probably best for now, but we should dig up older districts maps before too long. Also, zip code search is going to be a pain for older elections.

noahmanger:
Ugh yeah. Ok, I'll merge this and make a separate issue for hiding maps on older pages.

Issue #591, opened 2015-09-04
Add election lookup to Data Home


noahmanger:
[Full mock here](https://invis.io/SE3ZFQF8Y). Specifically this part:

![screen shot 2015-09-04 at 1 03 40 pm](https://cloud.githubusercontent.com/assets/1696495/9694116/ad8f5740-530a-11e5-8537-55548a574aef.png)

If @jmcarp gets the basic functionality there, I can style.

jmcarp:
For the right-hand panel, I can hard-code this to look up the 2016 presidential election for now, but what should this do after the election next year? If we switch to the 2020 presidential election after the 2016 election, this widget will be pretty empty for a few years.

Thoughts @noahmanger?

noahmanger:
Good question. Curious what @jenniferthibault thinks, but I think in the future it could disappear, or be a sort of retrospective. There's always a lot of reporting around the total numbers post-election, so I think that could go on for a while.

jmcarp:
In any case, sounds like we can hard-code to the 2016 Presidential election for now. I'll do that next unless folks want to discuss more--let's just remember to make a decision about the long-term behavior before the election.

noahmanger:
:+1: I've got the placeholder markup for it here: https://github.com/18F/openFEC-web-app/pull/605

jenniferthibault:
Nice! I love the idea of a retrospective for a certain amount of time after the election, (maybe a year, even?) which will be relevant to people researching. 

@amykort @jwchumley @paulclark2 Any thoughts on what you'd like to be able to show users here in the future that would be similarly useful and interesting? Feel free to pitch a few ideas!

PaulClark2:
One option would be to switch to house or senate candidates and numbers for 2017-2018.  A natural time switch would be following the 2017 April Quarterly reports.  They are due 4/15/2017. These reports will be the first financial reports for house and senate candidates for the 2017-2018 two-year period.

jmcarp:
Agree with @PaulClark2 about switching to Congressional candidates in 2017, but not sure we want to implement that piece now. Thoughts @noahmanger?

noahmanger:
Yeah, I agree that can wait.

On Fri, Sep 11, 2015 at 9:33 AM, Joshua Carp <notifications@github.com>
wrote:

> Agree with @PaulClark2 <https://github.com/PaulClark2> about switching to
> Congressional candidates in 2017, but not sure we want to implement that
[...]

PR #595, opened 2015-09-08
Feature/update election lookup

### LINKED ISSUE #592: Finish remaining features on election lookup
A few remaining features to add to the election lookup page:
- [x] Add URL updating on searches
- [x] Show the full list of candidates in a state upon selecting a state
- [x] Make districts on the map have a stroke width of 1px
- [x] Show a message when more than on district match a zipcode

And a thought: should the election map first just display state outlines? And then if you select a state it zooms in and shows the districts? Would definitely make the map less cluttered.

[Full mockup here](https://invis.io/9N43H4YBJ)

jmcarp:
* Add persistent URLs for election searches
* Show results for states without district selection
* Show warning on multiple districts per zip code

Note: the last item from #592 is addressed in 18F/fec-style/pull/69.

[Resolves #592]


jmcarp:
@noahmanger: the warning for zip codes with multiple districts is basically unstyled--can you take a look? To test, try 60442 (four districts!) or 19041 (two).

noahmanger:
This is working great. The only regression I caught is that the list of districts no longer updates based on the districts in that state.

jmcarp:
Should be fixed now @noahmanger. I also added the left half of #591 for review.

noahmanger:
Beautiful. I'll style the component on the home page separately.

PR #599, opened 2015-09-08
Feature/star incumbents

noahmanger:
Adds an option to include a star next to incumbents in datatables.

I know there's some data cleanup that needs to happen for this to be accurate, but wanted to get this at least into a PR while we sort that out.


static/js/modules/tables.js
```diff
@@ -70,13 +70,18 @@ function buildCycle(datum) {
...
-function buildEntityLink(data, url, category) {
+function buildEntityLink(data, url, category, incumbent) {
   var anchor = document.createElement('a');
   anchor.textContent = data;
   anchor.setAttribute('href', url);
   anchor.setAttribute('title', data);
   anchor.setAttribute('data-category', category);
   anchor.classList.add('single-link');
+
+  if (incumbent === 'Incumbent') {
```
jmcarp: Prefer a boolean flag like `isIncumbent` over a string variable here. I'll change during merge.

PR #602, opened 2015-09-09
Content stylings

emileighoutlaw:
Tightens a small portion of content. Restyles a variety of headers, subheads, and fields to adhere to sentence case capitalization

### DIFF
--- a/templates/committees-single.html
+++ b/templates/committees-single.html
@@ -42,7 +42,7 @@
                 {% if candidates|length > 1 %}
                   candidates
                 {% else %}
-                  Candidate
+                  candidate
                 {% endif %}
               </span>
               <ul class="entity__term__data">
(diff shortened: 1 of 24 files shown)


templates/committees-single.html
```diff
@@ -40,7 +40,7 @@ <h1 class="entity__name">{{ name }}</h1>
             <li class="entity__term">
               <span class="entity__term__label">Related
                 {% if candidates|length > 1 %}
-                  Candidates
+                  candidates
                 {% else %}
                   Candidate
```
noahmanger: We should also lowercase the singular Candidate

emileighoutlaw: Good catch!

emileighoutlaw:
Thanks @noahmanger ! :D

Issue #360, opened 2015-07-23
Order "Report type" options


noahmanger:
And maybe think of a better way to group these or something. Not sure at the moment.

![screen shot 2015-07-23 at 10 22 52 am](https://cloud.githubusercontent.com/assets/1696495/8857193/dd135faa-3124-11e5-8e46-70421166f1da.png)

noahmanger:
Update: I see that they're actually grouped in a logical way (all monthly reports are together, for instance). But maybe those should be broken out into their own dropdown?

noahmanger:
Is this still something we want to do? If so @PaulClark2 could you share the order you shared earlier? I can't find it.

PaulClark2:
@noahmanger  will also send to you in Slack DM

**Form Types Selections**:  
Statements Of Organization - Form 1
Multicandidate Status - Form 1M
Statements Of Candidacy - Form 2
Congressional Candidate Financial Reports - Form 3
Presidential Financial Reports - Form 3P
PAC and Party Financial Reports - Form 3X
Bundled Contributions Reports - Form 3L
Convention Financial Reports - Form 4
Independent Expenditure Reports and Notices - Forms 5 & 24
Contributions and Loans Notices - Form 6
Communication Cost Reports - Form 7 
Debt Settlement Plans - Form 8
Electioneering Communications Notices - Form 9
Inaugural Committee Donation Reports - Form 13
Miscellaneous Submissions
Requests for Additional Information
Expenditure of Personal Funds Notices - Form 10
Opposition Personal Funds Notices - Form 11
Suspension of Increased Limits Notices - Form 12

**Report Type Selections** (including or excluding the report code)
Monthly Reports
"M2" February Monthly
"M3" March Monthly
"M4" April Monthly
"M5" May Monthly
"M6" June Monthly
"M7" July Monthly
"M8" August Monthly
"M9" September Monthly
"M10" October Monthly
"M11" November Monthly
"M12" December Monthly
"YE" Year-End

Quarterly Reports
"Q1" April Quarterly
"Q2" July Quarterly
"Q3" October Quarterly
"YE" Year-End

Semi-annual Reports 
"MY" Mid-Year
"YE" Year-End

Election Sensitive Reports
"12P" Pre-Primary
"12C" Pre-Convention
"12G" Pre-General
"12R" Pre-Runoff
"12S" Pre-Special
"30G" Post-General
"30R" Post-Run-Off
"30S" Post-Special
"30P" Post-Primary
[...]

PR #604, opened 2015-09-09
Re-ordering reports and forms

noahmanger:
Re-orders the report dropdowns to the order described in #360 

Update:
- Breaks report types out into separate filters per the discussion in #360 
- Instead of showing all the new filters on the filings tab on the committee page, this removes those filters altogether, further differentiating this tab from the main filings browse page (something we learned may be helpful from user testing)

screen shot 2015-09-10 at 10 49 54 am

screen shot 2015-09-10 at 10 53 02 am


jmcarp:
These codes don't match the order or content of #360:
* The 10* codes should be in the "Election Sensitive Reports" section
* The MSA and MSY codes have the wrong text (both of those codes also have the same text in our version, which is confusing)
* We're using abbreviations instead of full terms for codes like ADJ (we say "COMP ADJUST AMEND"; #360 says "Comprehensive Adjusted Amendment").

I also see that code 60D is described as post-election in #360 but post-convention here (and in http://www.fec.gov/finance/disclosure/metadata/ReportTypeCodes.shtml). @PaulClark2, can you clarify the description of code 60D?

I think we should use the text and order from #360 verbatim here. Thoughts @noahmanger @LindsayYoung?

PaulClark2:
Agreed we should use #360 and we should avoid abbreviations if possible. 

"60D' should be Post-Convention

For the bundling reports:  

"MSA"	Monthly Semi-Annual (MY)
"MSY"	Monthly Semi-Annual (YE)

We could use the following descriptions if they are clearer.  We could also drop the parenthetical

"MSA"	Monthly Semi-Annual Mid-Year (MY)
"MSY"	Monthly Semi-Annual Year-End (YE)

noahmanger:
Ah, thanks for catching @jmcarp . I wasn't sure if we chose the other labels for a reason. I've updated to more closely reflect that list, but have a few more questions in #360 .

noahmanger:
Updated the main note above with recent changes. I'm pretty sure I have the list captured correctly with the right order and labels, but would appreciate a second set of eyes to make sure I got it right.

jmcarp:
LGTM

Issue #449, opened 2015-08-07
Election page aggregates should link to itemized transactions


noahmanger:
However, I know this is a little tricky because we're using totals from all of a candidate's authorized committees. So flagging this for discussion.

jmcarp:
I proposed a small change to the API that includes a list of committee IDs in the elections endpoint, which gives the webapp the information it needs to fetch transactions for all of a candidate's principal and authorized committees. The receipts and disbursements API endpoints also accept multiple values for `committee_id`, so we can get that data. The tricky part is handling this in the filter interface: we don't currently allow users to filter by multiple committees, even though the API can handle those queries.

noahmanger:
We have everything in place (or about to be in place) in order to implement this now, right? This will be valuable to link to these views from the candidate pages as well.

PR #606, opened 2015-09-09
Link election aggregates to individual transactions.

jmcarp:
Link election aggregates by size, location, and contributor type to itemized receipts.

@noahmanger: did we want to link the totals in the top election table to itemized receipts and disbursements as well? This patch just covers our custom aggregates so far.

Part of #449.

### DIFF
--- a/static/js/modules/columns.js
+++ b/static/js/modules/columns.js
@@ -7,6 +7,14 @@ var _ = require('underscore');
 var tables = require('./tables');
 var decoders = require('./decoders');
 
+var sizeInfo = {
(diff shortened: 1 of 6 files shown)


noahmanger:
Not sure. We hadn't talked about it. That would be nice, but I don't think is critical.

jmcarp:
Easy enough--just added. Ready for review.

noahmanger:
This is working really well. Good call adding the "Show contributions from" filter. I have some questions how to best design this in conjunction with the "Unique only" filter, but will raise that in a separate issue.

I'd like for @msecret to review the js of this before we merge though.

PR #608, opened 2015-09-09
moving superpac filter

noahmanger:
Resolves https://github.com/18F/openFEC/issues/1185


msecret:
The failed tests seem to be saying that `.button--primary` can't be found. Although that doesn't make sense in the context of this PR because you made no modifications to it. Could this be due to a previous PR that made this type of change.

jmcarp:
@msecret the test is fixed in #610.

PR #610, opened 2015-09-09
Feature/election summary

jmcarp:
Add election summary data to homepage. Completes #591.

Branched off https://github.com/18F/openFEC-web-app/pull/605.
Depends on https://github.com/18F/openFEC/pull/1188.

### DIFF
--- a/static/js/modules/election-summary.js
+++ b/static/js/modules/election-summary.js
@@ -8,7 +8,7 @@ var URI = require('URIjs');
 var helpers = require('./helpers');
 
 function buildUrl(opts) {
-  var parts = [opts.office, opts.state, opts.district, opts.cycle]
+  var parts = ['elections', opts.office, opts.state, opts.district, opts.cycle]
     .filter(function(part) {
       return !!part;
     })
--- a/static/js/modules/election-summary.js
+++ b/static/js/modules/election-summary.js
@@ -1,0 +1,43 @@
+'use strict';
+
+/* global require, module */
+
+var $ = require('jquery');
+var URI = require('URIjs');
+
+var helpers = require('./helpers');
+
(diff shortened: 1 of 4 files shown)


static/js/modules/election-summary.js
```diff
@@ -0,0 +1,43 @@
...
+function ElectionSummary(selector, opts) {
+  this.$elm = $(selector);
+  this.opts = opts;
+
+  this.$count = this.$elm.find('.count');
+  this.$receipts = this.$elm.find('.receipts');
+  this.$disbursements = this.$elm.find('.disbursements');
+
+  this.fetch();
+  this.$elm.find('a.button--election').attr('href', this.buildUrl());
```
msecret: Is the `a` in the selector actually needed?

I'd also suggest this should be a js-prefixed class rather then using a styling class, but will defer to you for best judgement.

jmcarp: I think the `a` makes sense, since this behavior can only apply to anchor tags. Agreed re. `js-` prefixed classes; will update.

jmcarp:
Updated!

noahmanger:
The "view all" link in the summary has the wrong URL "http://localhost:3000/president/2016/". Missing "/elections/"

noahmanger:
Other than that looks good to me whenever the api update gets merged. It's a little slow for me running the api locally, but hopefully that's helped when it's not local. If not, may be worth adding a loading animation or something. Or, like, a cool ticker effect.

LindsayYoung:
I think if we are showing candidate totals, we need independent expenditure totals for that race too.

LindsayYoung:
This looks good, I think we need to add independent expenditures to this, but that can be a separate pr.

PR #620, opened 2015-09-10
Release/public beta 20150910

jmcarp:
Something new this time around: auto-generated changelogs! We can and should add real release notes as well, but the auto-generated logs are a good start. See https://github.com/18F/openFEC-web-app/blob/release/public-beta-20150910/CHANGELOG.md.


noahmanger:
This is super old. I'm closing.

Issue #489, opened 2015-08-19
Zip Code Search produces no results for 80306 (Boulder, CO)


onezerojeremy:
URL: https://fec-dev-web.18f.gov/elections/

![screen shot 2015-08-18 at 8 26 03 pm](https://cloud.githubusercontent.com/assets/474916/9346169/8acdd2ca-45e7-11e5-8a04-c1b6b6b4c282.png)

jmcarp:
Interesting--that zip code isn't in the Census database (https://www.census.gov/geo/maps-data/data/cd_national.html), and it doesn't show up in http://congressional-district.insidegov.com/. But it seems to be in the Colorado 2nd district. We can add that lookup manually, which doesn't seem great, or find a better data source. Is there a more authoritative database than the Census @noahmanger @LindsayYoung?

PaulClark2:
Similar issue for Miami, FL.  I get president and senate but no house races

![untitled](https://cloud.githubusercontent.com/assets/11817734/9581008/df75c840-4fca-11e5-92c3-9a854a9e4bfd.png)

PaulClark2:
Look up Alexandria VA 22304 (results as expected)
Next look up Miami, FL 33129 (Map stays focused on Alexandria, VA, no House races displayed)

![untitled](https://cloud.githubusercontent.com/assets/11817734/9581103/67d5c802-4fcb-11e5-96f5-90002d642514.png)

jmcarp:
@PaulClark2 @jwchumley: I believe you mentioned that FEC might already have purchased better zip code / congressional district data. Could you check and see whether you have that data? If not, we'll find something else.

PaulClark2:
@jmcarp I should have updated you. I checked our Data Services tool uses the USPS ZIP files but we don't have access to them.

jwchumley:
Here is a bit of text I grabbed from 
http://www.house.gov/representatives/find/#wrongrep

It is their response when people put in a zipcode on their representative 
finder and the wrong person comes up. Clearly it is an issue for them too.

What should I do when I enter my ZIP code information and I get the wrong 
elected Representative?
The Find Your Representative service matches the ZIP code information you 
provide with a list of congressional districts. If you receive an error 
due to a missing ZIP code or incorrect member information, please use the 
Contact Webmaster form to report the problem. Select the appropriate error 
category (Report an error in the Find Your Representative service.) and 
provide as much information as possible to assist us in researching the 
[...]

PR #623, opened 2015-09-10
Added no zip code help text with test

LindsayYoung:
@noahmanger this doesn't have any extra styling just adds the contact info if we did not find your zip code.


LindsayYoung:
closes #489

noahmanger:
LGTM. Will do a separate PR with styles.

PR #628, opened 2015-09-10
styling the no results message

noahmanger:
Styles the message that shows up if the zip code isn't found. @emileighoutlaw how doest this language work for you? Made a slight adjustment from @LindsayYoung 's original.

screen shot 2015-09-10 at 4 52 13 pm

### DIFF
--- a/static/templates/electionNoResults.hbs
+++ b/static/templates/electionNoResults.hbs
@@ -1,6 +1,6 @@
-<div>
-  Can't find your zip code?
-  Let us know:
-  <a href="mailto:18F-FEC@gsa.gov">email us</a> or
+<div class="message message--alert">
+  <h2>We can't find any results for this ZIP code</h2>
+  <p>Double-check that you entered it correctly, and if it still doesn't work, let us know:</p>
+  <a href="mailto:18F-FEC@gsa.gov">Email us</a> or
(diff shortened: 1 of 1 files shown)


jmcarp:
The language looks good, but I realized there's a minor issue here: it's possible (although pretty unlikely) for users to see this message even when searching by state and district. Occasionally, we run into missing data (e.g., Virgin Islands 1997-98) and show no results. In those cases, saying we couldn't find results for the specified zip code doesn't make sense.

I'm going to condition the warning text on search state: if the user specified a zip code, we'll warn that we couldn't find results "for this ZIP code"; else, that we couldn't find results "for this location". If we want to update that language or behavior, we can file another patch. 

@noahmanger: I'm also going to fix the unit tests on merging. @LindsayYoung added a JS unit test asserting that the expected warning text was added when no results were found, and that test broke when the text changed.

noahmanger:
Sounds good to me. Thanks.

On Fri, Sep 11, 2015 at 6:52 AM, Joshua Carp <notifications@github.com>
wrote:

> Merged #628 <https://github.com/18F/openFEC-web-app/pull/628>.
>
> —
> <https://github.com/18F/openFEC-web-app/pull/628#event-407098559>.
>



-- 
Noah Manger

emileighoutlaw:
I missed the boat, but wanted to pop on and say anyway that this language has my thumbs up.

PR #629, opened 2015-09-10
Ensure mobile nav has aria atts and good focus

msecret:
Requires https://github.com/18F/fec-style/pull/79.

When I was attempting to test this I could not because browserify
complained that it couldn't find my new module. Is there a way around
this? This means that currently, this code hasn't been fully tested.


noahmanger:
You can always target that branch of fec-style in this app's package.json. I'm testing it now that way.

noahmanger:
This looks good. Can you make the same additions to https://github.com/18F/fec-cms ?

I'll merge once we merge in fec-style

PR #641, opened 2015-09-11
Added independent expenditures the spacing is very messed up

LindsayYoung:
@noahmanger I added the total independent expenditures and it seems pretty straight forward but the widget formatting fails for some reason.


noahmanger:
Looks right to me? I made a small style tweak but it should be good even without that.

screen shot 2015-09-11 at 9 57 48 am

PR #642, opened 2015-09-11
Feature/even legend ticks

jmcarp:
* Use legend tick values like 0, 50, 100 instead of 1.17, 47.9, 92.
* Fix legend update on change / add / delete map


noahmanger:
Nice! Looks good. I just caught this though: Can you add dollar signs to the legends?

jmcarp:
Added!

PR #644, opened 2015-09-11
Feature/distinct zero

jmcarp:
* Exclude zero from color bounds, giving zero a distinct color
* Add tooltips to committee state map

@noahmanger the specific color values will probably need some adjustment, but this is a start.

### DIFF
--- a/static/js/modules/maps.js
+++ b/static/js/modules/maps.js
@@ -36,7 +36,7 @@ var compactRules = [
   ['', 0]
 ];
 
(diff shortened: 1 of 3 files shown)


noahmanger:
Ugh. I had to pull requests open in different tabs and accidentally merged this one when I meant to merge the other. This doesn't seem to be working as expected. For example, these states that have $0 still have color:

screen shot 2015-09-15 at 12 31 40 am

Issue #653, opened 2015-09-12
Data:  Other Spending - Electioneering Communcations


PaulClark2:
Started at 2012 Presidential Outside Spending (Other Spending) and scrolled down to electioneering communications

https://fec-dev-web.18f.gov/elections/president/2012/?tab=spending
![obama-ec-12](https://cloud.githubusercontent.com/assets/11817734/9828712/9f904428-58bf-11e5-86e5-ae7613c28f89.png)

Then selected Crossroads Grassroots Policy Strategies -- Landed on a 2009-2010 page for some reason

![crps-2010](https://cloud.githubusercontent.com/assets/11817734/9828723/c58a7e1e-58bf-11e5-8c7a-ecaa6eb9d5dd.png)

https://fec-dev-web.18f.gov/committee/C30001655
Should have data for 2010, 2011 and 2012, but that data doesn’t show up.
select extract (year from cvg_end_dt), f.mrf_rec, f.cvg_end_dt, f.form_tp, f.ttl_disb_this_stmt  from form_9 f where cmte_id = 'C30001655' and mrf_rec = 'Y';

jmcarp:
Thanks for reporting. Here's what's going on:

* The cycle defaults to 2009-2010 because this committee's only F1 was filed in 2010
* No reports show up because this committee has no reports in `factpacsandparties_f3x`

If you look up this committee in FEC Viewer, we get the same message: "No financial activity reported."

We can change either of these behaviors--currently, the committee detail views don't use Form 9 filings, but we can add them if that's the behavior you're looking for. Let me know what you think @PaulClark2.

CC @LindsayYoung @jwchumley

PaulClark2:
This committee will never show up in factpacsandparties_f3x because it is a Form 9 filer.

jmcarp:
Right, what I'm saying is that the financial reports we're currently using on the committee detail page only pull from the fact* tables--and it looks like FEC Viewer isn't using Form 9 filings either. We can easily add Form 9 data for electioneering committees, but we'll have to decide what information to show in those cases.

What's the priority of adding Form 9 data? Is this something we should do before launch?

PaulClark2:
If that's the case then all of the electioneering communications (ECs) and communication cost (CCs) links from the "Other Spending" tab will break.  ECs file Form 9 and CCs file Form 7.   This is a presentation and internal testing problem.  These should probably be removed from the Other Spending display.  

[...]

PR #655, opened 2015-09-12
Hide all but filings tabs for electioneering committees.

jmcarp:
All committee detail tabs except the filings tab are empty for
electioneering committees. This patch hides irrelevant tabs for these
committees. Template partials specific to the committee detail page are
also moved into their own folder.

@LindsayYoung @PaulClark2 are there other committee types that won't show up in any of the `fact*` tables? If so, we can make them behave like electioneering committees and hide irrelevant tabs.


jmcarp:
See discussion in #653.

PaulClark2:
I am out of the office until 09/15/2015.

I will respond to you email when I return.  If you need immediate
assistance please contact Jeff Chumley, jchumley@fec.gov, or Wei Luo,
wluo@fec.gov.



Note: This is an automated response to your message  "Re: [openFEC-web-app]
Hide all but filings tabs for electioneering committees. (#655)" sent on
9/14/2015 3:39:35 PM.

This is the only notification you will receive while this person is away.

PR #659, opened 2015-09-14
Fix regression on election pin colors.

jmcarp:
Map pin colors are currently all set to gray. This patch restores the
expected pin colors.


msecret:
Was this regression caught with automated tests? Is it possible to test with automated tests?

jmcarp:
No / yes. I'm planning to bring up test coverage for the leaflet maps once the interface stabilizes--right now it's still changing pretty often.

PR #665, opened 2015-09-14
Adding coverage period but need to format dates

LindsayYoung:
On the candidate page it will give the time period for the totals


LindsayYoung:
@jmcarp I am having trouble adding date formatting, can you look if you have a sec?

Issue #530, opened 2015-08-27
Browse Receipts "This Month" and "This Quarter" options


jwchumley:
https://fec-dev-web.18f.gov/receipts?min_date=08%2F01%2F2015&max_date=08%2F31%2F2015

"This Month" is not a valid option. We should almost never have data for this month. If you browse "this month" now you get six transactions. Five of them are data entry errors (8 in the month should actually be 6) and one of them is an anomaly. A committee filed their October quarterly 65 days early.

"This Quarter" is iffy. For the first fifty days of each quarter there should be no transactions available for "this quarter". Then you get the first month's transactions of monthly filing committees.   Quarterly filing committees will never show up under "this quarter" except for data entry errors and a few anomalies. 

I think "This year" is OK. It technically has the same issues as the others and it will never have the last quarterly filing or the last monthly filing but it doesn't seem to imply completeness like Month and Quarter do.

leahbannon:
If we wanted to keep something that was shorter than a year, is there a more recent option with good data? Is "last quarter" an option?

jwchumley:
As in "most recently completed?" I agree that it would be nice to have 
something smaller than year.

noahmanger:
What if we show:

- Last quarter (Q1 2015)
- This year (2015)
- This cycle (2015 - 2016)

leahbannon:
Sounds good to me. @jwchumley thoughts?

AmyKort:
I like Noah's idea. 

@jwchumley

noahmanger:
Passing to @LindsayYoung

PR #671, opened 2015-09-15
Feature/add filings dates

LindsayYoung:
the month, quarter, year checkbox and custom date box.


noahmanger:
Working well, but they're different than what we said we wanted here: https://github.com/18F/openFEC-web-app/issues/530

Last quarter
This year
This cycle

Maybe for filings it makes sense to have "This quarter" as well? But I don't know if it makes sense for Receipts and Disbursements.

LindsayYoung:
This makes the filters the same as the disbursement date filter.

LindsayYoung:
This also gets rid of the stray "Today" on disbursements, so it would be good to merge

noahmanger:
Just gotta sort out an errant diff in tables.js then should be good.

noahmanger:
I think we still may want to add ones for "cycle" and "last quarter" separately though.

PR #675, opened 2015-09-15
Feature/hide filings column

jmcarp:
This is a start on https://github.com/18F/openFEC/issues/1198, but we'll have to discuss more before closing.


LindsayYoung:
looks good will merge when tests pass

PR #678, opened 2015-09-16
Counting candidates is hard, lets just show the money

LindsayYoung:
We won't have conflicting candidate counts and we still have big numbers for context

### DIFF
--- a/templates/search.html
+++ b/templates/search.html
@@ -15,7 +15,7 @@
       <div class="card card--wide">
         <div class="card__image__container">
           <img class="card__image icon--complex" src="/static/img/i-candidates--neutral.svg"  alt="Icon representing candidates">
-          <h2 class="card__title">Candidate data</h2>          
+          <h2 class="card__title">Candidate data</h2>
         </div>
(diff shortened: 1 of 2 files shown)


templates/search.html
```diff
@@ -128,10 +128,6 @@ <h2 class="icon-heading__title widget__title"><span class="t-short">Find electio
           <h2 class="widget__title"><span class="t-short">So far in the 2016 presidential election:</span></h2>
           <div class="widget__content">
             <div>
-              <h5 class="t-data-header">Candidates running</h5>
-              <span class="t-big-data js-count"></span>
```
msecret: Is the js code that is using js-count still needed elsewhere, or should it also be removed?

LindsayYoung: I am not sure. I will take a look again. Thanks Marco!

noahmanger:
@LindsayYoung did you resolve that question?

noahmanger:
I just removed the js references to js-count. Will merge once tests pass.

PR #707, opened 2015-09-17
Feature/update maps

jmcarp:
* Build US state topojson from Census data (fixes FIPS code for Minnesota)
* Use the same FIPS data as the API for developer sanity
* Fix a few miscellaneous election map bugs
* Add data provenance notes


noahmanger:
Which election map bugs does this fix? Just so I know where to look.

jmcarp:
Good call--I've been meaning to institute some standards for documenting issues and pull requests, and the description of this patch wouldn't have made the cut.

Fixes:
* Clicking on Minnesota doesn't zoom in on Nevada
* Clicking on Alaska doesn't highlight all states (this was happening because Alaska so big that zooming to it triggers the map to show all states)
* If the URL includes a zip / state / district, don't draw the national map and then switch to the zip / state / district map

noahmanger:
Great. Functionally works great. @msecret do you want to review the js?

jmcarp:
@msecret do you have time to review? Would like to get this squared away soon if possible.

jmcarp:
@noahmanger turns out Marco's OOO for the week. We may want to merge now and ask for a review next week.

PR #728, opened 2015-09-21
Use hyphens for all dates in URLs.

### LINKED ISSUE #713: Back button on /receipts not working
Once on /receipts, I'm not able to use the browser back button to return to whatever page I was on before. I'm not seeing this issue on other pages. We probably should have a test for this, no?

jmcarp:
This makes for more readable URLs and fixes a bug on back and forward
navigation resulting from conversion between slash- and hyphen-delimited
dates.

[Resolves #713]

### DIFF
--- a/templates/macros/filters/date.html
+++ b/templates/macros/filters/date.html
@@ -4,15 +4,15 @@
   <label for="{{ name }}" class="label">{{ label }}</label>
   <ul>
     <li>
-      <input id="radio-2" name="{{ name }}" type="radio" data-min-date={{ dates['month'][0] | date }} data-max-date={{ dates['month'][1] | date }}>
+      <input id="radio-2" name="{{ name }}" type="radio" data-min-date={{ dates['month'][0] | date(fmt='%m-%d-%Y') }} data-max-date={{ dates['month'][1] | date(fmt='%m-%d-%Y') }}>
       <label for="radio-2">This month</label>
     </li>
     <li>
-      <input id="radio-3" name="{{ name }}" type="radio" data-min-date={{ dates['quarter'][0] | date }} data-max-date={{ dates['quarter'][1] | date }}>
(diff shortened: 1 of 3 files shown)


templates/macros/filters/date.html
```diff
@@ -22,9 +22,9 @@
   </ul>
   <div class="date-range-input">
     <label for="min_{{ name }}">From</label>
-    <input type="text" id="min_{{ name }}" name="min_{{ name }}" class="js-min-date" data-inputmask="'alias': 'mm/dd/yyyy'">
+    <input type="text" id="min_{{ name }}" name="min_{{ name }}" class="js-min-date" data-inputmask="'alias': 'mm-dd-yyyy'">
```
noahmanger: Can we not keep the slashes here?

templates/search.html
```diff
@@ -63,7 +71,13 @@ <h2 class="card__title">Disbursement data</h2>
         <div class="card__content">
           <h2 class="card__title">Disbursement data</h2>
           <ul class="card__links list--bulleted">
-            <li><a href="{{ url_for('disbursements', min_date=dates['year'][0] | date, max_date=dates['year'][1] | date) }}">Disbursements from committees this year &raquo;</a></li>
+            <li>
+              <a href="{{ url_for(
+                'disbursements',
+                min_date=dates['year'][0] | date,
+                max_date=dates['year'][1] | date)
```
noahmanger: Needs the proper date filter.

templates/macros/filters/date.html
jmcarp: Yes, but not without making other changes first. We're basically plugging the filter values directly into the URL at the moment, which means we need to use the same date format in the URL and in the filters. I can get around this you think it would be better to use slashes here.

noahmanger: Ok. Fine to implement later.

templates/search.html
jmcarp: Good catch, fixed.

PR #731, opened 2015-09-22
Updates US

emileighoutlaw:
Styles all U.S. as US for site consistency, per discussion Monday.


jmcarp:
What should we do about "U.S.A."? This crops up once: "Socialist Party U.S.A.".

emileighoutlaw:
Good call. I think we should be consistent, so I'll change to USA.

Issue #701, opened 2015-09-17
There is an expectation that the committee information will be on the committee profile page. We need to figure out a solution for showing "Other spending" aggregates on committee pages


noahmanger:
In order for users to have a comprehensive understanding of the IEs, ECs or communication costs that name a committees make, we should show those "Other spending" categories aggregated by the candidate they name and if it's in support of or opposition to that candidate. 

This allows the user to see an "other spending" aggregate on an election or candidate page, and then click through to that committee to see who all they've been supporting or opposing. 

This needs some design work to figure out how this piece lives on these pages and what the rules are.

jmcarp:
Was this issue referring to the other spending tab on the candidate page or the committee page? We added the first but not the second.

noahmanger:
Oops! I misread.

noahmanger:
Card: https://18f.storiesonboard.com/m/fec/!card/vfBqyymmLoMAzznTwluFuw

noahmanger:
Duplicate of https://github.com/18F/openFEC-web-app/issues/978

PR #732, opened 2015-09-22
Feature/candidate other spending

### LINKED ISSUE #700: Implement "other spending" tab on candidate pages
In order for users to see which other groups are making IEs for or against a candidate, we should show them on the candidate's page.

Question for @jenniferthibault and @onezerojeremy : do you think this should just be on a separate tab? We had previously tried to remove tabs from candidate pages but this might be a fine reason to bring them back.

jmcarp:
Resolves #700. Flagging @noahmanger and @LindsayYoung to review behavior and code.

### DIFF
--- a/static/js/modules/columns.js
+++ b/static/js/modules/columns.js
@@ -15,6 +15,26 @@ var sizeInfo = {
   2000: {limits: [2000, null], label: 'Over $2000'},
 };
 
+var supportOpposeMap = {
+  S: 'Support',
+  O: 'Oppose',
+};
+
+var supportOpposeColumn = {
+  data: 'support_oppose_indicator',
+  render: function(data, type, row, meta) {
+    return supportOpposeMap[data] || 'Unknown';
+  }
+};
+
+var amendmentIndicatorColumn = {
+  data: 'amendment_indicator',
+  className: 'min-desktop',
+  render: function(data) {
+    return decoders.amendments[data] || '';
+  },
+};
+
 var filings = {
   pdf_url: tables.urlColumn('pdf_url', {data: 'document_description', className: 'all', orderable: false}),
   filer_name: {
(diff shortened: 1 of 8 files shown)


LindsayYoung:
I am so excited about this!

noahmanger:
Nice! Working well. Can you add sorting to the third and fourth columns as well?

LindsayYoung:
For the candidate page, can we show the candidate aggregates instead of the line items? So we would just show "$100 Fake PAC oppose" on the candidate page. What you have here, I think, would be perfect for the committee page.

Also @mahmanger @emileighoutlaw any ideas how to make it very clear that this money is not spent by the candidate?

noahmanger:
Ah I was wondering about showing itemized vs. aggregates. I think I agree: let's show aggregates here and once we have an IE browse table, we can show the itemized there. 

And yes, we need a blurb explaining what IEs are and how they are not made in coordination with the candidate.

LindsayYoung:
Now that I am thinking about it, I think that schedule_e should be in a browse view and committees should have the aggregates too, like the election pages.

noahmanger:
schedule_e is IEs? We have an issue started for getting these on committee pages: https://github.com/18F/openFEC-web-app/issues/701

LindsayYoung:
Yea, #701 says to aggregate by candidate and that sounds right to me.

jmcarp:
Patch updated to use aggregates.

noahmanger:
This looks good to me except "Given by" should be "Spent by". 

Are we heeding all these Hound warnings?

jmcarp:
Updated. When I wrote the patch, we were still using "given by". I also changed the text on the election other spending tab.

For now, I'm paying attention to some Hound warnings more than others. I'm not fanatical about line length requirements, so am going to ignore those for now. We can customize the Hound configuration to skip certain warnings, which I'll get to soon.

noahmanger:
Great. Makes sense to me. Hound does seem helpful for sure.

PR #746, opened 2015-09-23
Feature/data explinations

### LINKED ISSUE #726: add link to pdf for receipts and disbursements
In the panel view it would be great to have a link view the source.

Here is an example pattern for lining to a particular page:
http://docquery.fec.gov/pdf/718/201507159000204718/201507159000204718_000002.pdf

### LINKED ISSUE #736: Add disclaimer to /receipts + /disbursements about how far back data goes
We need to specify on each page how far back the available data goes. And eventually would be good to add validation to the date fields so you can't set a date before the earliest limit.

LindsayYoung:
- Add date disclaimer for schedule A and B 
- Add link to page of the record for Schedule A and B
- Standardize some capitalization: @emileighoutlaw asked for sentences case 
- We had "date" as a heading in cases were there were 2 kinds of dates so I added "disbursement date" or "receipt date" 
- Changed receipt date to filing date in the schedule A and B panel descriptions, because recipt date is confusing on the recipes page. 
- I moved description and memo text next to each other because the memo filed can sometimes flesh out the description.

fix #736
fix #726

### DIFF
--- a/static/templates/disbursements.hbs
+++ b/static/templates/disbursements.hbs
@@ -52,7 +52,7 @@
   <h4 class="panel__title">Filing information</h4>
   <table>
     <tr>
-      <td class="panel__term">Receipt date</td>
+      <td class="panel__term">Filing date</td>
       <td>{{{ datetime receipt_date }}}</td>
     </tr>
     <tr>
--- a/static/templates/disbursements.hbs
+++ b/static/templates/disbursements.hbs
@@ -24,7 +24,11 @@
       <td>{{{ currency disbursement_amount }}}</td>
     </tr>
     <tr>
-      <td class="panel__term">Receipt date</td>
+      <td class="panel__term">Description</td>
+      <td>{{ disbursement_description }}</td>
+    </tr>
+    <tr>
+      <td class="panel__term">Disbursement date</td>
       <td>{{{ datetime disbursement_date }}}</td>
     </tr>
     <tr>
@@ -32,8 +36,8 @@
       <td>{{ memo_text }}</td>
     </tr>
     <tr>
-      <td class="panel__term">Description</td>
-      <td>{{ disbursement_description }}</td>
+      <td class="panel__term">Memo code</td>
+      <td>{{ memo_code }}</td>
     </tr>
     {{#if election_type_full }}
     <tr>
@@ -42,6 +46,20 @@
     </tr>
     {{/if}}
   </table>
+</div>
+
+<div class="panel__row">
+  <h4 class="panel__title">Filing information</h4>
+  <table>
+    <tr>
+      <td class="panel__term">Receipt date</td>
+      <td>{{{ datetime receipt_date }}}</td>
+    </tr>
+    <tr>
+      <td class="panel__term">View report page</td>
+      <td><a href="{{{ pdf_url }}}" target="_blank">View on fec.gov</a></td>
+    </tr>
+ </table>
 </div>
 
 <div class="panel__row">
--- a/static/templates/disbursements.hbs
+++ b/static/templates/disbursements.hbs
@@ -24,12 +24,12 @@
       <td>{{{ currency disbursement_amount }}}</td>
     </tr>
     <tr>
-      <td class="panel__term">Description</td>
-      <td>{{ disbursement_description }}</td>
-    </tr>
-    <tr>
       <td class="panel__term">Disbursement date</td>
       <td>{{{ datetime disbursement_date }}}</td>
+    </tr>
+    <tr>
+      <td class="panel__term">Description</td>
+      <td>{{ disbursement_description }}</td>
     </tr>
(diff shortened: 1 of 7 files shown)


emileighoutlaw:
Content looks good from my end.

LindsayYoung:
@noahmanger this adds some fine print in the filters. I also specified what date we are looking at in the tables when there are more than one date and for committees, I changed "file date" to "first file date" because I thought it was kind of confusing without the first. I wanted to flag this to your attention because those changes that makes some headings a little longer, and I know space is at a premium.

static/templates/disbursements.hbs
```diff
@@ -45,6 +49,20 @@
 </div>
 
 <div class="panel__row">
+  <h4 class="panel__title">Filing information</h4>
+  <table>
+    <tr>
+      <td class="panel__term">Filing date</td>
+      <td>{{{ datetime receipt_date }}}</td>
```
noahmanger: should this be filing_date ?

templates/partials/receipts-table.html
```diff
@@ -1,11 +1,11 @@
 <table id="results" class="data-table" data-type="donation">
   <thead>
     <tr>
-      <th scope="col">Contributor Name</th>
+      <th scope="col">Contributor name</th>
       <th scope="col">State</th>
       <th scope="col">Employer</th>
       <th scope="col">Amount</th>
-      <th scope="col">Date</th>
+      <th scope="col">Receipt date</th>
```
noahmanger: Is Receipt date what we want and not filing date?

static/templates/disbursements.hbs
```diff
@@ -45,6 +49,20 @@
...
+    </tr>
+    <tr>
+      <td class="panel__term">View report page</td>
+      <td><a href="{{{ pdf_url }}}" target="_blank">View on fec.gov</a></td>
```
jmcarp: This duplicates the "view original image" link at the top of the panel.

static/templates/receipts.hbs
```diff
@@ -51,6 +51,20 @@
...
 <div class="panel__row">
+  <h4 class="panel__title">Filing information</h4>
+  <table>
+    <tr>
+      <td class="panel__term">Filing date</td>
+      <td>{{{ datetime receipt_date }}}</td>
+    </tr>
+    <tr>
+      <td class="panel__term">View report page</td>
+      <td><a href="{{{ pdf_url }}}" target="_blank">View on fec.gov</a></td>
```
jmcarp: This duplicates the "view original image" link at the top of the panel.

static/templates/disbursements.hbs
LindsayYoung: They use "receipt date" for the day the filing was received by the FEC. But that becomes confusing on the Receipts page, where there is "receipt date" and "contribution receipt date" I thought if we called "receipt date" "filing date" that makes the two dates more clear. I changed it on the disbursements since those tables are similar and if there is a section called receipts, receipt date might be confusing. If you have a better idea of how to make it less confusing we can change it.

templates/partials/receipts-table.html
LindsayYoung: good catch-- thanks!

LindsayYoung: So, I checked and that column is correct, we are sorting on the day the receipt was made- "Receipt date" 

Like I said, the column names are confusing and I am up to tweaking them. Because it is confusing, I think it is important to be explicit about which date we are showing, rather than just saying date

static/templates/disbursements.hbs
LindsayYoung: Well, I am blind. Thanks for spotting that! 

A thought on the language, what if it says--  "View report page" instead of "View original image"  Does @emileighoutlaw or @noahmanger have strong feelings on this micro copy?

emileighoutlaw: How about "View original report" ?

LindsayYoung: We are showing the page that the record is on, which is handy because the reports can be thousands of pages.

noahmanger: I like "View report page", but then in the filings panels we'd need it to be "View report". Right now we use "View original image" for both.

I was looking at this and think it might make sense for the link under filing date to link to the full report, just so people can have it. I think we should also identify the report it was filed on by name, so you can see, e.g. April Quarterly 2015.

noahmanger:
I can style the disclaimers in a separate PR after we merge this.

jmcarp:
What do we need to do to get this ready for merge @LindsayYoung @noahmanger? I'd like to get rid of the duplicate pdf links, and it sounds like we may want to change some text, but hopefully we can resolve other issues in separate patches.

LindsayYoung:
Merging and doing additional, small changes sounds good to me.

PR #748, opened 2015-09-23
Remove `tabindex` from main section.

jmcarp:
This fixes an interaction with dropdown toggles such that clicking
options within the dropdown panel didn't select the items. Thanks
@noahmanger for finding the fix.


jmcarp:
Lulz, didn't see this in #747.

PR #753, opened 2015-09-24
Replace empty tables with placeholder text.

### LINKED ISSUE #512: Design and implement empty results for data tables
Some data tables may be empty frequently (e.g., electioneering, communication costs). We should design and implement some text and/or visuals to show when a table is empty, or decide that the current behavior is adequate. Here's how this looks now (e.g. https://fec-dev-web.18f.gov/elections/house/NY/01/2012/?tab=spending):
screen shot 2015-08-24 at 3 35 35 pm

jmcarp:
On fetching results from tables without filters, destroy the `DataTable`
instance and add placeholder text if no results found. Note: this patch
uses a placeholder placeholder, which would probably benefit from style
and content review by @noahmanger and @emileighoutlaw.

[Resolves #512]

### DIFF
--- a/static/js/modules/tables.js
+++ b/static/js/modules/tables.js
@@ -40,11 +40,6 @@ $.fn.DataTable.Api.register('seekIndex()', function(length, start, value) {
   }
 });
 
-// Only show table after draw
-$(document.body).on('draw.dt', function() {
-  $('.datatable__container').css('opacity', '1');
-});
-
 function yearRange(first, last) {
   if (first === last) {
     return first;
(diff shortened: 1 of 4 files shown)


static/js/modules/tables.js
```diff
@@ -479,6 +476,20 @@ function initTable($table, $form, path, baseQuery, columns, callbacks, opts) {
...
+/**
+ * Replace a `DataTable` with placeholder text if no results found. Should only
+ * be used with unfiltered tables, else tables may be destroyed on restrictive
+ * filtering.
+ */
+function hideEmpty(api, data, response) {
+  if (!response.pagination.count) {
+    api.destroy();
+    var $table = $(api.table().node());
+    $table.before('No data found.');
```
noahmanger: Can we put this in a `<div class="message message--alert">`?

noahmanger:
Not all of the tables on the single committee page use the default `tableOpts` so they don't have the correct setting.

jmcarp:
Updated!

PR #769, opened 2015-09-26
Specifying filters to show "contributor employer" and "contributor occupation"

jenniferthibault:
@noahmanger and I had talked about this in  https://github.com/18F/fec-style/issues/82 but then never made a separate issue to track it


noahmanger:
Oh nice. Can you change "State" to "Contributor state" too?

noahmanger:
*If that's a good idea. Not sure.

jenniferthibault:
All set! I wasn't sure on that either, but looking at it again, it does make sense

PR #781, opened 2015-09-27
Update built-in date filters.

### LINKED ISSUE #725: Disbursements and Receipts time filters don't all make sense
From @jwchumley

> The Disbursement date options are a problem.  "Today" and "This Month" will always be null  because filing deadlines covering this month are at best next month. "This Quarter" will always be 0 for quarterly filers and will alwasy  miss the third month in the quarter for monthly filers. Sometimes the second month will be missing too.  The data in this system are not real time.

>In the Filter Receipts column, "Receipt Date" section some of the options don't make sense.
>"Today" Will never yield results because the report covering today won't be filed until next month if it is a monthly and as many as three months from now if Quarterly.
>"This Month" has similar issues. All regularly scheduled reports covering a given month are filed with us the following month or quarter so we can't display the data for this month until next month when it is no longer this month. There are some exceptions. Pre Generals typically cover the first few weeks of October and are filed toward the end of October. Probably some Pre Specials as well but these exceptions are few.

jmcarp:
[Resolves #725]

### DIFF
--- a/templates/macros/filters/date.html
+++ b/templates/macros/filters/date.html
@@ -5,20 +5,16 @@
     <legend for="{{ name }}" class="label">{{ label }}</legend>
     <ul>
       <li>
-        <input id="radio-2" name="{{ name }}" type="radio" data-min-date={{ dates['month'][0] | date(fmt='%m-%d-%Y') }} data-max-date={{ dates['month'][1] | date(fmt='%m-%d-%Y') }}>
-        <label for="radio-2">This month</label>
+        <input id="radio-1" name="{{ name }}" type="radio" data-min-date={{ dates['year'][0] | date(fmt='%m-%d-%Y') }} data-max-date={{ dates['year'][1] | date(fmt='%m-%d-%Y') }}>
+        <label for="radio-1">This year</label>
       </li>
       <li>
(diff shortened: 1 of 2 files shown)


noahmanger:
Not positive, but I think we want "Last quarter" too. Fine to merge either way.

templates/macros/filters/date.html
```diff
@@ -5,20 +5,16 @@
     <legend for="{{ name }}" class="label">{{ label }}</legend>
```
LindsayYoung: This would change all the date filters, and not just the receipts and disbursements. This filter works well for filings. (If it is too annoying to have both filters now, it can wait until after launch)

jmcarp:
@noahmanger: as of #725, looks like we want "last quarter" out, but we may have changed our collective minds in another issue or conversation. @LindsayYoung: has FEC weighed in on which filters are appropriate for filings? Simple enough to use different options for different tables if necessary.

noahmanger:
Yeah, I feel like this is still an open question because there's some disagreement in the group. They were very firm that we should remove "Today" and "This month" and "This quarter" from the other pages though, so I'm going to merge this now and then we can add in further options as we agree. 

Also, obviously, one can set manual date filters to whatever they want.

PR #786, opened 2015-09-28
Feature/updating meta tags

noahmanger:
- Removes references to 18F in the Twitter cards
- Replaces hard-coded URL's with template tags
- Changes the title of the web-app home page to "Campaign finance data" instead of "Home"


noahmanger:
@jmcarp how does this look?

PR #788, opened 2015-09-28
Glossary links

emileighoutlaw:
Adds glossary links 

Addresses sentence-case capitalization issues

Resolves 18F/openFEC-web-app/issues/555#issuecomment-143491873


emileighoutlaw:
@noahmanger ugh. If you have time tomorrow or Monday, can you talk me through this? It looks like it failed a bunch of stuff and since it's been sitting here for five days is now behind the site and had merge conflicts.

noahmanger:
Oh yeah we've got ourselves some conflicts! I'll take a look.

noahmanger:
Ok, I resolved the conflicts but | have some questions:
- I'm not sure if it works well to have column headers as glossary terms. Clicking on table headers should sort the column, but when it pulls out the glossary it conflicts with that. I'd vote for moving the definitions to the filter labels instead.
- I worry a little about term overload. I think we should only make the first instance of a term on a page a glossary term, and I think with terms that are a little more frequent and/or colloquial (like Candidate), we don't need to use the glossary unless it provides significant added meaning. 

Thoughts?

emileighoutlaw:
I think you're completely right. I'll go back in and winnow

emileighoutlaw:
Okay @noahmanger! Thanks for your great edits and patience. I think this is ready for your review (again).

static/templates/districts.hbs
```diff
@@ -1,4 +1,4 @@
-<option value="">Select district</option>
+<option value="">Select <span class="term" data-term="District">district</span></option>
```
noahmanger: This is an option in the select, so shouldn't have a span.

templates/committees-single.html
```diff
@@ -2,7 +2,7 @@
 {% import 'macros/null.html' as null %}
 
 {% block title %}
-    {{ name }} - Committee overview
+    {{ name }} - <span class="term" data-term="Committee">committee</span> overview
```
noahmanger: This is the page title so shouldn't have the term class.

templates/elections.html
```diff
@@ -21,7 +21,7 @@ <h1 class="entity__name">{{ cycle }} Presidential election</h1>
           {% else %}
             <h1 class="entity__name">US {{ office|title }} election: {{ state|fmt_state_full }}</h1>
             {% if district %}
-              <h2>United States House District {{ district }}</h2>
+              <h2>United States House <span class="term" data-term="District">District</span> {{ district }}</h2>
```
noahmanger: I think this shouldn't be a term because it's read as "District 2" not like "District: 2".

templates/partials/committees-table.html
```diff
@@ -8,7 +8,7 @@
             <th scope="col">First file date</th>
             <th scope="col">Type</th>
             <th scope="col">Designation</th>
-            <th scope="col">Organization type</th>
+            <th scope="col"><span class="term" data-term="Organization type">Organization type</span></th>
```
noahmanger: This is a table header, so shouldn't be a term.

templates/partials/search-results-candidate.html
```diff
@@ -26,7 +26,7 @@ <h3 class="result__title">
     <div class="result__column">
       {% if candidate.principal_committees %}
         {% with committee=candidate.principal_committees[0] %}
-        <span class="t-block t-bold">Principal Campaign Committee:</span>
+        <span class="t-block t-bold">Principal campaign committee:</span>
```
noahmanger: Are you sure we're not capitalizing Principal campaign committees? Are they a proper noun?

amypike: @noahmanger, @amykort and I have no preference and defer to @emileighoutlaw and whatever she thinks users will find most approachable/least off-putting.  I believe Emily thought we should capitalize as little as possible.

emileighoutlaw: Yes! FEC.gov's current style is not capitalized, and I'm a huge fan of no caps whenever possible. People tend to read lots of big capital letters as scary and dictatorial, especially in a gov setting.

emileighoutlaw:
Well the silver lining of you having to look at this a million times @noahmanger is that I'm learning a ton about HTML. Made your edits, and thank you.

templates/partials/search-results-candidate.html
noahmanger: Works for me!

emileighoutlaw:
ping @noahmanger Anything I need to do?

noahmanger:
I don't think so...waiting for tests to pass.

emileighoutlaw:
Awesome! Just wanted to make sure I didn't drop the ball somehow. :)

PR #793, opened 2015-10-01
Feature/adding usersnap

noahmanger:
This adds widget code for Usersnap. We're currently on a 14 day trial, but this could allow us to see if it meets the needs. 

I think we'll need to do some configuration to get it to work behind a password. They have this doc: https://usersnap.com/help/troubleshooting/protected


noahmanger:
Demoted to in progress while we wait for an answer from DevOps.

Issue #778, opened 2015-09-26
Have consistent "target=blank" behavior on document links


noahmanger:
From Eric: On this filings page, the document link to docquery.fec.gov opens in a new tab, but in the expanded “more” tab that opens when you click the arrow, the “View original image” button loads the document in the same tab.

noahmanger:
Resolved by #794

PR #794, opened 2015-10-01
Feature/updating filings panel

### LINKED ISSUE #789: Add filer info to filings panel
We should include the filer's name and maybe other information on the details panel for filings.

screen shot 2015-09-27 at 8 52 44 pm

noahmanger:
Adds missing info to filings panels:
- Filer name
- Receipt date
- Amendment indicator

Also adds a more readable datetime helper. @emileighoutlaw what do you think about this format for dates in this situation: `Jan 1, 2015`? I think it makes sense to stick with `01-01-2015` format in the columns and in the filters, but if we're just displaying a date in a regular text area, I find the prettier version helpful.

Last, makes the pdf link a `target="_blank"` to make it consistent with the main url column in the table.

Resolves #789 and #778 

screen shot 2015-10-01 at 2 27 11 pm

### DIFF
--- a/static/js/modules/helpers.js
+++ b/static/js/modules/helpers.js
@@ -5,6 +5,7 @@
 var URI = require('URIjs');
 var _ = require('underscore');
 var moment = require('moment');
+var decoders = require('./decoders');
 var Handlebars = require('hbsfy/runtime');
 
 var intl = require('intl');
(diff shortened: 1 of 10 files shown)


noahmanger:
This should be good to merge if anyone can review it.

static/js/modules/helpers.js
```diff
@@ -24,8 +25,22 @@ function datetime(value) {
   var parsed = moment(value, 'YYYY-MM-DDTHH:mm:ss');
   return parsed.isValid() ? parsed.format('MM-DD-YYYY') : null;
 }
+
+function prettyDatetime(value) {
```
jmcarp: This is a copy and paste of the `datetime` helper, aside from the format string. How about modifying the existing function to take a `format` or `pretty` option:
```
{{{datetime this.date pretty=true}}}
```

static/js/modules/helpers.js
```diff
@@ -24,8 +25,22 @@ function datetime(value) {
...
+  var parsed = moment(value, 'YYYY-MM-DDTHH:mm:ss');
+  return parsed.isValid() ? parsed.format('MMM D, YYYY') : null;
+}
+
+function amendmentDecoder(value) {
```
jmcarp: Prefer imperative function names--how about `decodeAmendment`?

static/js/modules/helpers.js
noahmanger: :+1:

noahmanger:
@jmcarp made those changes. Thanks for the suggestions.

LindsayYoung:
@noahmanger -- Do we want the button on the top to say "View original report"? I feel like we talked about that previously.

Issue #785, opened 2015-09-27
As a user who wants to know about real candidates, we will limit typeahead to statutory candidates.


LindsayYoung:
This is a brilliant suggestion from @boblannon that will filer out a lot of the prank filers, with beautiful names like "AAAAAAAAAA"

PaulClark2:
I like this solution better than excluding F2Zs (re:  https://github.com/18F/openFEC/issues/964) but we'd need to explain that we are excluding certain Form 2 filers, ballot list candidates and draft committee candidates.

LindsayYoung:
This is just for the Candidate browse view, people would default to the statutory candidates, but be able to see all candidates that filed an F2 if they unselect statutory candidate checkbox. 

We have been excluding candidates who we only have F2Z records, since those records are in poor condition. Simple data like district is wrong or missing. 

Also 18F/openFEC#964 is only affecting the cycle creation, making sure we only create a cycle dropdown for candidates that have a  F2, that is not working yet.

LindsayYoung:
I think "C" will work it looks like there is a bug in the dev api in /candidate it looks like it is using the oldest name and status, instead of the newest. 

18F/openFEC#1244

LindsayYoung:
Nope, the logic is right, but the receipt date on Obama's records, seems to have changed.

noahmanger:
We've decided against this.

boblannon:
oh, really?  I have the same problem with my typeahead, and was thinking of doing the same thing. what made you decide against? what are you doing instead?

noahmanger:
@LindsayYoung correct me if I'm wrong, but we decided against it because of the number of legit candidates that it filtered out. We decided better to leave it off for now, but in the future add a more intelligent way of weighting results to surface the accurate ones without accidentally cutting things out. 

Also, I just realized that this issue started out referring to a statutory candidate checkbox, which we've added.

PR #795, opened 2015-10-01
Adding a statutory candidate check box

LindsayYoung:
The statutory candidate data is much cleaner, so it makes a good default.

We can just add 'candidate_status=C' to all the links to that page. Is there a better way to handle that default?

I am simplifying this a bit, the full list of codes codes is:

C = Statutory candidate
F = Statutory candidate for future election
N = Not yet a statutory candidate
P = Statutory candidate in prior cycle

### DIFF
--- a/openfecwebapp/constants.py
+++ b/openfecwebapp/constants.py
@@ -211,3 +211,10 @@ amendment_indicators_extended = OrderedDict([
 candidate_status = OrderedDict([
     ('C', 'Statutory candidate'),
 ])
+candidate_status_extended = OrderedDict([
+    ('C', 'Statutory candidate'),
+    ('F', 'Future candidate'),
+    ('N', 'Not yet a candidate'),
+    ('P', 'Statutory candidate in prior cycle'),
+
+])
--- a/openfecwebapp/constants.py
+++ b/openfecwebapp/constants.py
@@ -212,7 +212,6 @@ candidate_status = OrderedDict([
     ('C', 'Statutory candidate'),
 ])
 candidate_status_extended = OrderedDict([
-    ('C', 'Statutory candidate'),
     ('F', 'Future candidate'),
     ('N', 'Not yet a candidate'),
     ('P', 'Statutory candidate in prior cycle'),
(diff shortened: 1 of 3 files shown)


LindsayYoung:
Fixes #785

LindsayYoung:
Just C, might be too limiting, we might want to include all non-nulls, what do you think @PaulClark2

PaulClark2:
If the default is "statutory candidate" we probably want to use "C," "F" and "P" as all of these codes indicate the candidate has spent or raised $5000.   The default value for candidate status (cand_status) is "N."  We may need to try including and excluding codes to determine what works best.

jmcarp:
What's the status of this patch @LindsayYoung? Do we want to stick with "C" only or add more options? Regardless of which statuses we choose by default, it seems like we'd want to allow users to filter by any status.

LindsayYoung:
Sorry, did not see this comment before, I will take a look!

LindsayYoung:
We were talking about defaulting to the statutory candidates for the browse views, so I am going to change the link slightly, I think we should change the language on the following to reflect the new, curated view, Instead of "All 2016 presidential candidates", just " 2016 presidential candidates" and so on. 

Would that be OK @emileighoutlaw and @noahmanger ?

screen shot 2015-10-08 at 11 10 53 am

emileighoutlaw:
I think that sounds great to me!

emileighoutlaw:
Great thought

LindsayYoung:
Going to change links in the fec-style dropdown next

noahmanger:
Great!

On Thu, Oct 8, 2015 at 9:55 AM, Lindsay Young <notifications@github.com>
wrote:

> Going to change links in the fec-style dropdown next
>
> —
> <https://github.com/18F/openFEC-web-app/pull/795#issuecomment-146621344>.
>



-- 
Noah Manger

LindsayYoung:
The API part of this is working locally and on production, but is having issues on dev, I will look into that now.

(After this, merge 18F/fec-cms#89)

noahmanger:
Where's this at?

LindsayYoung:
I think the only thing left is debugging the tests.

LindsayYoung:
It looks like the tests are comparing results against the API, so we need to filter those results the same, I am not finding where it is making the comparison call in the test to add the same filtering.

PR #798, opened 2015-10-02
Content edits

emileighoutlaw:
Fixes two minor 
capitalization things.
Should be a quick merge.


jmcarp:
@emileighoutlaw it looks like tests are failing here due to a bug that was fixed a few weeks ago, and the parent of the most recent commit comes from about a month back. Your changes didn't break the tests, so I'm going to merge now, but in general, it's good practice to pull the latest changes before submitting a patch.

PR #808, opened 2015-10-05
Check HMAC signature in production.

jmcarp:
Sounds like @mbland is going to pull `HmacMiddleware` into a separate
library, so much of this code can be retired once that's ready.

Requires https://github.com/18F/fec-proxy/pull/4.


jmcarp:
Will be ready to merge once hmac_authentication_py is on the PyPI.
cc @mbland

jmcarp:
This is ready for review @LindsayYoung. There isn't a lot of code here, but it's a little complicated, so might be worth talking over.

PR #811, opened 2015-10-06
Customize linter configuration.

jmcarp:
* Set longer max length.
* Update pre-defined globals.

This also reminds me: are we all using jshint locally? Consider this my passive-aggressive reminder to @LindsayYoung and @noahmanger to grab the sublime plugin if you don't already have it. And if you think the linter is being too strict, feel free to update .jshintrc.

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,0 +1,2 @@
+javascript:
+  config_file: .jshintrc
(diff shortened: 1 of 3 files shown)


noahmanger:
Ah yep. Totally not using it locally. Adding now.

LindsayYoung:
:+1:

PR #816, opened 2015-10-07
Update travis secrets.

jmcarp:
Update with new continuous deployment credentials.


LindsayYoung:
The test that is failing says it can't import args from webargs.

jmcarp:
That error is unrelated to this patch--it's happening because the API and data app are using different (and incompatible) versions of webargs, and travis installs both sets of requirements to run the tests. This should be resolved in #817.

PR #823, opened 2015-10-13
Fixes ranges— easy merge

emileighoutlaw:
Moves dash from *minus*
to *n dash*— how ranges should
be represented.


emileighoutlaw:
cc @LindsayYoung not sure why this failed, but it's just a really easy content edit.

jmcarp:
Tests are failing because of an unrelated issue that's since been fixed.

PR #826, opened 2015-10-14
Miscellaneous manifest updates.

jmcarp:
* Remove unnecessary `server_name` configuration (now set by proxy)
* Increase timeout (minification can be slow)
* Set default domain to 18f.gov
* Remove open.api.gov route


jmcarp:
cc @LindsayYoung

PR #838, opened 2015-10-15
Feature/cycle tooltips

### LINKED ISSUE #824: To help users understand the time period they're looking at, add a tooltip to explain the two-year period select
Using the pattern implemented elsewhere (such as with the "Unique only" filter on /receipts), add a tooltip to two-year period selects on candidate, committee and election pages. The tooltip should explain what this means and also include links to a GitHub issue that contains wireframes for the 2/4/6 year version and invites the user to comment. 

**Steps:**
- [x] @emileighoutlaw write content for the tooltip (see below for sample language)
- [x] I'll implement
- [x] @jenniferthibault create new issue with mockups

**Draft language:** 
> This site currently groups data by two-year period. This means that the page you're looking at doesn't necessarily include all of this candidate or committee's activity since their last election. We're working on a new interface to allow you to view data for the entire four or six years between presidential or senate elections, and we want your feedback. Visit this issue on GitHub to view the latest designs and comment there or on the feedback button at the bottom of this page.

That language is rough and verbose, but wanted to include the main points I think we need to hit.

noahmanger:
- Moves the cycle select to the headers of the candidate and committee pages and removes it from the subtitle areas of each individual tab
- Adds a tooltip with an explanation and link to the GitHub issue

screen shot 2015-10-14 at 6 31 41 pm

Depends on https://github.com/18F/fec-style/pull/146
Resolves #824

### DIFF
--- a/templates/macros/cycle-select.html
+++ b/templates/macros/cycle-select.html
@@ -1,15 +1,15 @@
 {% macro cycle_select(cycles, cycle=none, location='query', id='cycle-select', tooltip='true') %}
 {% if cycles %}
-  {% if tooltip %}
-    <label for="{{ id }}" class="tooltip-trigger">Two-year period</label>
-    <div id="unique-tooltip" role="tooltip" class="tooltip tooltip--under">
-      <div class="tooltip__content">
-        <p>This site groups data in two-year periods— if a candidate or committee has been active more than two years, this page won't show all their data. We're working on a four-year view for presidential data and a six-year view for Senate data. But we need your feedback.</p>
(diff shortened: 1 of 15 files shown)


templates/macros/cycle-select.html
```diff
@@ -1,6 +1,15 @@
-{% macro cycle_select(cycles, cycle=none, location='query', id='cycle-select') %}
+{% macro cycle_select(cycles, cycle=none, location='query', id='cycle-select', tooltip='true') %}
```
jmcarp: How about using booleans for `tooltip`, like `tooltip=True`?

noahmanger: Oops. Thanks.

PR #842, opened 2015-10-15
Fix candidate URL on election page.

jmcarp:
* Use URL helper on election page.
* Delete unused code.

### DIFF
--- a/static/js/pages/elections.js
+++ b/static/js/pages/elections.js
@@ -4,7 +4,6 @@
 
 var d3 = require('d3');
 var $ = require('jquery');
-var URI = require('URIjs');
 var _ = require('underscore');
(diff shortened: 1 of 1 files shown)


static/js/pages/elections.js
```diff
@@ -154,7 +153,7 @@ function drawComparison(results) {
   var $comparison = $('#comparison');
   var context = {selected: results.slice(0, 10), options: results.slice(10)};
   $comparison.prepend(comparisonTemplate(context));
- new dropdown.Dropdown($comparison.find('.js-dropdown'));
+  new dropdown.Dropdown($comparison.find('.js-dropdown'));
```
noahmanger: Can we turn off this warning?

PR #845, opened 2015-10-15
Removing tooltips from election selects

noahmanger:
Removes the cycle select tooltips from the election page selects.


jmcarp:
Is the idea that the election pages aren't going to have configurable time windows like the committee detail pages? I was imagining they would.

noahmanger:
No, I believe you're right. I was just removing these because the new format of showing a label for the select that also has a tooltip associated with it works best if it's in the header of the page. But I'm now realizing it probably makes sense to just do that...so... I'm going to close this and make an issue for that.

Issue #847, opened 2015-10-16
So that cycle select placement is consistent, move them to the headers of election pages


noahmanger:
Election pages still have cycle selects in line with the section titles, whereas candidate and committee pages have them in the header area. 

I'd like to move them to the same area on candidate pages, but it's complicated by the presence of the map. 

At the same time, there's the opportunity to make the titling of these pages more consistent with how we structure candidate and committee pages, where the main heading area is:

```
# Page title
### Type of page
```
If / when we re-work these pages to show more office info ([along the lines of these wireframes](https://invis.io/2J4KL9KTN)), these areas will be fleshed out more.

Here's suggestions:

## Presidential
![screen shot 2015-10-15 at 5 06 01 pm](https://cloud.githubusercontent.com/assets/1696495/10530716/4d19332a-735f-11e5-93bd-a11628c1df4d.png)

## Senate
(Note, the map is currently broken)
![screen shot 2015-10-15 at 5 06 25 pm](https://cloud.githubusercontent.com/assets/1696495/10530719/56214b38-735f-11e5-8e4b-1eebad1273d7.png)

## House
![screen shot 2015-10-15 at 5 06 19 pm](https://cloud.githubusercontent.com/assets/1696495/10530739/615eadb0-735f-11e5-898d-013659d6197a.png)

noahmanger:
Just realized I've been waiting for feedback on this. @jenniferthibault do you have thoughts about this as an interim solution?

jenniferthibault:
@noahmanger this makes sense as an interim solution to me, the one small adjustment I'd make is to make sure the underline under the election doesn't bump into the map, it should end at whatever grid column is before the map div. (looking at the third example of California District 2).

Sorry I missed it! I only have email notifications set up for issues I'm tagged in, but maybe I should change that.

noahmanger:
Good catch!

PR #848, opened 2015-10-16
Updating cycle select placements

noahmanger:
- Adds a text string of the cycle to each section heading on candidate, committee and election page sections
- Moves the cycle select to the header of the election pages

Resolves https://github.com/18F/openFEC-web-app/issues/847


noahmanger:
Ok this is good to go. https://github.com/18F/fec-style/pull/153 resolves @jenniferthibault 's one piece of design feedback.

templates/partials/committee/receipts-tab.html
```diff
@@ -6,7 +6,7 @@
   <div class="container">
     <div class="section__heading">
       <h2 class="heading__title" id="section-2-heading">
-        Money received by the <span class="term" data-term="Committee">committee</span>
+        Money received by the committee - {{ cycle|fmt_year_range }}
```
jmcarp: Just curious--why drop the glossary lookup here? I'm not clear on when we want to wrap terms in glossary lookups and when we don't.

noahmanger: Good question. I think it makes sense to not make words in headings glossary links. @emileighoutlaw ?

noahmanger:
Conflicts resolved.

PR #857, opened 2015-10-16
adding candidate status filter to nav item

noahmanger:
Adds the statutory candidate filters on the URL for the "Candidates" link in the nav.


noahmanger:
This failure is from the glossary toggle, which was just fixed. So this should be good to go.

noahmanger:
@LindsayYoung we're no longer filtering by candidate status, right? So I should close this?

LindsayYoung:
Correct.

PR #860, opened 2015-10-19
updates URL capitalization

emileighoutlaw:
from fec.gov to FEC.gov

[ci skip]

### DIFF
--- a/templates/layouts/main.html
+++ b/templates/layouts/main.html
@@ -45,7 +45,7 @@
 
   <header class="site-header">
     <div class="disclaimer">
-      <span class="disclaimer__left">Looking for <a href="http://www.fec.gov">fec.gov?</a></span>
+      <span class="disclaimer__left">Looking for <a href="http://www.fec.gov">FEC.gov?</a></span>
       <span class="disclaimer__center">
         An official website of the United States Government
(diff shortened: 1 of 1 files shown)


templates/layouts/main.html
```diff
@@ -45,7 +45,7 @@
 
   <header class="site-header">
     <div class="disclaimer">
-      <span class="disclaimer__left">Looking for <a href="http://www.fec.gov">fec.gov?</a></span>
+      <span class="disclaimer__left">Looking for <a href="http://www.fec.gov">FEC.gov?</a></span>
```
noahmanger: I noticed our content guide now recommends all lowercase urls. Thoughts?

emileighoutlaw: Excellent note. And so glad you're reading our content guide!

We've been using `FEC.gov` and `beta.FEC.gov` in our other writings. But you're correct that it isn't aligned with our (well reasoned, well supported)18F style of always using lowercase. 

We *can* switch to lowercase (for consistency), but I think we'll need to change all our references to `fec.gov` or `beta.fec.gov`. 

@leahbannon — thoughts on this?

noahmanger: I'll merge this for now to keep things consistent, and then if we want to make a universal change we can.

emileighoutlaw:
Sounds good. I think this is something we should talk about, and I'm glad
you brought it up.

PR #861, opened 2015-10-19
Hide district if 00.

jmcarp:
Pending changes in https://github.com/18F/openFEC/pull/1278, some null
districts will be represented as "00" rather then `null`. This patch
hides both "00" and `null` districts.


jmcarp:
Ready for review @LindsayYoung or @noahmanger.

PR #863, opened 2015-10-20
Feature/history analytics

jmcarp:
Add virtual page views to Google Analytics. Sends page views on:

* Switching tabs
* Updating table filters
* Updating election search

Also sorts query parameters so that /?cycle=2012&tab=disbursements is handled like /?tab=disbursements&cycle=2012.

See https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications for details.

cc @noahmanger

### DIFF
--- a/static/js/modules/analytics.js
+++ b/static/js/modules/analytics.js
@@ -1,0 +1,51 @@
+'use strict';
+
+/* global ga */
+
+var URI = require('URIjs');
+var _ = require('underscore');
+
+function init() {
+  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+
+  ga('set', 'forceSSL', true);
+  ga('set', 'anonymizeIp', true);
+  ga('create', 'UA-48605964-22', 'auto');
+}
+
+function pageView() {
+  if (typeof ga === 'undefined') { return; }
+  var path = document.location.pathname;
+  if (document.location.search) {
+    var query = URI.parseQuery(document.location.search);
+    path += '?' + sortQuery(query);
+  }
+  ga('send', 'pageview', path);
+}
+
+function sortQuery(query) {
+  return _.chain(query)
+    .pairs()
+    .map(function(pair) {
+      return [pair[0], _.isArray(pair[1]) ? pair[1] : [pair[1]]];
+    })
+    .reduce(function(memo, pair) {
+      return memo.concat(_.map(pair[1], function(value) {
+        return [pair[0], value];
+      }));
+    }, [])
+    .sort()
+    .map(function(pair) {
+      return pair.join('=');
+    })
+    .join('&')
+    .value();
+}
+
+module.exports = {
+  init: init,
+  pageView: pageView
+};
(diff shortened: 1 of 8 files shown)


static/js/modules/analytics.js
```diff
@@ -0,0 +1,52 @@
...
+  if (typeof ga === 'undefined') { return; }
+  var path = document.location.pathname;
+  if (document.location.search) {
+    var query = URI.parseQuery(document.location.search);
+    path += '?' + sortQuery(query);
+  }
+  ga('send', 'pageview', path);
+}
+
+function sortQuery(query) {
```
noahmanger: Not sure how to test this. @msecret would you mind just peeking at this function and let us know if it makes sense?

tests/unit/modules/analytics.js
```diff
@@ -0,0 +1,18 @@
...
+var chai = require('chai');
+var expect = chai.expect;
+
+var analytics = require('../../../static/js/modules/analytics');
+
+describe('analytics', function() {
+  it('sorts query parameters', function() {
+    var query = {
+      cycle: [2016, 2014],
+      sort: 'name',
```
noahmanger: What's this "sort" parameter? We don't have this yet do we?

noahmanger:
All makes sense to me. I just pinged Marco about taking a look at the `sortQuery()` function.

static/js/modules/election-lookup.js
```diff
@@ -222,6 +223,7 @@ ElectionLookup.prototype.search = function(e, opts) {
     self.serialized = serialized;
     if (opts.pushState) {
       window.history.pushState(serialized, null, URI('').query(serialized).toString());
+      analytics.pageView();
```
noahmanger: Does this get called if you search from the home page too?

tests/unit/modules/analytics.js
jmcarp: Nope. This is a test, and it's verifying that query parameters are sorted correctly--it doesn't matter what the values are.

static/js/modules/election-lookup.js
jmcarp: No, but submitting the election search form on the home page triggers a new page load, which sends the event to GA.

static/js/modules/analytics.js
jmcarp: To test manually, you can build the JS with `FEC_WEB_GOOGLE_ANALYTICS=true` and verify that the function is called--you can add a break point here, or open GA and watch the requests show up. Also would appreciate thoughts from @msecret.

noahmanger: Got it. Yeah, looks like it's working.

noahmanger:
I'm comfortable merging this, or we can wait for Marco to review. Fine either way.

noahmanger:
A thought that might merit some discussion: do we want to be able to track _overall_ which filters on any given page are most used? Will this way of tracking things make it easy for us to find that data?

jmcarp:
That could be useful to know. This patch adds a page view for each filter URL, but I'm guessing it wouldn't be straightforward to extract views per filter from GA, since every combination of filters is a distinct view (i.e., `?cycle=2012` and `cycle=2012&office=P` are completely different views). We could try adding a "custom dimension" to the data:

```
ga('set', 'filter', 'cycle');
```

But that doesn't seem right either--custom dimensions are sent along with regular page views, but we'd often want to send multiple values of the `filter` dimension with a single view, which might not be possible. Documentation on that features lives here: https://support.google.com/analytics/answer/2709828?hl=en.

This might be simpler to solve by downloading data from GA or the API Umbrella analytics periodically and counting up filters with a custom script. But I'm feeling a little out of my depth here--might be useful to get feedback from the analytics guild.

noahmanger:
Right. That makes sense. I was thinking that simply downloading the data could work. I'm going to merge.

PR #867, opened 2015-10-20
Feature/fix receipts link

jmcarp:
Adds a few improvements related to https://github.com/18F/FEC/issues/72. It turned out that the receipts link on the committee detail page included a "cycle" parameter, which isn't one of the filters for the browse receipts page. This caused an immediate URL change upon loading the browse page.

* Fix link to receipts per cycle from the committee detail page
* Don't change URL based on irrelevant query parameters

cc @noahmanger

Note: looking at this code reminds me that it's kind of hairy in places and definitely in need of tests. I'm going to take a crack at refactoring the filter logic later this week--planning to add the missing tests there.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "hbsfy": "2.2.1",
     "intl": "1.0.0-rc-4",
     "jquery": "2.1.4",
-    "jquery.inputmask": "3.1.63",
+    "jquery.inputmask": "3.2.2",
     "keyboardjs": "0.4.2",
     "leaflet": "0.7.3",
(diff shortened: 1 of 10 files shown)


noahmanger:
Only thing I noticed is the individual transaction links from the aggregates on election pages still have the `cycle` param.

jmcarp:
Good catch @noahmanger. If I'm reading correctly, the URLs on the election page should be fixed in #876, so this patch should be good to go.

jmcarp:
Patch updated @noahmanger.

jmcarp:
Patch updated.

Issue #858, opened 2015-10-18
So users don't misunderstand outside spending add a disclaimer to candidate pages.


LindsayYoung:
User testing revealed that showing independent expenditures on the candidate page without explanation confused users. 

If we can't add a definition, we should at least add a disclaimer. Something like, "This money is not directly given to or spent by the candidate." (I am sure @emileighoutlaw could spruce the phrasing up.)

emileighoutlaw:
Oh interesting thought! I think we probably should have two sentences, one about what the tab *is* and one about what the tab *isn't*. How about something like:

"This tab shows spending that opposes or supports this candidate. None of the funds are directly given to or spent by the candidate."

Can we say "outside spending" ? I'd like it even better as:

"This tab shows outside spending that opposes or supports this candidate. None of the funds are directly given to or spent by the candidate."

@jenniferthibault, do you think this would be a good tooltip?

jenniferthibault:
Actually, I think this will work just fine as regular body intro text. We use that in a few other tabs as well. 

For example, here's what the Financial summary tab looks like:

<img width="1273" alt="screen shot 2015-10-20 at 4 55 31 pm" src="https://cloud.githubusercontent.com/assets/11636908/10620880/7bf6953a-774b-11e5-8d60-93b41e30c808.png">

I'd imagine that text going in the same place under "Other Spending" here, which would make the page flow much better anyways
<img width="1277" alt="screen shot 2015-10-20 at 4 55 46 pm" src="https://cloud.githubusercontent.com/assets/11636908/10620899/87b1465e-774b-11e5-8a80-3ff19880c4c9.png">

With that in mind, @emileighoutlaw if you need a little more text to explain, you have the room, but I'm all for concise text if the above does the trick!

Let's see if I can figure out how to implement this. If not, I'll come back and tag Noah :)

Next steps:
 - [ ] Implement text on tab : @jenniferthibault

noahmanger:
Resolved by #868

PR #868, opened 2015-10-20
Other spending disclaimer

jenniferthibault:
Adds introductory body text to the other spending tab that shows what the tab is and isn’t. Addresses @emileighoutlaw ’s https://github.com/18F/openFEC-web-app/issues/858

Review:
@noahmanger 
Question: I noticed that the header structure for the Financial Summary tab didn’t match the structure of this Other Spending tab. I made them consistent, but if there was a reason for, please advise :)


noahmanger:
Oh thanks for catching that!

PR #876, opened 2015-10-21
Update URLs and columns in top-level election table.

### LINKED ISSUE #873: So that users understand how totals and sub-totals are generated, provide consistent and accurate linking to source data
## Background
FEC data includes two types of similar but almost-always-unequal data points (here's my attempt to label them as clearly as possible): Totals data and Itemized data. In turn, we display each of these types in two different ways each:

### 1. Totals data
A. **Reported totals:** These include things like "Total Receipts" and "Total Disbursements" for a particular reporting period (e.g. Q1 2015). These numbers are taken directly from a [Form 3](http://www.fec.gov/pdf/forms/fecfrm3.pdf) and displayed on the site in the details panel for filings:
image

B. **Combined totals:**
We can also add together these totals for multiple reports in a broader period, such as all reports filed in 2015 - 2016. These totals are displayed places like the financial summary tab:
image

The candidate page:
image

And the top part of the elections pages:
image

### 2. Itemized data
A. **Itemized transactions:** At the same time, committees report _itemized_ transactions via Schedule A and B forms. So (and this is a little oversimplified) a committee will report "Total receipts" as a single number in one place on the form, and then on the attached Schedules will report the itemized transactions that _should in theory_ add up to that total. The thing is, they rarely do, for various reasons.  We display these itemized transactions on the /receipts and /disbursements pages.

B. **Aggregated itemized transactions:** Just as we can add up the top level totals, we also run our own math on these itemized transactions to create what we call aggregates or sub-totals, such as total raised from individual states, different occupations, etc. These are visible on the sub-tabs on the committee page:

image

And at the bottom of the elections page:
image

## The problem
The fact that Totals Data and Itemized Data seem like they should be equal but fundamentally aren't poses a challenge. How can we build an interface where we have consistent behavior when linking to the raw data that underlies a number we show when in that underlying data is going to be different for similar looking numbers: for Totals (Reported or Combined) the source data is necessarily the Filings that the numbers were reported on; for Itemized data, the source data is the Receipts or Disbursements we're adding up in different ways to generate an aggregate.

If this sounds complicated and annoying, it's because it is :)

## Proposed rule
So here's me trying to distill what we just talked about into as concise a rule as possible:

1. **Totals data:** If we _cannot_ show a list of itemized transactions — such as when dealing with top-level totals like Total Receipts or Total Disbursements — **do not** make the number a link, but instead include a link nearby that says something along the lines of "View source forms". 

2. **Aggregated itemized transactions:** If we can generate a list of itemized transactions (receipts or disbursements) that constitute a sub-total / aggregate (e.g. all contributions from California; all contributions between $500 and $999), make number a link with a `title` attribute that says "View itemized transactions" on hover.

## Specific follow-up:
On the top comparison table on elections pages:
- [ ] Remove links from the "total receipts" and "total disbursements" columns as these should not link to itemized data
- [ ] Replace the "Most recent report" column with a "Source forms" column (not sure what the label should be) that links to /filings filtered down to the reports that make up that Combined Total.

jmcarp:
* Remove total receipts and disbursements links.
* Add links to source filings.

[Resolves #873]

This isn't quite ready for merge. Two outstanding questions:
* I've tentatively labeled the column that links to the base filings "Financial reports". @LindsayYoung and @emileighoutlaw, feel free to weigh in with a more precise name if this is too fuzzy.
* I'm not sure what text / icon to use for the new column linking to the F3 filings. Since we're now linking to a browse view of all relevant filings for a candidate, we can't use text like "May Quarterly 2016"--the text will be the same for each row. I'm using the placeholder text of "Link" for now, but that needs to change. Maybe a link-out icon would be useful here.

cc @jenniferthibault @noahmanger 

screen shot 2015-10-21 at 6 24 30 pm

### DIFF
--- a/static/js/pages/elections.js
+++ b/static/js/pages/elections.js
@@ -70,7 +70,7 @@ var electionColumns = [
         }
       );
       var anchor = document.createElement('a');
-      anchor.textContent = 'Link';
+      anchor.textContent = 'View';
       anchor.setAttribute('href', url);
       anchor.setAttribute('target', '_blank');
       return anchor.outerHTML;
--- a/static/js/pages/elections.js
+++ b/static/js/pages/elections.js
@@ -55,28 +55,29 @@ var electionColumns = [
     }
   },
   {data: 'party_full', className: 'all'},
-  {
-    data: 'total_receipts',
-    orderSequence: ['desc', 'asc'],
-    render: tables.buildTotalLink(['receipts'], function(data, type, row, meta) {
-      return {
-        committee_id: row.committee_ids,
-        cycle: context.election.cycle
-      };
-    })
-  },
-  {
-    data: 'total_disbursements',
-    orderSequence: ['desc', 'asc'],
-    render: tables.buildTotalLink(['disbursements'], function(data, type, row, meta) {
-      return {
-        committee_id: row.committee_ids,
-        cycle: context.election.cycle
-      };
-    })
-  },
+  tables.currencyColumn({data: 'total_receipts', orderSequence: ['desc', 'asc']}),
+  tables.currencyColumn({data: 'total_disbursements', orderSequence: ['desc', 'asc']}),
   tables.barCurrencyColumn({data: 'cash_on_hand_end_period'}),
-  tables.urlColumn('pdf_url', {data: 'document_description', className: 'all', orderable: false})
+  {
+    render: function(data, type, row, meta) {
+      var dates = helpers.cycleDates(context.election.cycle);
+      var url = helpers.buildAppUrl(
+        ['filings'],
+        {
+          committee_id: row.committee_ids,
+          min_receipt_date: dates.min,
+          max_receipt_date: dates.max
+        }
+      );
+      var anchor = document.createElement('a');
+      anchor.textContent = 'Link';
+      anchor.setAttribute('href', url);
+      anchor.setAttribute('target', '_blank');
+      return anchor.outerHTML;
+    },
+    className: 'all',
+    orderable: false,
+  }
 ];
 
 function makeCommitteeColumn(opts, factory) {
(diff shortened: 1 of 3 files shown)


LindsayYoung:
How do people feel about "view source forms" instead of links, that is what we have on the candidate page. We can say something else, but I would like it to be consistent.

noahmanger:
I still find that wording a little awkward. Closer, but awkward.

jmcarp:
If you're talking about the column header, sounds good. If you're talking about the text in each row, I think we want something shorter, or an icon, since it's going to be repeated for every row--I don't want to repeat "View source forms" 30 times per page.

noahmanger:
+1

noahmanger:
What about: "Source reports" for the header and then just "View" for the links?

emileighoutlaw:
We have "View source forms" other places on the site, so I'm tempted to say we should keep things consistent and have "Source forms" for the header and "View" for the links. Thoughts?

jenniferthibault:
Source forms / View 
works for me for consistency.

@noahmanger if you feel strongly about "reports" instead of "forms" we should look at those other places and make them all consistent.

noahmanger:
So the term we use to describe them everywhere else is actually "Filings". Should it be "View source filings" everywhere?

jenniferthibault:
I'd be in favor of that....and it makes it clear that you're looking at something that was filed by the candidate or committee, not something generated by the FEC. @emileighoutlaw ?

emileighoutlaw:
Yes! I am definitely in favor of consistency. So we're going with

Source forms / View

? :+1:

noahmanger:
Works for me. @LindsayYoung ?

LindsayYoung:
Works for me. We will want to update #878 too.

noahmanger:
Made these changes in a PR to Josh's branch.

PR #878, opened 2015-10-22
Feature/big number links

### LINKED ISSUE #697: In order to show users where the data is coming from, link to the details from the summary number on the 2016 presidential election
I really like the summary number on the 2016 presidential election. Can we link the summary number to the detail data page?

image

LindsayYoung:
Adding links to forms that have the relevant information
for the Presidential and independent expenditure totals.

screen shot 2015-10-22 at 11 27 19 am

We don't show the IE spending on that page anymore, but
There isn't a better way to explain where we get that
number. Not having a link for it might imply that the
presidential link would give the information, which it
doesn't.

We could improve this in the future, when we have a
view for IEs.

resolves #697


jmcarp:
LGTM, but we should always use `url_for` over hard-coded URLs when possible. This wouldn't work on production as-is, since it always points to the dev proxy.

LindsayYoung:
Thanks for catching that! I will update it now!

PR #881, opened 2015-10-22
Handle multiple null values for state and district.

jmcarp:
Following a recent change in the API, missing values for `state` and
`district` have mostly changed from `null` to "US" and "00",
respectively. This patch checks for all possible missing values upon
receiving data from the API.

To check, verify that searching for elections by zip, state, and district work as expected.

h/t @noahmanger

### DIFF
--- a/static/js/modules/election-lookup.js
+++ b/static/js/modules/election-lookup.js
@@ -54,7 +54,7 @@ function formatResult(result, lookup) {
 
 function formatName(result) {
   var parts = [result.state, officeMap[result.office]];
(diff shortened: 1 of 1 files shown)


noahmanger:
Almost there. Searching by zip code is only returning house races; no senate or president.

jmcarp:
Derp. That looks like a distinct issue--investigating.

jmcarp:
@noahmanger that second issue should be fixed in https://github.com/18F/openFEC/pull/1292.

PR #889, opened 2015-10-23
Skip links on null employer and occupation.

jmcarp:
Since we don't currently expose filters to restrict transctions by null
employer or occupation, we don't have a good way to build links to
itemized browse views for aggregates with null values. This patch omits
links for these aggregate rows.

To verify: browse to a committee detail page, receipts tab, employer or occupation aggregates; there should be the usual links to aggregates for non-null employers and occupations, but no links for null values.

h/t @noahmanger

### DIFF
--- a/static/js/modules/tables.js
+++ b/static/js/modules/tables.js
@@ -104,20 +104,24 @@ function buildAggregateUrl(cycle) {
 function buildTotalLink(path, getParams) {
   return function(data, type, row, meta) {
     data = data || 0;
+    var params = getParams(data, type, row, meta);
     var span = document.createElement('div');
     span.setAttribute('data-value', data);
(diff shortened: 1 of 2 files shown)


noahmanger:
Just to verify: Null employer and occupations get displayed as `N/A` in these tables, and the links are still present to /receipts filtered by `N/A`. Is this right? The back button works in these scenarios.

jmcarp:
Null values will continue to be reported as "NOT REPORTED", and they won't have links at all, since there isn't a link we can build that would fetch transactions with null employers or occupations. This is based on Slack discussion yesterday, but happy to revise if there's a better approach.

noahmanger:
Oh, gotcha. I didn't realize "NOT REPORTED" was different than "N/A".

PR #891, opened 2015-10-23
Allow cross-origin requests to feedback endpoint.

jmcarp:
Necessary for feedback submission via the CMS application. Going
forward, we may want to restrict the origin to beta.fec.gov, but this is
fine for now. This becomes moot if the feedback widget is exported to a
separate application.

Required for https://github.com/18F/fec-cms/pull/98.

cc @LindsayYoung @noahmanger

### DIFF
--- a/openfecwebapp/views.py
+++ b/openfecwebapp/views.py
@@ -5,6 +5,7 @@ import furl
 
 from flask.views import MethodView
 from flask import request, render_template, jsonify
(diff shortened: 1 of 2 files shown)


noahmanger:
Is there a way to test this?

jmcarp:
Yep--submitting feedback issues from the CMS shouldn't work without this patch.

PR #895, opened 2015-10-26
Replace history on table initialization.

jmcarp:
This patch attempts to fix a longtime issue with permalinks for table
filters. On table initialization, we grab filter values from the query
string and set them in the filter fields. Because some fields change
their inputs (mostly fields with input masks, but potentially other
fields as well), this can trigger a change in the URL. In these cases,
the back button brings the user back to the initial URL, which again
triggers an immediate URL change when the input is changed by the filter
field. This patch sidesteps the issue by using `history.replaceState`
rather than `history.pushState` on load, meaning that we don't push a
new entry to this history.

To verify, enter a URL that will be automatically transformed on load,
such as http://localhost:3000/filings/?min_receipt_date=10/01/1985, and
verify that the back button works as expected.

cc @noahmanger

Note: this matters less once we stop using input masking, but it's theoretically a problem as long as any of our filters are capable of changing their values automatically. If we decide that we don't ever want filters to do this, we can roll back this change. Also, there's no need to include this in the launch--just wanted to make sure I got the proposal checked in before I forgot about it.

### DIFF
--- a/static/js/modules/tables.js
+++ b/static/js/modules/tables.js
@@ -1,6 +1,6 @@
 'use strict';
 
-/* global require, module, window, document */
+/* global window, document */
 
 var $ = require('jquery');
 var URI = require('URIjs');
@@ -222,7 +222,7 @@ function compareQuery(first, second, keys) {
   return !different;
 }
 
-function pushQuery(params) {
+function pushQuery(params, push) {
   var query = URI.parseQuery(window.location.search);
   var fields = filters.getFields();
   if (!compareQuery(query, params, fields)) {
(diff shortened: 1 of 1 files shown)


jmcarp:
Now that we've survived the launch, we can get back to this proposal. @noahmanger, have time to take a look? We can also ask for an outside code review if that would be useful--although not a lot of code changes here. Also, I didn't add unit tests, but that's because I'm planning to refactor and test all the table code soon.

static/js/modules/tables.js
```diff
@@ -222,7 +222,7 @@ function compareQuery(first, second, keys) {
   return !different;
 }
 
-function pushQuery(params) {
+function pushQuery(params, push) {
```
msecret: My only comment on this is just general advice that you can take or leave. I'd say the code would be a little more readable if it didn't have a boolean parameter but had two separate functions to call. I often consider it a potential code smell when theres a boolean parameter.

I'll leave it to you if you want to change that or not.

msecret:
LG. left more of a suggestion

noahmanger:
This works for me. I'll merge unless you want to address Marco's comment.

static/js/modules/tables.js
jmcarp: I'm less persuaded by the no-booleans argument for languages that support named function parameters:

```python
my_function(flag_1=True, flag_2=False, ...)
```

And ES6 will sort of support this syntax with parameter destructuring. But, since we're not using ES6 here (yet!), agreed--I'll update the patch.

jmcarp:
Patch updated @noahmanger!

PR #896, opened 2015-10-26
Handle missing dates in totals.

jmcarp:
This fixes an uncaught exception on rendering Bernie Sanders's candidate
page. Thanks @jwchumley for reporting.

[Resolves https://github.com/18F/openFEC/issues/1304]

cc @LindsayYoung


LindsayYoung:
This is still erring out for me when there is a candidate with no information http://localhost:3000/candidate/P60014446/?cycle=2016

jmcarp:
Good call. Patch updated.

Issue #865, opened 2015-10-20
To improve the reliability of the site, improve test coverage and refactor as needed


jmcarp:
# Unit tests

Much of the client-side logic was written before the project started to use unit tests. As a result, most older javascript doesn't have unit tests, and isn't written in a format amenable to unit testing. This increases the time necessary to fix bugs and add features. The following features are in need of tests:
- [ ] Tab lists
- [x] Data tables
- [x] Table filters
- [ ] Table panels
- [ ] Election pages
- [ ] Miscellaneous utilities
- [ ] Accordions (in fec-style)
# Integration tests

We have a handful of integration tests implemented using Selenium that verify that all the components of the application work together correctly. These tests are potentially very useful, since we don't have a manual QA team, but we haven't updated them for a while, so there's a lot they don't cover. They're also written using the low-level Selenium API, meaning they're verbose and hard to read and write. I'd like to write Selenium tests using the "page object" pattern going forward: http://martinfowler.com/bliki/PageObject.html. This will help organize our Selenium tests and make the code more concise.

Pages to test:
- [ ] Election detail
- [ ] Election search
# Notes

None of this is urgent, but I think it's important that we unit-test a feature before we change it. This is relevant to the work that @jenniferthibault has been doing on the election page: I'd like to hold off on changes until we have tests, and since there are designs that are ready for implementation, adding tests here is a relatively high priority. Our accordions also don't have unit tests, but we also aren't planning to make major changes to them, so tests are less urgent there (although still worthwhile).

I can start working through the higher-priority testing items anytime, and if @LindsayYoung and @noahmanger have the time and interest, we can split up the work.

noahmanger:
Card: https://18f.storiesonboard.com/m/fec/!card/xqcdHVJ85BIaCznTwmqB7g

PR #900, opened 2015-10-28
[WIP] Feature/refactor filter panel

jmcarp:
Much of the code for the filter panel on our data table views was written a while back. It's mostly unstructured and (in my experience) a challenge to debug. It's also untested, and not particularly easy to test. This patch refactors the filter code to make it amenable to unit testing, and unit tests the updated code.

* Refactor unstructured filtering code into `FilterPanel`, `FilterSet`, and `Filter` widgets, plus subclasses
* Refactor cycle select code into `CycleSelect` widget, plus subclasses
* Update logic in tables.js accordingly
* Add unit tests for all new widgets

Part of #865.

This is ready for review. I'd like @noahmanger to take a look, since I wound up dropping some old code that seemed not to be used (e.g. accordions inside filter panels), but this should be verified. I'm also hoping to get a review from @msecret or @shawnbot, if either of you has time.


jmcarp:
@msecret or @shawnbot: would either of you be able to look this over in the next week or so? I'm going to cite @kaitlin and @micahsaul, who've said that cross-team code reviews are copacetic, and not a problem for billing.

static/js/modules/cycle-select.js
```diff
@@ -0,0 +1,62 @@
...
+var URI = require('URIjs');
+
+function CycleSelect(elm) {
+  this.$elm = $(elm);
+  this.$elm.on('change', this.handleChange.bind(this));
+}
+
+CycleSelect.build = function($elm) {
+  switch ($elm.data('cycle-location')) {
+  case 'query':
```
msecret: Are the case statements supposed to be indented? I can't find the rule in the airbnb js styleguide but I'd expect it to be.

static/js/modules/filter-panel.js
```diff
@@ -0,0 +1,68 @@
...
+  this.filterSet = new FilterSet(this.$form).activate();
+  if (!_.isEmpty(this.filterSet.serialize())) {
+    this.show();
+  }
+
+  this.adjust();
+}
+
+FilterPanel.prototype.adjust = function() {
+  if ($('body').width() > 768) {
```
msecret: What does the number 768 represent? Is it a breakpoint that's shared with the CSS?

static/js/modules/filter-panel.js
```diff
@@ -0,0 +1,68 @@
...
+    this.show();
+  } else if (!this.isOpen) {
+    this.hide();
+  }
+};
+
+FilterPanel.prototype.show = function() {
+  this.$body.addClass('is-open');
+  this.$toggle.addClass('is-active');
+  this.$toggle.find('.filters__toggle__text').html('Hide filters');
```
msecret: If the style of the filter panel changes and it uses a different class from `filters__toggle__text`, would this break?

I like separating my classes that are for JS hooks from my classes that are for styles, but will also respect what has been used and works in this codebase.

static/js/modules/filter-panel.js
```diff
@@ -0,0 +1,68 @@
...
+  accessibility.restoreTabindex(this.$form);
+  $('body').addClass('is-showing-filters');
+  this.isOpen = true;
+};
+
+FilterPanel.prototype.hide = function() {
+  this.$body.removeClass('is-open');
+  this.$toggle.removeClass('is-active');
+  this.$toggle.find('.filters__toggle__text').html('Show filters');
+  $('#results tr:first-child').focus();
```
msecret: I could see this selector being a potential problem in that it's accessing global information (a #results selector) and is using an ID, so you can only have one on a page.

That being said, the filter panel doesn't sound like a component that would be included twice and "results" is pretty straightforward so I don't think this is a required fix.

static/js/modules/cycle-select.js
jmcarp: I've seen both styles--for example, https://contribute.jquery.org/style-guide/js/#switch-statements. But now that I look at style recommendations, I also see that folks generally don't like `switch` statements, since they're easy to mess up. I'll just use a regular conditional here.

static/js/modules/filter-panel.js
jmcarp: Yep. This is a more general issue. I don't think we need to address it in this patch, but I filed a separate issue at https://github.com/18F/openFEC-web-app/issues/924.

static/js/modules/filter-panel.js
jmcarp: Agreed, will change. For context, this patch is basically rearranging old code into classes and modules, so some bad patterns like this were carried over. Thanks for catching--I'm sure there's more work to do on this front @noahmanger.

static/js/modules/filter-panel.js
jmcarp: Agreed, I'll add the selector of the focus target to `options`.

jmcarp:
Made a few updates based on feedback from @msecret. Before this gets merged, I'd still like to get some feedback from @shawnbot (if you still have time), and I'm hoping @noahmanger can help verify that everything's still behaving as expected.

static/js/modules/filter-set.js
```diff
@@ -0,0 +1,65 @@
...
+var $ = require('jquery');
+var _ = require('underscore');
+var URI = require('URIjs');
+
+var Filter = require('./filters').Filter;
+
+var KEYCODE_ENTER = 13;
+
+function FilterSet(elm) {
+  this.$elm = $(elm);
```
noahmanger: Super small thing, but haven't we been going with `this.$body` for the top-most selector?

static/js/modules/filters.js
```diff
@@ -1,217 +1,122 @@
...
 
-var prepareValue = function($elm, value) {
-  if ($elm.attr('type') === 'checkbox') {
-    return $.isArray(value) ? value : [value];
+  this.name = this.$input.eq(0).attr('name');
+  this.fields = [this.name];
+}
+
+Filter.build = function($elm) {
+  if ($elm.hasClass('date-choice-field')) {
```
noahmanger: We should prefix this class with `js-`.

static/js/modules/filters.js
```diff
@@ -1,217 +1,122 @@
...
+
+Filter.prototype.handleClear = function() {
+  this.setValue();
+  this.$input.focus();
+};
+
+Filter.prototype.handleKeydown = function(e) {
+  if (e.which === KEYCODE_ENTER) {
+    e.preventDefault();
+    this.$input.change();
```
noahmanger: Not sure if it's caused by this function, but I noticed that when hitting enter on a regular text filter it triggers a form submit / page reload. Same with clicking "Apply filters". This does not seem to be the case on master.

templates/partials/filters/filter-buttons.html
```diff
@@ -1,4 +1,4 @@
-<div class="filter">
+<div>
```
noahmanger: `.filter` is a styling class that adds the necessary amount of margin / padding to this element. Can we keep it here? If it's controlling js, we should just add a separate `.js-` class.

static/js/modules/filters.js
```diff
@@ -1,217 +1,122 @@
...
-      $('body').addClass('is-showing-filters');
-    }
+DateFilter.prototype.handleRadioChange = function(e) {
+  var $input = $(e.target);
+  if (!$input.is(':checked')) { return; }
+  if ($input.attr('data-min-date')) {
+    this.$minDate.val($input.data('min-date'));
+    this.$maxDate.val($input.data('max-date'));
+  }
+  this.$minDate.focus();
```
noahmanger: Can we keep focus on the radio? This will allow users to more easily make selections with the keyboard.

noahmanger:
Made a couple small comments but functionally this is working for me.

templates/partials/filters/filter-buttons.html
jmcarp: Sure. Sorry, looks like I was overzealous in trying to clear out classes we weren't using anymore!

static/js/modules/filters.js
jmcarp: Good catch.

static/js/modules/filter-set.js
jmcarp: :+1:

static/js/modules/filters.js
jmcarp: Sure, that's just a holdover from the existing code. We must have wanted it at some point, but I don't remember why.

static/js/modules/filters.js
jmcarp: Ah, forgot to bind the `handleKeydown` callback. Fixed that, and changed the type of the apply button from "submit" to "button".

noahmanger:
Noticed that date filters are generating weird URLs. Notice the `date=on` param:

`http://localhost:3000/receipts/?date=on&min_date=01-01-2015&max_date=12-31-2015`

PR #902, opened 2015-10-28
Hotfix/fix disbursement columns

noahmanger:
The demo for tomorrow runs on a smaller width desktop screen, so it's painfully obvious how the disbursement and receipt columns ran into each other. This adjusts some column widths to mitigate the worst cases. It's not a complete fix for DataTables wonky column width rendering, but it should help.

Also, I found an unused part of the template in the disbursements.hbs file that was showing up. I think we just never implemented it. So I removed.


noahmanger:
Oops, meant to file against master.

PR #913, opened 2015-10-29
Skip basic auth unless credentials set.

jmcarp:
Follow-up from taking the password down yesterday. We should use basic auth if credentials are found in the environment, not otherwise; no need to key on `FEC_WEB_TEST` or other envvars.

Note: this isn't urgent--I'm sending as a hotfix because it should be merged into master before the next deploy. Attention @LindsayYoung.


LindsayYoung:
LGTM

PR #919, opened 2015-11-04
Content update

emileighoutlaw:
Fixes some capitalization inconsistencies.
Changes `Financial summary  - ` to `Financial summary:` for ease of scanning.

[ci skip]


tests/unit/modules/election-lookup.js
```diff
@@ -108,7 +108,7 @@ describe('election lookup', function() {
     var titles = $rendered.map(function(idx, elm) {
       return $(elm).find('h3').text().trim();
     }).get();
-    expect(titles).to.deep.equal(['US President', 'NJ Senate', 'NJ House District 09']);
+    expect(titles).to.deep.equal(['US president', 'NJ Senate', 'NJ House District 09']);
```
noahmanger: This is a test, not the actual code. Do we not want "president" capitalized in the results here?

image

emileighoutlaw: Technically for style, it should be lowercase `president`.  But I think you raise a really good point about consistency — the headers would be more jarring if they didn't look the same. I changed it back to the `President`.

templates/partials/committee/financial-summary.html
```diff
@@ -5,14 +5,14 @@
     <div class="content__section">
       <div class="section__heading">
         <h2 class="heading__title" id="section-1-heading">
-          Financial summary - {{ cycle|fmt_year_range }}
+          Financial summary: {{ cycle|fmt_year_range }}
```
noahmanger: Good call. This is a change that we should make on all of the tabs on candidate, committee and election pages.

emileighoutlaw: Good call. Think I got that change made both other places now.

noahmanger: I see them but can just make the changes myself.

emileighoutlaw: Arg! Why couldn't I find them!! :( Thanks, Noah.

noahmanger:
Test failures aren't related to this patch. Merging.

PR #927, opened 2015-11-05
Miscellaneous updates to election map.

jmcarp:
* Correctly render ZIP codes in DC
* Cancel pending AJAX request on new search
* Fix bug such that maps didn't update on selecting a state without
  changing search criteria

This fixes a few minor bugs that I noticed on the election search widget. To test manually, verify that the relevant production bugs don't happen with this branch checked out: DC ZIP codes should render, and clicking a state after clicking and zooming out again should re-highlight the state. This one's a little tricky to replicate--you do it like this:

* Open the election search page
* Click on a state
* Zoom out until all states are rendered in color on the map
* Click on the same state again

In production, this *doesn't* re-highlight the state as it should; under this patch, it does.

cc @noahmanger @msecret

### DIFF
--- a/static/js/modules/election-lookup.js
+++ b/static/js/modules/election-lookup.js
@@ -133,7 +133,7 @@ function getDistrictPalette(scale) {
 
 var ElectionFormMixin = {
   handleZipChange: function() {
-    this.$state.val('');
+    this.$state.val('').change();
     this.$district.val('');
   },
 
@@ -165,6 +165,8 @@ _.extend(ElectionLookup.prototype, ElectionFormMixin);
 ElectionLookup.prototype.init = function() {
   this.districts = 0;
   this.serialized = {};
+  this.results = [];
+  this.xhr = null;
 
   this.$form = this.$elm.find('form');
   this.$zip = this.$form.find('[name="zip"]');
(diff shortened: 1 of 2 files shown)


static/js/modules/election-lookup.js
```diff
@@ -214,16 +216,27 @@ ElectionLookup.prototype.search = function(e, opts) {
...
-    });
-    self.serialized = serialized;
-    if (opts.pushState) {
-      window.history.pushState(serialized, null, URI('').query(serialized).toString());
-      analytics.pageView();
+  if (self.shouldSearch(serialized)) {
+    if (!_.isEqual(serialized, self.serialized)) {
+      // Requested search options differ from saved options; request new data.
+      self.xhr && self.xhr.abort();
+      self.xhr = $.getJSON(self.getUrl(serialized)).done(function(response) {
```
msecret: Does this need error handling for failed requests?

msecret:
One comment, but LG, feel free to merge.

static/js/modules/election-lookup.js
jmcarp: Good point, but maybe beyond the scope of this patch. I'll add this to https://github.com/18F/openFEC-web-app/issues/733--this may be a good one to work on when I get around to refactoring the data table code.

noahmanger:
Finally checked functionality and it works. Merging.

PR #929, opened 2015-11-06
[WIP] Log client and server errors to Sentry.

jmcarp:
Note: Not ready for merge; this is just for evaluating Sentry on CF!

This patch adds error reporting with Sentry. We get near-real-time reporting of both client and server errors, which will make debugging javascript much easier. We're also able to capture more information about the application state when the error was logged because the logs live on CF, not a third-party service.

To verify:
* Ask @dlapiduz for a Sentry account (I don't have permissions to create new accounts just now)
* Set the `SENTRY_DSN` and `SENTRY_PUBLIC_DSN` environment variables--you can can values from https://sentry.18f.gov/docs/platforms/python/ and https://sentry.18f.gov/docs/platforms/javascript/, respectively
* Introduce an error to the code
* Verify that errors are captured in Sentry

I'm so excited about this!

cc @LindsayYoung @dlapiduz


jmcarp:
Ready for review @LindsayYoung. Our Sentry setup may not be ready for production yet, but I think we should definitely use it for develop. This patch only enables Sentry if the credentials are found, so let's not set them on the production environment until we're confident that our Sentry install is robust.

dlapiduz:
@jmcarp what can we do to make it more production ready?

jmcarp:
Lazy answer: let it run on our dev instances for a few weeks and make sure it doesn't fall over.
Possibly better answer: send a bunch of error reports programmatically and make sure it doesn't fall over.

It *should* be pretty hard to break Sentry--worst case, the celery queue gets clogged up, and errors take a while to surface. I just don't want requests from our app to Sentry to hang. @dlapiduz: think it would be worthwhile for me to run a quick load test on the Sentry instance? We'd turn off email notifications first :wink:

dlapiduz:
@jmcarp hopefully all we have to do is find the right scale for our sentry setup. That said we should definitely do some load testing.

If you have some time to do it, it would be great!

PR #931, opened 2015-11-06
[WIP] Contributions received

emileighoutlaw:
Edits `independent contributions` to `contributions received` on committee pages. 

cc @jenniferthibault 

[ci skip]

PR #932, opened 2015-11-06
Feature/fix grouped charts

jmcarp:
This patch fixes issues around labels, tooltip text, and chart bar colors for independent expenditure committees, as described in https://github.com/18F/FEC/issues/134 and https://github.com/18F/FEC/issues/135. It includes language changes from @emileighoutlaw in #931. Note: this does *not* add itemized independent expenditures, so we shouldn't close https://github.com/18F/FEC/issues/134 after merging.

Depends on https://github.com/18F/fec-style/pull/175.

To verify:
* Check out https://github.com/18F/fec-style/pull/175
* Visit the committee detail page for an independent expenditure committee (e.g.  https://beta.fec.gov/data/committee/C90015579/?cycle=2016)
* Verify that chart labels, tooltip text, and colors are correct, as described in https://github.com/18F/FEC/issues/134
* Verify that candidate committee charts are still correct

cc @noahmanger

### DIFF
--- a/templates/partials/ie-chart.html
+++ b/templates/partials/ie-chart.html
@@ -7,16 +7,16 @@
 <div class="row">
   <figure class="chart-container chart--r-d">
     <div class="chart__header">
-      <h4 class="chart__title">Independent contributions and independent expenditures</h4>
+      <h4 class="chart__title"> Contributions received and independent expenditures</h4>
       <ul class="chart__key list--flat">
-        <li class="chart__key__item"><span class="swatch"></span> Contributions received</li>
-        <li class="chart__key__item"><span class="swatch"></span> Independent expenditures</li>
+        <li class="chart__key__item"><span class="swatch"></span> Total contributions received</li>
+        <li class="chart__key__item"><span class="swatch"></span> Total independent expenditures</li>
       </ul>
     </div>
     {{ charts.chart_series_grouped(
       reports | reverse | list,
       ('independent_contributions_period', 'independent_expenditures_period'),
-      labels=('Contributions received', 'Independent expenditures'),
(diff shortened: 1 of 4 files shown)


templates/partials/ie-chart.html
```diff
@@ -2,26 +2,24 @@
...
 {% import 'macros/charts.html' as charts %}
-{% import 'macros/null.html' as null %}
-{% macro group_bar_tooltip(value, key) %}
-<div class="chart-series__bar__tooltip">
-  <h5 class="chart-series__bar__tooltip__title">{{ key|replace('_', ' ')|title }}</h5>
-  <span class="chart-series__bar__tooltip__value">{{ null.null( value | currency )}}</span>
-</div>
-{% endmacro %}
 
 {% if series_group_has_data(reports, ('independent_contributions_period', 'independent_expenditures_period')) %}
```
noahmanger: Just noting that the variable from the API (`independent_contributions_period`) must also be wrong, so we'll want to update that eventually.

noahmanger:
Looks good to me. I found a minor markup change that I filed as  PR to this, but I'm just going to merge this and file my PR separately.

templates/partials/ie-chart.html
jmcarp: Sorry, why must it be wrong?

noahmanger: It looks like we just mis-labeled the variable originally. This PR changed "Independent contributions" to just "Contributions" on the frontend (which is how it's labeled on the original forms), so I'm assuming that's what we want the variable to be called. It works fine now, but if we want to make the data line up with the form fields, that fix would be in order.

PR #938, opened 2015-11-10
[WIP] Feature/refactor data tables

jmcarp:
* [x] Rewrite table logic as class
* [x] Add unit tests for tables
* [ ] Rewrite table panel as class
* [ ] Add unit tests for panels
* [x] Update consumers of modified code

Part of #865.

Update: in the interest of smaller patches, I think we can limit this one to a refactor of the main data table logic, and add tests for the table detail panel in a separate patch. @noahmanger @msecret @shawnbot, anybody have time to review?


static/js/modules/tables.js
```diff
@@ -396,127 +349,190 @@ function adjustFormHeight($table, $form) {
   }
 }
 
-var defaultCallbacks = {
-  preprocess: mapResponse
+function OffsetPaginator() {}
+
+OffsetPaginator.prototype.mapQuery = function(data) {
+  return {
+    per_page: data.length,
```
msecret: Can this be changed so it's not camel case? Or is it because of any API?

static/js/modules/tables.js
```diff
@@ -396,127 +349,197 @@ function adjustFormHeight($table, $form) {
...
-        api.ajax.reload();
-      }
-    });
+
+  if (this.opts.useFilters) {
+    $(window).on('popstate', this.handlePopState.bind(this));
+  }
+
+  this.$body.css('width', '100%');
+  this.$body.find('tbody').addClass('js-panel-toggle');
```
msecret: Shouldn't this calls just be in the HTML?

static/js/modules/tables.js
```diff
@@ -396,127 +349,197 @@ function adjustFormHeight($table, $form) {
...
+};
+
+DataTable.prototype.ensureWidgets = function() {
+  if (this.hasWidgets) { return; }
+  this.$processing = $('<div class="overlay is-loading"></div>').hide();
+  this.$body.before(this.$processing);
+
+  if (this.opts.useHideNull) {
+    this.$hideNullWidget = $(
+      '<input id="null-checkbox" type="checkbox" name="sort_hide_null" checked>' +
```
msecret: Are there JS frontend templates on this project already?

static/js/modules/tables.js
```diff
@@ -396,127 +349,197 @@ function adjustFormHeight($table, $form) {
...
+      '<label for="null-checkbox" class="results-info__null">' +
+        'Hide results with missing values when sorting' +
+      '</label>'
+    );
+    var $paging = this.$body.closest('.dataTables_wrapper').find('.results-info--top');
```
msecret: `.results-info--top` selector should be JS specific, not style specific.

static/js/modules/tables.js
jmcarp: Right, this is part of the backend pagination API and shouldn't be changed.

static/js/modules/tables.js
jmcarp: The `tbody` isn't in the HTML--it's drawn by DataTables on AJAX complete. I'm not sure about the `width` call. @noahmanger, do you remember where that came from? This refactor basically takes our existing code and bundles it up into classes, including some bits, like this, that might benefit from more revision.

jmcarp:
Patch updated. Got time help verify behavior @noahmanger? Refactoring untested code is always a little dicey, so I'd like to get another set of eyes on this before merging.

static/js/modules/tables.js
```diff
@@ -396,127 +351,192 @@ function adjustFormHeight($table, $form) {
...
+
+DataTable.prototype.fetch = function(data, callback) {
+  var self = this;
+  self.ensureWidgets();
+  if (self.filterSet) {
+    pushQuery(self.filterSet.serialize(), self.filterSet.fields);
+    self.filters = self.filterSet.serialize();
+  }
+  var url = self.buildUrl(data);
+  self.$processing && self.$processing.show();
```
shawnbot: I've always found this "shorthand if" style a little obtuse, and I'm surprised that jshint didn't pick up on it. I would prefer to see:

```js
if (self.$processing) {
  self.$processing.show();
}
```

or, if that's used often and you want to keep it DRY, create a new prototype method that does the same. Maybe something like:

```js
DataTable.prototype.toggleProcessing = function(processing) {
  if (this.$processing) this.$processing.toggle(processing);
};
```

jmcarp: JSHint didn't pick up on it because I told it not to. I don't mind short-circuit conditionals in simple cases like this--plus, the JSHint checker that would have complained here also catches lots of false positives with chai asserts. Anyway, I'm happy to use the standard conditional if if improves readability. Not a fan of single-line conditionals, though :stuck_out_tongue_winking_eye:

shawnbot: Yeah, I totally failed there by offering a change that violates a different linting rule. :laughing: :hand: (that's my best github emoji face-palm)

noahmanger:
This seems to be working well for me. All the tables are behaving as expected. I see tables.js still only has about 50% coverage. Is that just because the details panel tests are coming? Or do you want to still add more tests for the other parts?

jmcarp:
The core data table code is less than half of tables.js, and this patch doesn't touch the filter panel (trying to keep pull requests small). I am going to push a few more tests tonight though.

noahmanger:
It worked well for me, so I'm down to merge.

PR #939, opened 2015-11-11
[hotfix] Pin buildpacks.

jmcarp:
The Cloud Foundry team recently updated the node buildpack and the
default version of node installed. This causes deployments to fail on
buildpack compilation. This patch pins our node and buildpack versions
to the last known working releases.

Same as https://github.com/18F/openFEC/pull/1347.


LindsayYoung:
this looks fine to me

PR #940, opened 2015-11-11
Skip phantomjs install.

jmcarp:
Installing karma-phantomjs-launcher and its peer dependency, phantomjs,
cause deploys to Cloud Foundry to fail. This patch removes the phantomjs
launcher from the dependencies and manually installs within Travis.


LindsayYoung:
This looks fine to me.

PR #943, opened 2015-11-13
Feature/add ie to committee pages

LindsayYoung:
This adds independent expenditure tab for all committees, that can make independent expenditures. Most won't have any, in the future we might want to load that only if a committee has made expenditures. 

I did notice one error that I fixed. All committee pages should have filings, even if they are electioneering or communication cost.

@jmcarp I would like to add a unit test, but I am not sure where to start.


noahmanger:
Re: conditionally showing tabs only when relevant, we have a story card for that: https://18f.storiesonboard.com/m/fec/!card/KyvZREVXsH_KvjnUD8RvAQ

jmcarp:
If this isn't urgent, I'd like to get #938 merged first. Once that's done, I'll handle the inevitable merge conflicts and review this.

jmcarp:
I sent a patch to update table invocations, which have changed since we merged #938.

@LindsayYoung: which committees can make independent expenditures? This patch shows the expenditures panel for all committees other than Presidential committees--is that right?

LindsayYoung:
I meant to exclude principal committees, rather than presidential. I will double check that coding now.

jmcarp:
This LGTM. @noahmanger and/or @emileighoutlaw, want to do a quick functional / language review before merge?

@LindsayYoung: we've been writing unit tests for isolated JS modules, but not for JS pages. That's more the domain of functional testing, which still lives in the python selenium tests, but which we should probably reevaluate at some point.

templates/partials/committee/independent-expenditures-tab.html
```diff
@@ -0,0 +1,32 @@
+{% import 'macros/cycle-select.html' as select %}
+
+<section class="main" id="section-6" role="tabpanel" aria-hidden="true" aria-labelledby="section-6-heading">
+  <div class="container">
+  <div class="content__section">
+    <div class="section__heading">
+      <h2 class="heading__title">
```
noahmanger: This h2 needs `id="section-6-heading"`.

noahmanger:
Functionality and code-wise looks good. I made a few minor fixes to the tab template.

jmcarp:
LGTM. Anybody else want to weigh in before merge?

LindsayYoung:
I think this can be merged.

noahmanger:
Do we want to add the conditional logic to the committee pages first though? Merging anything into develop now puts it in the next release.

LindsayYoung:
We can add more logic later. This corrects an error- all committee pages should have filings. It also only add the new IE tab for committees that could possibly have IEs.

noahmanger:
Oh cool. I'll take another look and then merge.

static/js/pages/committee-single.js
```diff
@@ -180,6 +180,12 @@ var disbursementRecipientIDColumns = [
   }
 ];
 
+var expendituresColumns = [
+  tables.currencyColumn({data: 'total'}),
+  columns.supportOpposeColumn,
+  tables.committeeColumn({data: 'candidate', className: 'all'}),
```
noahmanger: This needs to be ` tables.candidateColumn({data: 'candidate', className: 'all'}),`

LindsayYoung: That makes sense!

static/js/pages/committee-single.js
```diff
@@ -356,6 +362,19 @@ $(document).ready(function() {
         })
       );
       break;
+    case 'independent-expenditure-committee':
+      path = ['committee', committeeId, 'schedules', 'schedule_e', 'by_candidate'];
+      tables.DataTable.defer($table, {
+        path: path,
+        query: query,
+        columns: expendituresColumns,
+        // Order by receipt date descending
```
noahmanger: Super small thing, but we should remove this comment because it's not actually what's happening.

noahmanger:
Just a couple things but then looks good.

static/js/pages/committee-single.js
LindsayYoung: adding a function for that now

noahmanger:
I can't tell why these tests are failing. Is it just Travis?

LindsayYoung:
The website is working for me locally and looks fine. The failing tests are saying that it needs pandas, but that is not a requirement for this repo. I tried re-running on travis but it still failed. I'm trying to figure it out locally.

noahmanger:
Any insight into this?

noahmanger:
Merged in from develop and tests are passing. But it looks like maybe a recent API change has it so that the candidate and committee links in ie tables are no longer showing up. I see that this is fixed in the upcoming candidate-election-aggregate branch, so I'm not going to fix it here. So let's wait to merge this until that is merged.

noahmanger:
So this _works_ now, but in combination with the new styles introduced for the cycle update, it makes the tab list run off the page:
image

@jenniferthibault do you have any ideas? We could add a horizontal scroll (this is how we handle it on mobile) but it would show a scrollbar on windows machines. Alternatively, maybe @emileighoutlaw we change "Receipts from individuals" to something shorter? That would buy us a little space, but not fully solve it.

emileighoutlaw:
I think we're pretty stuck with "Receipts from individuals," at least for now: Both `individuals` and `receipts` are words with specific legal definitions at the FEC. We can sometimes write around it, but only by adding lots more words. :/

PR #949, opened 2015-11-14
Updates the feedback template with new field names

noahmanger:
Makes the necessary changes to work with the language edits in https://github.com/18F/fec-style/pull/182

Not sure where to make these changes on the CMS.


jmcarp:
All GitHub issues are submitted through the data site at the moment, so there's no corresponding text to edit in the CMS.

PR #950, opened 2015-11-16
[WIP] Feature/candidate election aggregate

jmcarp:
Show election aggregates on candidate detail pages.

* [x] Implement election and cycle select
* [x] Optionally aggregate candidate totals
* [x] Optionally aggregate candidate other spending
* [x] Write unit tests for new code

Depends on https://github.com/18F/openFEC/pull/1352.


static/templates/electionCycles.hbs
```diff
@@ -0,0 +1,20 @@
+<fieldset class="js-subcycle-select toggles">
+  {{#each this}}
+  <label>
+    <input
+        type="radio"
+        name="cycle-toggle"
+        value="{{this.max}}"
+        class="panel-toggle-control"
```
noahmanger: The inputs don't need a class.

static/templates/electionCycles.hbs
```diff
@@ -0,0 +1,20 @@
+<fieldset class="js-subcycle-select toggles">
+  {{#each this}}
```
noahmanger: Small thing, but this should have a legend right after fieldset:
`<legend class="label">Time period:</legend>`

static/templates/electionCycles.hbs
jmcarp: Oh, I modeled the markup after the toggles on the committee detail page. Do we not need that class there either? I can get rid of them all if they don't do anything.

jmcarp: Never mind, got it.

noahmanger: Yeah looks like that's for the js so should be prefixed with `.js-`.

static/templates/electionCycles.hbs
noahmanger: Yep.

Issue #750, opened 2015-09-23
add net cycle to date for presidental


LindsayYoung:
Presidential net cycle to date figures got updated. This is relevant for the filings detail panel.

changed in 18F/openFEC/pull/1213

- this is not urgent

PR #952, opened 2015-11-17
Fixing presidential net values

noahmanger:
Resolves https://github.com/18F/openFEC-web-app/issues/750

Ping @LindsayYoung . This look right?

### DIFF
--- a/static/templates/filings/candidate.hbs
+++ b/static/templates/filings/candidate.hbs
@@ -56,14 +56,28 @@
       <td class="panel__term">Ending cash on hand</td>
       <td class="panel__data">{{{currency cash_on_hand_end_period}}}</td>
     </tr>
+    {{#if net_contributions_period}}
     <tr>
       <td class="panel__term">Net contributions</td>
       <td class="panel__data">{{{currency net_contributions_period}}}</td>
     </tr>
+    {{else}}
+    <tr>
+      <td class="panel__term">Net contributions</td>
+      <td class="panel__data">{{{currency net_contributions_cycle_to_date}}}</td>
+    </tr>
+    {{/if}}
+    {{#if net_operation_expenditures_period}}
     <tr>
       <td class="panel__term">Net operating expenditures</td>
       <td class="panel__data">{{{currency net_operating_expenditures_period}}}</td>
     </tr>
+    {{else}}
+    <tr>
+      <td class="panel__term">Net operating expenditures</td>
+      <td class="panel__data">{{{currency net_operating_expenditures_cycle_to_date}}}</td>
+    </tr>
+    {{/if}}
     <tr>
       <td class="panel__term">Debts owed by committee</td>
       <td class="panel__data">{{{currency debts_owed_by_committee}}}</td>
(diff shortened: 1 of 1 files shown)


static/templates/filings/candidate.hbs
```diff
@@ -56,14 +56,28 @@
...
       <td class="panel__term">Net contributions</td>
       <td class="panel__data">{{{currency net_contributions_period}}}</td>
     </tr>
+    {{else}}
+    <tr>
+      <td class="panel__term">Net contributions</td>
+      <td class="panel__data">{{{currency net_contributions_cycle_to_date}}}</td>
+    </tr>
+    {{/if}}
+    {{#if net_operation_expenditures_period}}
```
jmcarp: "net_operating_expenditures_period"?

LindsayYoung: Sorry I did not write up that ticket with enough specificity. The change I was talking about was specific to presidential, and not all candidates and it is for the reports rather than totals. 

The new variable is called "net_operating_expenditures_cycle_to_date" and the name of it should be similar. 

It should go in the reports slide out of the filings view:
screen shot 2015-11-17 at 9 39 00 am

I will make a PR to this branch.

LindsayYoung: So we used to show all financial summary information for reports on the website, when we were showing most recent report. I remember doing a through check to make sure we had everything. It looks like we don't show that information anymore on the website. 

I guess we can close this. I do wish all the data was accessible somewhere on the website, perhaps we could do a reports view in the future?

noahmanger: @jmcarp thanks for the catch. Fixed it.

@LindsayYoung this template is for the reports slide out on the filings view (the screenshot you posted). This is the template that's used for candidate reports. I wrapped it in a conditional so that it will only show if the report doesn't have the `net_contributions_period`

noahmanger: As for your second point, i don't understand what's different. Isn't this report view showing all financial summary information from the report? If it's not, this is where it should be.

noahmanger:
Made my comments on a line comment that got removed when I changed it. So here they are again:

@jmcarp thanks for the catch. Fixed it.

@LindsayYoung this template is for the reports slide out on the filings view (the screenshot you posted). This is the template that's used for candidate reports. I wrapped it in a conditional so that it will only show if the report doesn't have the `net_contributions_period` / `net_operating_expenditures_period`.

As for your second point, i don't understand what's different. Isn't this report view showing all financial summary information from the report? If it's not, this is where it should be.

jmcarp:
What's the status of this patch @noahmanger @LindsayYoung?

LindsayYoung:
@noahmanger this is missing a lot of summary information- there are 3 different kinds of summary information from the respective forms of the Presidential, House/Senate and PACs/Parties. 

For example on a presidential candidate:

Here is what we have:
screen shot 2015-11-19 at 11 41 28 am

Here are all of the fields:
screen shot 2015-11-19 at 11 42 19 am

noahmanger:
Gotcha. Thanks for clarifying. Does this patch at least solve the initial problem you identified?

I think we should open up a new open issue for fleshing out the rest of the fields.

noahmanger:
Discussed this with @LindsayYoung and it should be good to go now.

PR #954, opened 2015-11-17
Feature/fix unit tests

jmcarp:
* Fix variable names for GitHub feedback and associated tests. I should've noticed this earlier, but #949 introduced inconsistent variable names within the feedback template, and between the template and the view.
* Fix a failing Selenium test.


noahmanger:
Oops. I thought I caught that too. Thanks.

PR #960, opened 2015-11-18
Add .about.yml.

jmcarp:
EOM

### DIFF
--- a/.about.yml
+++ b/.about.yml
@@ -59,13 +59,14 @@ services:
 #     name: Name of the license from the Software Package Data Exchange (SPDX): https://spdx.org/licenses/
 #     url: URL for the text of the license
 licenses:
-  betaFEC:
+  openFEC-web-app:
     name: CC0-1.0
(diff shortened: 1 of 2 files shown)


mbland:
LGTM, but other `team:` folks should review/merge.

.about.yml
```diff
@@ -0,0 +1,72 @@
+---
+# .about.yml project metadata
+#
+# Short name that acts as the project identifier (required)
+name: openFEC-web-app
+
+# Full proper name of the project (required)
+full_name: openFEC-web-app
```
noahmanger: Is this what's visible on the dashboard? Do we want a different public-facing name? @emileighoutlaw ?

jmcarp: No, this shouldn't appear on the dashboard--@mbland added a whitelist so that only pre-specified projects appear in the dashboard, not all projects with an .about.yml. That said, I'm happy to change the name here if folks would like to.

jmcarp:
ping @noahmanger

Issue #859, opened 2015-10-19
So users can better understand and apply filters, consider language edits to >$5K filter


emileighoutlaw:
Considering alternatives to this language, which could confuse non-expert users:

<img width="180" alt="screen_shot_2015-10-16_at_4 37 27_pm_360" src="https://cloud.githubusercontent.com/assets/10552702/10579937/f5709644-7648-11e5-9889-6af8c0a3ab80.png">

Is this correct? The three options under `Candidates raising more than $5,000` are:
- `Statutory candidate` (which means they've raised more than $5,000 for the current cycle?), 
- `Future candidate` (which means they’ve raised more than $5,000 but for a future cycle?), and 
- `Statutory candidate in prior cycle` (which means they raised more than $5,000 but for a past cycle?) 

cc @LindsayYoung @noahmanger @jmcarp

emileighoutlaw:
Jamming on this again. What are your thoughts on: 

> Candidates raising more than $5,000
- Prior to selected cycles (or election years)
- During selected cycles (or election years)
- After selected cycles (or election years)

AmyKort:
How about something more like this (edit me):

    Candidates raising more than $5,000

        For election held prior to selected cycle
        For selected election cycle
        For election held after selected cycle

@emileighoutlaw

emileighoutlaw:
I like this a lot more than what we have now.

Thoughts @jmcarp @noahmanger @LindsayYoung ?

noahmanger:
Better, but I still find the structure of the field label awkward. What about:

`Limit to candidates who raised more than $5,000`

AmyKort:
I think that's fine. I have a knee-jerk reaction to think contribution limit whenever I see "limit" in this context. Is that because I've worked here to long?

AmyKort:
Limited to candidates who raised more than $5,000

AmyKort:
?

noahmanger:
Hmmmm. I'll let Emileigh be the final say on verb conjugation. I like the imperative (is that right?) form more than past participle in this case, but I get your point. Maybe "Restrict to candidates...." ?

_Hoping I'm remembering my 10th grade english well_

_Instantly regretting getting into a grammar discussion with people with graduate degrees in English_

AmyKort:
Restricted to is perfect. Thank you!

AmyKort:
Sorry: restrict to.

I live in the past.

emileighoutlaw:
[...]

PR #961, opened 2015-11-19
Updating labels for candidate status filter

noahmanger:
This makes the language changes discussed here https://github.com/18F/openFEC-web-app/issues/859

image

@onezerojeremy can this get on the list for testing friday?

### DIFF
--- a/openfecwebapp/constants.py
+++ b/openfecwebapp/constants.py
@@ -209,11 +209,10 @@ amendment_indicators_extended = OrderedDict([
 ])
 
 candidate_status = OrderedDict([
-    ('C', 'Statutory candidate'),
+    ('C', 'For selected election cycle'),
 ])
 candidate_status_extended = OrderedDict([
-    ('F', 'Future candidate'),
+    ('P', 'For election held prior to selected cycle'),
+    ('F', 'For election held after selected cycle'),
     ('N', 'Not yet a candidate'),
-    ('P', 'Statutory candidate in prior cycle'),
-
 ])


noahmanger:
To clarify the point on testing: this is kind of a weird language issue and we just want to make sure that it makes sense to folks.

openfecwebapp/constants.py
```diff
@@ -209,11 +209,10 @@
 ])
 
 candidate_status = OrderedDict([
-    ('C', 'Statutory candidate'),
+    ('C', 'For selected election cycle'),
```
jmcarp: Maybe too nit-picky, but I'm not sure this is right. First of all, there isn't always a selected election cycle. Second, the candidate records are using the *most recent* candidate status value, not the candidate status value(s) of the selected cycle(s), if any.

noahmanger: Oh thanks for catching this. @emileighoutlaw and @amykort what do you think?

onezerojeremy:
@noahmanger yes of course!

openfecwebapp/constants.py
jwchumley: Not sure I understand the context.

noahmanger: Sorry, original discussion in this issue: https://github.com/18F/openFEC-web-app/issues/859

onezerojeremy:
From initial usability testing: 
No one had any problem filter by candidates who had raised more than 5k, though in retrospect it's clear that the test was not testing the thing it really needed to test, which was "MORE" dropdown, which included options that are not as clear. We aren't sure why these options are needed.   


see also #859

jmcarp:
Ping @noahmanger et al.: where's this at?

emileighoutlaw:
Thanks for pinging @jmcarp! We still have a lingering question:

If the candidate records (past, future, present) are always the most recent status value, why do we need these three options? Why not just have "candidates who raised more than $5,000" regardless of the cycle? Who would use the past/present/future filters, and how?

noahmanger:
Ping ping ping

emileighoutlaw:
:/ I still need an answer to:

If the candidate records (past, future, present) are always the most recent status value, why do we need these three options? Why not just have "candidates who raised more than $5,000" regardless of the cycle? Who would use the past/present/future filters, and how?

onezerojeremy:
I agree with @emileighoutlaw

noahmanger:
Still waiting on the API addition for this.

AmyKort:
I am out of the office until 02/18/2016.




Note: This is an automated response to your message  "Re: [openFEC-web-app]
Updating labels for candidate status filter (#961)" sent on 2/12/2016
5:27:28 PM.

This is the only notification you will receive while this person is away.

noahmanger:
This PR is so out of date I'm going to close it. Once the API update is done we can start fresh.

PR #962, opened 2015-11-19
Feature/add ie to committee pages

jmcarp:
* Merge develop into feature branch
* Use updated DataTables helpers (usage has changed since a few weeks ago), and invoke helpers consistently with other tables on the committee page


LindsayYoung:
Our JS is a lot more readable this way!

PR #964, opened 2015-11-19
Hotfix/pin fec style

jmcarp:
Pin fec-style to https://github.com/18F/fec-style/commit/d3c0fea4e31db64f886cea25aa7b39e2491b8bbd.

@noahmanger: can you take a minute and verify that the master branch looks right with this commit of fec-style? After this is merged, we'll update fec-style, deploy to npm as 1.0.0, and update the develop branch of this repo to install from npm.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -86,12 +86,16 @@ def cycle_end(value):
     return datetime.datetime(value, 12, 31)
 
 
+def nullify(value, *nulls):
+    return value if value not in nulls else None
+
+
 def get_election_url(candidate, cycle, district=None):
     return url_for(
         'elections',
         office=candidate['office_full'].lower(),
-        state=candidate['state'] if candidate['state'] != 'US' else None,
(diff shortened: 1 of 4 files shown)


__init__.py
```diff
@@ -86,12 +86,16 @@ def cycle_end(value):
...
+    return value if value not in nulls else None
+
+
 def get_election_url(candidate, cycle, district=None):
     return url_for(
         'elections',
         office=candidate['office_full'].lower(),
-        state=candidate['state'] if candidate['state'] != 'US' else None,
-        district=district or candidate['district'],
+        state=nullify(candidate['state'], 'US'),
```
noahmanger: Does this fix the broken senate links?

noahmanger:
Looks good to me. I also looked at commits after that point and it doesn't look like anything necessary is there. So all good.

jmcarp:
Merged manually.

PR #970, opened 2015-11-30
Feature/cfenv/slack

jmcarp:
Equivalent of https://github.com/18F/openFEC/pull/1379, plus move to py-cfenv for Cloud Foundry environment variables.


jmcarp:
Note: tests here should pass after https://github.com/18F/openFEC/pull/1379 is merged.

cc @LindsayYoung @noahmanger

Issue #971, opened 2015-12-01
Prototype download interaction


jmcarp:
Description by @noahmanger from https://github.com/18F/openFEC-web-app/issues/946#issuecomment-160769789:

## V1: Notification bar only
- Both instant and queued downloads will cause the notification bar to appear
- Once the download is ready, a link to the csv will appear in that bar 
- The bar will contain information on A) what's happening while the download processes and B) a suggestion to add more filters to get quicker downloads in the event of a queued download
- We will save the download URLs in user's localstorage so that when they return to the site on the same device, we will show the status of their download

* [x] Implement notification bar
* [x] Implement download button
* [x] Read from and write to local storage
* [x] Unit tests

PR #972, opened 2015-12-01
[WIP] Basic prototype of data downloads.

jmcarp:
Implements #971. Still needs tests, and revisions to style and content.


jmcarp:
@noahmanger: I fixed a few issues that I noticed while we were going over this yesterday, and I think it's ready for you and others to start work on style and content. You should be able to check out this branch to start looking at the behavior of pending downloads, but the downloads will never complete unless you run the download-utils branch of the API locally, or we push that branch to the dev space. The download-utils API branch requires running a few more processes than normal, so I can either walk you through that or deploy to dev.

static/js/modules/tables.js
```diff
@@ -12,11 +10,30 @@ require('datatables');
...
 var simpleDOM = 't<"results-info"ip>';
+var browseDOM = '<"js-results-info results-info results-info--top"pilfr>' +
+                '<"panel__main"t>' +
+                '<"results-info"ip>';
+
+var DOWNLOAD_CAP = 100000;
+
+var CAP_EXCEEDED =
+    'Exports are limited to ' +
+    DOWNLOAD_CAP +
```
noahmanger: Just noticed that this needs to be formatted (it was previously formatted with the Handlebars helper).

PR #987, opened 2015-12-12
Feature/itemized-IEs

noahmanger:
This adds a browse view for itemized independent expenditures at /independent-expenditures, and adds aggregate links to the IE tables on single candidate and election pages. (Will add for single committee pages once we merge that in).

Everything largely works, with the exception of the support / oppose filter, which doesn't seem to be supported by the API. 

Would love feedback on the proper columns, filters and the fields to include in the details panel. Here's what it looks like now:

image

image


noahmanger:
Updated with links from aggregates and the `is_notice

noahmanger:
Also just added filing and report info to the details panel, but definitely want verification that it's using it in the right way. Example:
image

noahmanger:
@LindsayYoung could you take a look at this today? Mainly I'd like your eyes on if it includes all the necessary data and if it uses it all in the right way (especially in the details panel).

jmcarp:
@noahmanger: looks like there are merge conflicts here. Can you resolve and ping @LindsayYoung when this is ready for review?

noahmanger:
I'll take a look.

LindsayYoung:
Sorry, I missed the first ping. 

Looks pretty good. I would add committee info in the details panel.

noahmanger:
More than what's currently in there? That most recent screenshot is incomplete.

LindsayYoung:
I looked at the code for the columns headers and those looked good. I did just look at the screen shot for the panel, disregard my comment if everything is there.

Also, can we configure hound to ignore non-camel case variable names?

noahmanger:
Conflicts resolved. And +1 to having hound ignore the non-camel-case names.

jmcarp:
@LindsayYoung: want to verify that the panel has the fields you think it should? If anything's missing, it's probably quickest to add fields during merge.

Also, you should both feel free to edit `.jshintrc` if you want to update the linter settings.

noahmanger:
Just added the export button too. Is the API all set up to handle exports for this table?

jmcarp:
Ping @LindsayYoung: anything left to do here? I sent in a little patch to the API to enable downloads on IEs at https://github.com/18F/openFEC/pull/1451.

LindsayYoung:
I will take a look now, thanks.

PR #989, opened 2015-12-15
Fix election title and link for off-year periods.

jmcarp:
Currently, viewing off-year periods builds incorrect election titles and
links: for example, viewing Barack Obama in 2010 builds a link called
"2010 Presidential Election". This patch builds the correct election
year and uses it for both the title and link.

### DIFF
--- a/openfecwebapp/views.py
+++ b/openfecwebapp/views.py
@@ -92,6 +92,10 @@ def render_candidate(candidate, committees, cycle, election_full=True):
         key=lambda pair: pair[0],
         reverse=True,
     )
+    tmpl_vars['election_year'] = next(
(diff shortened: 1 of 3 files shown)


noahmanger:
This fixes the problem of the incorrect election URLs, but I'm still seeing a problem on certain candidate pages. Take http://localhost:3000/candidate/S8OR00207/ : 
- Click "2015 - 2016" : it changes the `cycle` param to 2016.
- Now click "Full cycle": it changes `cycle` to 2020 and 404s

I can merge this if you want and we can address that next, or roll it in to here.

jmcarp:
@noahmanger that's an unrelated issue that I'm tracking in https://github.com/18F/openFEC/issues/1400, but I can include that fix here too once it's done.

PR #1000, opened 2015-12-17
[dontmerge] Hotfix/browse candidates election year

jmcarp:
Use election years for election years column.

Use the `election_years` field to populate the "Election years" column
in the browse candidates table, and to build the "cycle" query parameter
for links to candidates. This resolves a bug that was discovered during
the latest release such that candidate links pointed to the most recent
matching cycle, which in many cases was later than the most recent
election year, leading to a large number of effectively dead links.

h/t @leahbannon for spotting this


noahmanger:
LGTM

Issue #995, opened 2015-12-17
browse candidate to candidate page bug


leahbannon:
From browsing Candidates:

View a preview pane for more info on one candidate > scroll down towards the bottom of the page and click on a link to go directly to a different candidate page at the bottom of the page > you will jump back to the top of the page instead of going to the candidate page

PR #1005, opened 2015-12-17
Restore natural focus behavior on filter panel

noahmanger:
Resolves https://github.com/18F/openFEC-web-app/issues/995

### DIFF
--- a/static/js/modules/tables.js
+++ b/static/js/modules/tables.js
@@ -290,13 +290,6 @@ function modalRenderFactory(template, fetch) {
       e.preventDefault();
       hidePanel(api, $modal);
     });
-
-    /* Set focus to highlighted row on blurring anchors if tabbing out of the panel */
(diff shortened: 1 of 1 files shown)


noahmanger:
Gah, accidentally submitted without writing my explanation:

As #995 first noticed, if you have the filter panel open, clicking anywhere (not just the candidate link) would force focus to be set on the panel trigger button. 

This was because we had an event listener on `blur` from any `a` or `button` in the panel which would force focus to the panel trigger button. This was set up so tabbing through the elements of a panel would send you back to the row trigger, rather than the page controls below the table.

By removing this, it has the minor side-effect that that tab behavior now happens, however, the entire tool is entirely usable by keyboard still, and this solves the issue by not requiring you to double-click on things outside an open panel.

I imagine we could engineer a more complicated solution, but this seems like it works just fine.

PR #1006, opened 2015-12-18
Pass correct full election flag.

jmcarp:
Goes with https://github.com/18F/openFEC/pull/1411

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -379,6 +379,12 @@ def fmt_year_range(year):
 def fmt_state_full(value):
     return constants.states[value.upper()]
 
+@app.template_filter()
+def fmt_cycle_min_max(cycles):
+    if len(cycles) > 1:
+        return '{}–{}'.format(min(cycles), max(cycles))
+    return cycles[0]
+
 # If HTTPS is on, apply full HSTS as well, to all subdomains.
 # Only use when you're sure. 31536000 = 1 year.
 if config.force_https:
--- a/__init__.py
+++ b/__init__.py
@@ -207,7 +207,7 @@ def candidate_page(c_id, cycle=None, election_full=True):
     candidate, committees, cycle = load_with_nested(
         'candidate', c_id, 'committees',
         cycle=cycle, cycle_key='two_year_period',
-        election_full='true',
+        election_full=election_full,
     )
     if election_full and cycle and cycle not in candidate['election_years']:
         next_cycle = next(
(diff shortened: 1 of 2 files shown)


noahmanger:
This works as expected, but I have a UI thought:

image

In the full-cycle view, I think we should specify the cycles for which each committee was active. So in this case, note which one was 2013-2014 and which was 2015-2016. Or at the least make clear which one is currently active. 

@jenniferthibault and @LindsayYoung do you have thoughts?

noahmanger:
Also, sorry for not catching that earlier in the initial issue.

jenniferthibault:
That makes sense to me, thanks for bringing it up! This is the first I've seen of a candidate that had two principal campaign committees, and I see how specifying when each is active would be helpful. 

An idea on format:

HILLARY CLINTON FOR PRESIDENT
yyyy–yyyy Principal campaign committe

LindsayYoung:
Can we add some text that the finances are aggregated according to when the activity occurred and not what election the activity was for? I am afraid we are leading people astray. Committees can be paying off debts after the election and that doesn't help their current campaign.  

Having the date range for senate committees might occasionally be useful, I am not sure we need a lot of development around this edge case. But, if users want the date range, we should add it, perhaps that is something to test? There would be a tradeoff because there is already a lot of stuff on this page and more information would make it slightly busier.

jmcarp:
What are our next steps here? Can we merge this patch, or should it wait until more interface changes? If the latter, who can we assign to design / implement changes?

noahmanger:
I hear @LindsayYoung 's point about explaining that it's for the period there was activity and not the election, but I don't know that we need to worry about it. I think this communicates the main concept of "one of these is currently active, the other was active in the past":

image

@jenniferthibault I had to put the year range on a new line because it was causing the "principal campaign committee" to break onto a new line.

If this works, I have a patch that I can send a PR for.

noahmanger:
Pending the :+1: here's the PR https://github.com/jmcarp/openFEC-web-app/pull/19

emileighoutlaw:
This Q shouldn't hold a merge, but I'm wondering why we don't have the most recent principal campaign committee on top and cascade chronologically from there?

noahmanger:
Nice catch. I just reversed it (using the jinja `reverse` filter, but let me know if there's a better way to do it).

noahmanger:
Just realized that @jenniferthibault is out, so I think we can merge and adjust later if she has design feedback.

jenniferthibault:
What Emileigh said is what I would have said :) Looks good, the new line is totally fine. Merging.

PR #1007, opened 2015-12-18
Feature/extract filters

jmcarp:
See https://github.com/18F/fec-style/pull/211.


package.json
```diff
@@ -61,7 +62,7 @@
     "d3": "3.5.5",
     "datatables": "1.10.7",
     "drmonty-datatables-responsive": "1.0.6",
-    "fec-style": "1.1.0",
+    "fec-style": "git://github.com/jmcarp/fec-style#move-filters",
```
noahmanger: Did you forget to revert this?

jmcarp: I was planning to revert after we merged https://github.com/18F/fec-style/pull/211 (which I see you did) and released fec-style 1.2.0. If I hadn't pinned to my branch here, tests would've failed on travis. Cool if I bump the fec-style version and update it here?

noahmanger:
Ah, makes sense. Yep go for it.

jmcarp:
Updated fec-style and this patch.

PR #1008, opened 2015-12-18
Add treasurer filter to committee browse view.

jmcarp:
Depends on https://github.com/18F/openFEC/pull/1417

### DIFF
--- a/templates/partials/committee-filter.html
+++ b/templates/partials/committee-filter.html
@@ -5,6 +5,7 @@
 
 {% block filters %}
 {{ text.field('name', 'Name') }}
+{{ text.field('treasurer_name', 'Treasurer name') }}
 {% include 'partials/filters/election-years.html' %}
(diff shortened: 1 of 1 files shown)


templates/partials/committee-filter.html
```diff
@@ -8,5 +8,6 @@
 {% include 'partials/filters/election-years.html' %}
 {% include 'partials/filters/parties.html' %}
 {{ states.field('state') }}
+{{ text.field('treasurer', 'Treasurer name') }}
```
noahmanger: Can we just move this to after "Name"? It's the second column, so I think it makes sense there.

noahmanger:
LGTM, with one minor note.

templates/partials/committee-filter.html
jmcarp: Updated.

PR #1010, opened 2015-12-22
Adding link to candidates page from election

noahmanger:
- Adds a `office_code` variable which can be used to build the link to
the candidates browse page
- Adds a link on election pages to view all candidates who’ve filed for
an election

Might make sense to merge until after we've made the "cycle" / "Election years" filter switch, as this will need to change then.

Also, @emileighoutlaw and @jenniferthibault what do you think of language and display as so:

image

Resolves https://github.com/18F/openFEC/issues/1408

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -288,6 +288,7 @@ def elections(office, cycle, state=None, district=None):
     return render_template(
         'elections.html',
         office=office,
+        office_code=office[0],
         cycle=cycle,
         cycles=cycles,
         state=state,
--- a/templates/partials/elections/candidate-comparison-tab.html
+++ b/templates/partials/elections/candidate-comparison-tab.html
@@ -8,12 +8,12 @@
         <div class="row">
           <h3 class="results-info__title">Candidates in this election</h3>
         </div>
+        <div class="row u-padding--bottom">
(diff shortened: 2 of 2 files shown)


jenniferthibault:
Some thoughts for @noahmanger  and @emileighoutlaw :-) 
 - the Incumbent key is a little floaty right now, can we ground it closer to the table border, like so?
screen shot 2015-12-22 at 6 15 32 pm

- `Looking for all candidates who filed? Click here.` 
We've been eliminating things like "Click here" in favor of links that are more embedded in the action of the sentence. Maybe instead something like "[View all candidates](#) who have registered in this election, even if they haven't filed financial reports." ?

- In the second line, `Totals reflect the sum total reported by all of a candidate's authorized committees` is it possible to eliminate the second "total" ? or is that necessary on technicality?

- What do you think about adding a glossary definition for `incumbent` ?

emileighoutlaw:
I like the content. Can I suggest a tiny tweak? Since sum already means "the aggregate of two or more numbers," I think we need to delete "total," or it's redundant.

noahmanger:
Updated:

image

jenniferthibault:
Looks good to me!

PR #1012, opened 2015-12-28
Hide committee contributors on earmarked receipts.

jmcarp:
As @PaulClark2 explained, the `contbr_id` field for earmarked receipts
describes the earmarking committee and not the contributor. This patch
uses the `contbr_nm` field and hides the `contbr_id` field in these
cases.

[Resolves https://github.com/18F/openFEC/issues/1426]
[Resolves https://github.com/18F/openFEC/issues/1410]
[Resolves https://github.com/18F/openFEC/issues/1308]

### DIFF
--- a/static/js/pages/receipts.js
+++ b/static/js/pages/receipts.js
@@ -16,7 +16,7 @@ var columns = [
     className: 'all',
     width: '200px',
     render: function(data, type, row, meta) {
-      if (data) {
+      if (data && row.receipt_type !== '15E') {
         return tables.buildEntityLink(
           data.name,
           helpers.buildAppUrl(['committee', data.committee_id]),
--- a/static/js/pages/receipts.js
+++ b/static/js/pages/receipts.js
@@ -16,7 +16,7 @@ var columns = [
     className: 'all',
     width: '200px',
     render: function(data, type, row, meta) {
-      if (data && row.receipt_type !== '15E') {
+      if (data && row.receipt_type !== helpers.globals.EARMARKED_CODE) {
         return tables.buildEntityLink(
           data.name,
           helpers.buildAppUrl(['committee', data.committee_id]),
--- a/static/js/modules/helpers.js
+++ b/static/js/modules/helpers.js
@@ -35,6 +35,14 @@ Handlebars.registerHelper({
   }
 });
 
+var globals = {
+  EARMARKED_CODE: '15E'
+};
+
+Handlebars.registerHelper('global', function(value) {
+  return globals[value];
+});
+
 function decodeAmendment(value) {
   return decoders.amendments[value];
 }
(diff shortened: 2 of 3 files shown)


LindsayYoung:
This works, on the table view, but the slide out panel still says ActBlue for name. 

What if in these cases, we showed the contributor name as the name, with no over-ride and then added a field called "Earmarked by: " that would show the committee, in this case ActBlue.

screen shot 2015-12-28 at 9 02 57 am

PaulClark2:
I like @LindsayYoung 's idea of adding "Earmarked by: "   I think that makes the information about the transaction clear.

static/js/pages/receipts.js
```diff
@@ -16,7 +16,7 @@ var columns = [
     className: 'all',
     width: '200px',
     render: function(data, type, row, meta) {
-      if (data) {
+      if (data && row.receipt_type !== '15E') {
```
noahmanger: We don't generally reference these types of codes in the frontend code, so what would you think about setting a constant like `EARMARKED = '15E'` to make this check a little clearer?

noahmanger:
I just had the one comment / question on this in the code, but otherwise LGTM.

jmcarp:
Can this be merged @noahmanger @LindsayYoung? Happy to revise if anybody has more recommendations.

noahmanger:
I _think_ this is good.. The only issue I'm seeing is that [this issue](https://github.com/18F/openFEC/issues/1410) still seems to be there. I'm seeing occupation and employer values in contributions that are listed as coming from Actblue. Do we just need to put those fields in the details panel inside the conditional?

jmcarp:
Could you send examples of what you mean?

noahmanger:
Sorry, was scrambling out the door when I sent that. Example: http://localhost:3000/receipts/?contributor_id=C00401224

screen shot 2016-01-12 at 10 24 40 am

jmcarp:
Ah, I see what you mean. FWIW, the "conduit total listed in agg. field" actually appears in the filing, even though it's clearly not actually an occupation--check out the first item in http://docquery.fec.gov/cgi-bin/fecimg/?201512179004291981. I don't know if filers regularly use the occupation field for whatever kind of information this is on earmarked contributions.

Anyway, yes, we can conditionally hide employer and occupation on non-individual receipts, although we wouldn't using the conditional that this patch adds to the template. The examples in the screenshot aren't 15E receipts--I think we'd have to use a separate check on `is_individual` instead. Is that how we should handle this?

LindsayYoung:
Not showing employer for non-individuals on the web app seems reasonable to me. (There is only so much we can do about weird filings.) 

Adding line number in the panel will help us debug these kinds of things and also be useful for power users. Now that I think about it, having a line number explanation field in the API would be really helpful and could be used for the panel.

jmcarp:
All right, occupation and employer should now be hidden for non-individual receipts. If we ever want to show the text that ActBlue fills in for employer on earmarked receipts, we can revert the last commit.

noahmanger:
LGTM

PR #1015, opened 2016-01-01
[WIP] Feature/datatable containers

noahmanger:
Makes the necessary changes to the browse pages to make use of the style and component refactor for the calendar (https://github.com/18F/fec-style/pull/213/).

- Adds a `.js-filter` class to filters, which separates behavior from the style, and solves a previously unnoticed bug where if multiple `.filter` `<fieldsets>` had inputs that shared the same `name` (such as the various report type filters), the clear button wouldn't work.
- Updates the classnames and layouts for the filter + data table pattern to be more generalized and shared with the calendar
- Adds a widget container for filter tags and eventually the export buttons
- Extracts the height adjusting method to the filterPanel component so it can be shared with the calendar page


noahmanger:
I know tests are failing, but this might be time to start having someone look at this @jmcarp or @LindsayYoung . 

Apologies that it got so big. Most of the file changes are very small changes across several different files.

templates/macros/filters/checkbox.html
```diff
@@ -33,3 +33,15 @@
   </fieldset>
 </div>
 {% endmacro %}
+
+{% macro checkbox_dropdown_multiple(name, label, selected={}, options={}) %}
```
jmcarp: Other than by reading the source, it's not completely clear how `checkbox_dropdown` and `checkbox_dropdown_multiple` are different. We might be able to think of clearer names, or add comments.

templates/partials/candidates-filter.html
```diff
@@ -11,4 +11,5 @@
 {{ states.field('state') }}
 {% include 'partials/filters/districts.html' %}
 {% include 'partials/filters/candidate_status.html' %}
+{% include 'partials/filters/hideNull.html' %}
```
jmcarp: Not opposed to this change, but why is it in this pull request?

templates/partials/filters/hideNull.html
```diff
@@ -1,5 +1,5 @@
-<div class="results-info__null">
-  <input id="null-checkbox" type="checkbox" name="sort_hide_null" checked>
+<div class="js-filter filter">
+  <input id="null-checkbox" type="checkbox" name="sort_hide_null" value="true" checked>
```
jmcarp: FWIW, if this toggle lives in the filter set, the `checked` attribute won't have any effect, since initial filter values are pulled from the URL on page load.

templates/partials/candidates-filter.html
noahmanger: I probably could have waited, but the new designs that arranged the tags, export buttons, and table info required this to move to a filter.

templates/partials/filters/hideNull.html
noahmanger: Oops, meant to do that. Updated.

templates/macros/filters/checkbox.html
noahmanger: Yeah, I agree. The idea is that these are used when there's multiple dropdown components used in a single filter. Unless you have an idea for a name, I'm just going to add a comment.

PR #1035, opened 2016-01-16
Add URL of local instance of site to README

evankroske:
I tracked down the default port of the local instances of the site and added it to the readme.


LindsayYoung:
:+1:

PR #1036, opened 2016-01-20
Link candidate committees to most recent applicable cycle.

jmcarp:
Currently, candidate detail pages link to related committees at the
current cycle. On full election views, related committees may have no
data for the selected cycle, leading to a 404. This patch resolves the
404 issue by linking to each related primary committee using the most
recent applicable cycle.

[Resolves https://github.com/18F/FEC/issues/195]

To verify: as described in https://github.com/18F/FEC/issues/195, browse to /data/candidate/P00003392/, and check that both primary committee links resolve. Because "Hillary Clinton for President" last reported in 2014, that link should go to the 2014 cycle; the more recently active "Hillary for America" should go to the 2016 cycle.

cc @LindsayYoung @noahmanger

### DIFF
--- a/openfecwebapp/views.py
+++ b/openfecwebapp/views.py
@@ -69,7 +69,6 @@ def render_candidate(candidate, committees, cycle, election_full=True):
     tmpl_vars = candidate
 
     tmpl_vars['cycle'] = cycle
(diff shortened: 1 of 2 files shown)


jenniferthibault:
Nice!! 
(I am not reviewing, just cheering this on)

PR #1060, opened 2016-02-01
Update candidate totals URL.

jmcarp:
Use URLs from https://github.com/18F/openFEC/pull/1473/.


noahmanger:
If this is included in #1073 does this still need to be open?

PR #1063, opened 2016-02-02
Update node, npm, and node buildpack.

jmcarp:
The latest version of node.js is 5.5.0, but we were pinning to 0.12.7. This doesn't match the version I've been using locally, which could lead to hard-to-debug issues. Note: I'd encourage everyone to use a recent version of node locally as well.


jmcarp:
Not urgent, but pinging @ccostino or @LindsayYoung for review.

ccostino:
@jmcarp:  I just took a look and this appears fine to me.  I'm assuming the same environment settings/setup with `pip` and `nvm` exist where needed as well Travis?

I definitely agree on the using a more recent Node.js version, if not the latest if possible.

/cc @LindsayYoung

jmcarp:
@ccostino: this patch updates `.travis.yml` to install node `stable`. We could pin to 5.5.0 instead. Travis is also using python 3.4. Also, are you saying you don't want to use the latest version of node? Would you prefer sticking to the lts version?

ccostino:
@jmcarp: I'd have to defer to those of you more involved with the project; is stability and long-term support very critical or is it okay to run the latest and greatest stable release?  I'm fine either way, but my answer depends on the needs of the project.  I'm sorry if I wasn't clear above, my personal preference would be the latest stable release and to keep what you have (`nvm install stable`).

If you pin to a specific version, you're just going to run into the same issue again that you're addressing now.

jmcarp:
On second thought, I think we do want to pin node versions. We're already pinning buildpacks, python versions, python dependencies, and node dependencies--definitely been burned by breaking changes in "minor" releases before. Yes, upgrading is a manual process, but at least things aren't as likely to break without warning.

ccostino:
That is definitely true, you run the risk of dealing with breaking changes suddenly if you're constantly pulling the latest version of something.  I've been burned by that as well and my guess is that is certainly something we'd want to mitigate as much as possible based on what I'm hearing.  A positive side effect of the manual upgrade process is that it also forces you to check the application and run it through its tests before just accepting a new version of a dependency, so it's possible to catch any issues prior to them being released in the wild.

The only thing to keep in mind is making sure that as project dependencies evolve and new versions come are released, we're taking the time to evaluate them and keep our project(s) up-to-date with them as well in a timely fashion.  Perhaps we don't upgrade right away, but we also don't fall very far behind.  This is more process than anything, though, and something that can be worked into the project backlog and sprints.

It sounds like we have our answer then in terms of what to do. :-)

PR #1064, opened 2016-02-02
Exclude notices from independent expenditures views.

jmcarp:
EOM

### DIFF
--- a/static/js/pages/elections.js
+++ b/static/js/pages/elections.js
@@ -30,7 +30,7 @@ var independentExpenditureColumns = [
         return {
           support_oppose_indicator: row.support_oppose_indicator,
           candidate_id: row.candidate_id,
-          is_notice: false,
+          // is_notice: false
         };
     })
   },
--- a/static/js/pages/independent-expenditures.js
+++ b/static/js/pages/independent-expenditures.js
@@ -66,6 +66,7 @@ $(document).ready(function() {
   new tables.DataTable($table, {
     title: 'Independent expenditure',
     path: 'schedules/schedule_e',
+    query: {is_notice: 'false'},
     panel: filterPanel,
     columns: columns,
     paginator: tables.SeekPaginator,
--- a/static/js/pages/candidate-single.js
+++ b/static/js/pages/candidate-single.js
@@ -23,7 +23,7 @@ var expendituresColumns = [
         return {
           support_oppose_indicator: row.support_oppose_indicator,
           candidate_id: row.candidate_id,
-          is_notice: false,
+          // is_notice: false,
         };
     })
   },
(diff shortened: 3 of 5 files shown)


static/js/pages/independent-expenditures.js
```diff
@@ -66,6 +66,7 @@ $(document).ready(function() {
   new tables.DataTable($table, {
     title: 'Independent expenditure',
     path: 'schedules/schedule_e',
+    query: {is_notice: 'false'},
```
noahmanger: If we're setting it here, do we even need to add it to the URLs from the aggregate links above?

noahmanger: Preference would be for just not exposing the parameter at all

jmcarp: I thought you wanted to make the change on the frontend. And yes, this needs to be added to all tables that hit `/schedules/schedule_e/`, unless you want to hide notices in some views but not others.

jmcarp: Lol, misread--no, we don't need to change the query for aggregates.

noahmanger: Ah I see what's going on now. But I think we still don't need those added, as those are the aggregate tables not itemized tables, and we already exclude notices from those aggregates (correct?)

jmcarp: Right, I was agreeing above but got pulled away. I revised the patch so that we only touch the itemized IE view. Thanks for the catch!

static/js/pages/elections.js
```diff
@@ -30,7 +30,7 @@ var independentExpenditureColumns = [
         return {
           support_oppose_indicator: row.support_oppose_indicator,
           candidate_id: row.candidate_id,
-          is_notice: false,
+          is_notice: false
```
noahmanger: Looks like you missed one place :)

static/js/pages/independent-expenditures.js
noahmanger: Great! Thanks

static/js/pages/elections.js
jmcarp: This one was already here--I just took out a comma. We want the aggregates links to include `is_notice=false` in the URL, right?

noahmanger: Oops, sorry. We can leave it because it was there and we'll want to use it in the future, but it won't actually make a difference now will it?

jmcarp:
Commented out references to `is_notice` in URLs, plus the filters on `is_notice`. That seem right @noahmanger?

noahmanger:
Sounds great. Thanks. Will merge after tests pass.

PR #1066, opened 2016-02-03
Feature/ie improvements

noahmanger:
Minor column re-arranging / re-labeling based on feedback.

### DIFF
--- a/static/js/pages/independent-expenditures.js
+++ b/static/js/pages/independent-expenditures.js
@@ -1,7 +1,6 @@
 'use strict';
 
 var $ = require('jquery');
-var _ = require('underscore');
 
 var tables = require('../modules/tables');
 var helpers = require('../modules/helpers');
@@ -16,7 +15,7 @@ var columns = [
   {
     data: 'committee',
     orderable: false,
-    className: 'all',
+    className: 'min-desktop',
     render: function(data, type, row, meta) {
       if (data) {
         return tables.buildEntityLink(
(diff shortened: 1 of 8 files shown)


templates/partials/filters/ie-report-types.html
```diff
@@ -3,11 +3,11 @@
       <legend class="label" for="report_type">Report type</legend>
       <ul>
         <li>
-          <input id="notice-checkbox-1" name="is_notice" type="checkbox" value="true">
-          <label for="notice-checkbox-1">24 and 48 hour notices</label>
+          <input id="notice-checkbox-1" name="is_notice" type="radio" value="true">
```
jmcarp: If these are radio buttons, will it be annoying to un-select them after choosing one or the other? Is there a way to clear this filter without scrolling to the top or bottom of the page to find the clear button? Also, this whole filter is commented out for now.

static/js/pages/independent-expenditures.js
```diff
@@ -46,8 +46,9 @@ var columns = [
       }
     }
   },
-  tables.dateColumn({data: 'expenditure_date', className: 'min-tablet hide-panel-tablet'}),
-  tables.urlColumn('pdf_url', {data: 'expenditure_description', className: 'all hide-panel', orderable: false}),
+  _.extend(columns.supportOpposeColumn, {className: 'min-tablet'}),
```
jmcarp: This line mutates `columns.supportOpposeColumn`--if any code uses that column later on, it'll still have the added class name. That's probably fine in this case, but it's generally good practice to use something like `_.extend({}, columns.supportOpposeColumn, {className: 'min-tablet'})` to avoid mutating the original object.

templates/partials/filters/ie-report-types.html
noahmanger: Good point. I reverted since we're not showing it anyway.

static/js/pages/independent-expenditures.js
noahmanger: Cool. Thanks.

PR #1067, opened 2016-02-03
Feature/ie improvements

noahmanger:
cc @emileighoutlaw


emileighoutlaw:
I’m so slow with GH, I had only just gotten to edit the page when I saw that you'd already made the change. Looks good.

PR #1073, opened 2016-02-04
Feature/candidate office pages

jmcarp:
Add candidate data views for president, senate, and house. This isn't complete, but should provide enough structure to start discussion with @noahmanger and @jenniferthibault.

Note: includes #1060.


noahmanger:
Next step: Review with Jen and me.

jmcarp:
@noahmanger how do you want to move forward on this? It sounds like @LindsayYoung's going to review the associated patch against the API. Do you want to review https://github.com/18F/fec-style/pull/248 and this patch? Is there more I should do on these three PRs before review?

templates/partials/candidates-office-table.html
```diff
@@ -0,0 +1,17 @@
+<table id="results" class="data-table" data-type="candidate">
```
noahmanger: What do you think about moving the `*-table.html` partials to just include them in the main page template for these pages? Both don't have that much code, and we never reference them anywhere else, so it could cut down on the number of files you need to create for each page.

templates/candidates-office.html
```diff
@@ -0,0 +1,34 @@
+{% extends 'layouts/main.html' %}
+
+{% block title %}
+  Browse candidates
+{% endblock %}
+
+{% block body %}
+
+<header class="page-header slab slab--primary">
+  <h1 class="page-header__title tst-page-title">Candidates</h1>
```
noahmanger: We're not actually using this `tst-page-title` class for testing on this page are we? Might as well remove it.

noahmanger:
I'm having trouble running this locally to check it out for functionality (my local API seems to be having some data trouble). But this seems like a good place to use our new "merge unfinished work into develop and hide with feature flags if necessary" plan.

My only question I think is how this is set up to handle senate candidates. When you choose an election year and a time period, are you seeing _only_ senate candidates who filed to run that election year and then their financials for that time period? (This is what I'd expect, but just checking).

templates/partials/candidates-office-table.html
jmcarp: No strong preference, but if that change gets made, I think it can happen in a separate patch.

templates/candidates-office.html
jmcarp: Done.

templates/partials/candidates-office-table.html
noahmanger: :+1:

templates/partials/candidates-office-filter.html
```diff
@@ -0,0 +1,27 @@
...
+{{ text.field('min_disbursements', 'Minimum disbursements', {'data-suffix': 'or more', 'data-inputmask': '"alias": "currency", "rightAlign": false, "clearMaskOnLostFocus": true' }) }}
+{{ text.field('max_disbursements', 'Maximum disbursements', {'data-suffix': 'or less', 'data-inputmask': '"alias": "currency", "rightAlign": false, "clearMaskOnLostFocus": true'}) }}
+{{ text.field('min_cash_on_hand_end_period', 'Minimum cash on hand', {'data-suffix': 'or more', 'data-inputmask': '"alias": "currency", "rightAlign": false, "clearMaskOnLostFocus": true' }) }}
+{{ text.field('max_cash_on_hand_end_period', 'Maximum cash on hand', {'data-suffix': 'or less', 'data-inputmask': '"alias": "currency", "rightAlign": false, "clearMaskOnLostFocus": true'}) }}
+{{ text.field('min_debts_owed_by_committee', 'Minimum debt', {'data-suffix': 'or more', 'data-inputmask': '"alias": "currency", "rightAlign": false, "clearMaskOnLostFocus": true' }) }}
+{{ text.field('max_debts_owed_by_committee', 'Maximum debt', {'data-suffix': 'or less', 'data-inputmask': '"alias": "currency", "rightAlign": false, "clearMaskOnLostFocus": true'}) }}
+{% include 'partials/filters/hideNull.html' %}
```
noahmanger: Don't need hide null anymore

noahmanger:
Ok, I just noticed a couple small things so I included them in a PR: https://github.com/jmcarp/openFEC-web-app/pull/26

Behavior here is looking good. I had a couple notes on the fec-style patch, but this has my :+1: once my patch is merged in.

static/templates/candidates-office.hbs
```diff
@@ -0,0 +1,35 @@
+<div class="panel__row">
+  <div class="panel__heading">
+    <h4 class="panel__title"><a href="{{basePath}}/candidate/{{candidate_id}}">{{name}}</a></h4>
```
noahmanger: This link should have the current cycle included in the URL.

noahmanger:
Spoke too soon. I'm seeing some problems that seem to be on the API side:
- Financial filters aren't working for me (seeing 422 errors)
- The name filter isn't working (no error visible) and I tried replacing the new typeahead filters but that's not working either

noahmanger:
Also downloads aren't working locally. Is that expected?

noahmanger:
Could be a separate issue (doesn't need to hold up merge), but I'll capture it here: the range filters only seem to fire after hitting enter, rather than on tabbing away, like other text input filters work.

jmcarp:
Try again with https://github.com/18F/openFEC/pull/1541. Downloads won't work locally unless you're running redis and a celery worker--I'm guessing nobody but @LindsayYoung or me has set up downloads to run locally.

LindsayYoung:
I feel like the date choices should be oldest to newest. I was super confused for a bit. 
screen shot 2016-02-25 at 1 31 53 pm

jmcarp:
@LindsayYoung: the date choices are meant to mirror the analogous widget on the candidate detail page, which iirc is what @noahmanger and @jenniferthibault suggested. I don't have a preference about this.

LindsayYoung:
Another small thing is that the currency inputs will turn commas into decimals, I can live with it just wanted to flag it incase that behavior is unwanted.

jmcarp:
That's also the current behavior of the currency input masks throughout the site, and has been since we introduced input masking last year. Again, no preference on this--if it's a problem, I'd file a separate issue.

PR #1079, opened 2016-02-05
Updates the language on the IE disclaimer to be more accurate

noahmanger:
EOM


noahmanger:
@jmcarp can you just add this language to #1080 so it all gets merged in at the same time? I'll close this now.

PR #1086, opened 2016-02-09
Get committee contributions from disbursements endpoint.

jmcarp:
Depends on https://github.com/18F/openFEC/pull/1504.

### DIFF
--- a/static/js/pages/committee-single.js
+++ b/static/js/pages/committee-single.js
@@ -43,13 +43,13 @@ var sizeColumns = [
 
 var committeeColumns = [
   {
-    data: 'contributor_name',
+    data: 'committee_name',
     className: 'all',
     orderable: false,
     render: function(data, type, row, meta) {
       return tables.buildEntityLink(
         data,
-        helpers.buildAppUrl(['committee', row.contributor_id]),
+        helpers.buildAppUrl(['committee', row.committee_id]),
         'committee'
       );
     }
@@ -59,8 +59,11 @@ var committeeColumns = [
     className: 'all',
     orderable: false,
     orderSequence: ['desc', 'asc'],
-    render: tables.buildTotalLink(['receipts'], function(data, type, row, meta) {
-      return {contributor_id: row.contributor_id};
+    render: tables.buildTotalLink(['disbursements'], function(data, type, row, meta) {
+      return {
+        committee_id: row.committee_id,
+        recipient_committee_id: row.recipient_id
+      };
     })
   }
 ];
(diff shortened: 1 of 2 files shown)


LindsayYoung:
I am still seeing the links go to receipts, instead of disbursements I think that should change, since our methodology changed.

http://localhost:3000/committee/C00575795/?tab=between-committees

jmcarp:
Agreed, but I made that change in the original patch. When I run this branch, both aggregates under "between committees" link to itemized disbursements. Did the javascript rebuild successfully when you checked this out?

Related: we may want to purge this tab of the word "receipts", since we aren't strictly looking at receipts anymore. Instead of "receipts from other committees", we may want to say "disbursements from other committees"--although users might get confused that the language on the two sub-tabs is so similar. Thoughts @LindsayYoung @emileighoutlaw?

emileighoutlaw:
We're talking about disbursements from other committees *to* committees? Can someone take a stab at explaining to me why that isn't a receipt? I'm definitely not opposed to the idea, but I'm trying to wrap my head around it.

jmcarp:
@emileighoutlaw: the reason for the change is that we aren't able to count up committee-to-committee receipts in an accurate way--FEC staff indicated a preference for using disbursements instead. So we're going to figure out how much each committee gave to each other committee, and how much each committee received from each other committee, by summing itemized disbursements.

emileighoutlaw:
Ahh, cool. Thanks for explaining, Josh. So would it be accurate to call this `Committee to committee disbursements`? I like that phrasing, if it's technically accurate

jmcarp:
Looks like there are a few titles we could consider changing in this view:

* Tab title, currently "Between committees"
* Tab header, "Receipts and disbursements between committees"
* Grouping labels, "Receipts from committees" and "Disbursements from committees"
* Table headers, "Receipts from other committees" and "Disbursements to other committees"

I think "Committee to committee disbursements" is accurate for the tab header.

emileighoutlaw:
Okay! I think I'm finally with you. How about:

Tab title: `Between committees`
Tab header: `Committee to committee disbursements`
Grouping labels: `Disbursements received from committees` and `Disbursements to committees`
Table headers: `Disbursements received from committees` and `Disbursements to committees`

LindsayYoung:
I pulled and did an npm build and I am still not getting the right links.

jmcarp:
Language updated--thanks @emileighoutlaw! Looks like the linking issue that @LindsayYoung encountered was fixed by running `gulp build-js` again--not sure what was going on, but all seems well.

emileighoutlaw:
Fantastic! Thank *you*, @jmcarp!

PR #1089, opened 2016-02-10
Miscellaneous cleanup.

jmcarp:
* Restore ordering
* Revert seek pagination
* Revert calls to `getCycle` (no effect in this context)
* Revert `is_notice` flags on endpoints without that option

cc @noahmanger

### DIFF
--- a/static/js/modules/columns.js
+++ b/static/js/modules/columns.js
@@ -1,7 +1,5 @@
 'use strict';
 
-/* global require, module */
-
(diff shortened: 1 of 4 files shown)


jmcarp:
Note: depends on https://github.com/18F/openFEC/pull/1497.

By the way, it's clearly not a big deal, but it still strikes me as weird that we're including candidate information twice in the table details panel.

PR #1092, opened 2016-02-10
Feature/new accordions

noahmanger:
1. Updates the way we implement the accordions (and therefore the glossary) to make use of the separate and updated libraries we made out of the original code
2. Introduces accordion grouping to filters:
   screen shot 2016-02-10 at 1 07 33 pm
3. Updates the glossary (note that the full term is now a button!)
   screen shot 2016-02-10 at 1 06 59 pm
4. Updates the totals tables on committee pages. The new accordion style requires slightly different markup in the form of a `<button>` followed directly by a `<div>`. Because buttons can't have nested html tags inside them, it required a little bit of redesign of these tables. Curious what @jenniferthibault thinks about this, but I don't think we should spend too much time on it (happy to explain the rationale / constraints a little more too):

screen shot 2016-02-10 at 1 03 38 pm
## Remaining tasks:
- [x] Group rest of filters into accordions
- [x] Verify design changes work for Jen

Depends on https://github.com/18F/fec-style/pull/244
Resolves https://github.com/18F/openFEC/issues/1495


jmcarp:
@noahmanger can you take a look at the merge conflict and failing builds and ping me when you want me review?

noahmanger:
Ok, conflicts are fixed. We'll see how the build fares....

noahmanger:
I don't get why this build is failing: https://travis-ci.org/18F/openFEC-web-app/jobs/109987402#L727

It says it can't find underscore, but I see it in package.json . Any ideas?

Once that's resolved, this and it's related PRs are good to go.

noahmanger:
Now that the pull requests on the libraries have been merged in, this should be good for a review.

PR #1093, opened 2016-02-12
Feature/rollback hide null

### LINKED ISSUE #1065: Figure out "sort_hide_null" frontend behavior
When we moved the "Hide null when sorting" checkbox to be a regular filter, some things changed:

Previously, it was set using table options and it was hidden from the visible URL. This meant we would set it by default, which is useful when sorting by date so that results with no date don't show up at the top. 

But now, because it's a regular filter there's no way to set it without also setting a parameter in the URL. 

Also, now that we have filter tags, it means that treating it as a regular filter generates a tag for it, which is kind of weird:

screen shot 2016-02-02 at 3 57 27 pm

I'd like to figure out what we want the behavior to be:
- Do we want it to be checked by default?
- Do we want it visible in the URL?
- Do we want it to generate a tag?

Would love your thoughts on this @jenniferthibault .

jmcarp:
Always hide nulls on sorted column.

[Resolves #1065]

### DIFF
--- a/static/js/modules/tables.js
+++ b/static/js/modules/tables.js
@@ -556,7 +556,7 @@ DataTable.prototype.isPending = function() {
 };
 
 DataTable.prototype.buildUrl = function(data, paginate) {
-  var query = _.extend({}, this.filters || {});
+  var query = _.extend({sort_hide_null: true}, this.filters || {});
   paginate = typeof paginate === 'undefined' ? true : paginate;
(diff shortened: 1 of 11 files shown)


noahmanger:
These tests seem to just be Travis failing. Or are they legit?

jmcarp:
Ack, this is an actual failing test: https://travis-ci.org/18F/openFEC-web-app/jobs/108898841#L971-L985. Pushed up a fix.

PR #1094, opened 2016-02-13
Remove $'s from commands in readme for pastability

evankroske:
Now, if you copy a command from the readme with triple-click control-C,
you can paste it directly into a shell and run it.

Command:

    python - README.md <<EOF
    import re
    import sys

    fn = sys.argv[1]
    lines = None
    with open(fn) as f:
      lines = f.readlines()
    with sys.stdout as f:
      for line in lines:
        f.write(re.sub(r"^( +)\\$ ", r"\1", line))
    EOF


noahmanger:
Thanks!

PR #1095, opened 2016-02-13
Fixes run server command in the readme

evankroske:
In the readme, changes the command to run the server to use the prod
instance of the FEC API and a demo API key. Before this commit, the
server would try to use a local instance of the FEC API.


noahmanger:
My only question on this is do we want to point to production API by default? Or dev?

evankroske:
I think dev would be better because it's more current. This is the address, right?

http://fec-dev-api.18f.gov

LindsayYoung:
Thanks again for the feedback Evan. 

I think we might want both sets of instructions. Myself and others are frequently looking at API and web changes at the same time, so running locally is important to know. I can see that it would be less useful for many people so I think we should explain the options. 

And yes, that is the correct dev address.

noahmanger:
Cool. So yeah, let's just have both sets of instructions. Do you want to make that change @evankroske ?

evankroske:
Sure, I'll give it a shot.

evankroske:
Please take another look.

PR #1097, opened 2016-02-14
Use top-level committee information.

jmcarp:
Necessary to handle an update in the electioneering communication schema
such that nested committee information is removed and top-level
committee details added.

Review with https://github.com/18F/openFEC/pull/1518.


noahmanger:
I checked this out and it works with that API PR, so good to merge whenever that's merged.

Issue #1024, opened 2016-01-11
Quick accessibility fixes for downloads


noahmanger:
Via Nick:

- [x] When you click the export this data button, move focus to the download button. When using the keyboard, you have to tab through the rest of the page to get to the download button, which is a pain.
- [x] Second, for the download button, associate the file name with the button, `aria- describedby="id_of_file_name"`

PR #1105, opened 2016-02-24
Feature/download keyboard

noahmanger:
Resolves https://github.com/18F/openFEC-web-app/issues/1024

### DIFF
--- a/static/js/modules/tables.js
+++ b/static/js/modules/tables.js
@@ -589,8 +589,9 @@ DataTable.prototype.fetch = function(data, callback) {
 
 DataTable.prototype.export = function() {
   var url = this.buildUrl(this.api.ajax.params(), false);
-  download.download(url);
+  var item = download.download(url);
   this.disableExport({message: DOWNLOAD_MESSAGES.pending});
+  item.$body.find('.button').focus();
 };
 
 DataTable.prototype.isPending = function() {
--- a/static/js/modules/tables.js
+++ b/static/js/modules/tables.js
@@ -589,9 +589,8 @@ DataTable.prototype.fetch = function(data, callback) {
 
 DataTable.prototype.export = function() {
   var url = this.buildUrl(this.api.ajax.params(), false);
-  var item = download.download(url);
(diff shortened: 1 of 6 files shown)


static/js/modules/tables.js
```diff
@@ -589,8 +589,9 @@ DataTable.prototype.fetch = function(data, callback) {
 
 DataTable.prototype.export = function() {
   var url = this.buildUrl(this.api.ajax.params(), false);
-  download.download(url);
+  var item = download.download(url);
   this.disableExport({message: DOWNLOAD_MESSAGES.pending});
+  item.$body.find('.button').focus();
```
jmcarp: If possible, prefer not to have one component (the table) reach down into a different component (the download) to call methods on its children--this requires a developer working on the table component to know a lot about the internals of the download component. What if `DownloadItem.init` handled the focus instead?

noahmanger: Well the problem is that we only want to move focus if the user has initiated a download by clicking export. If we included it in `DownloadItem.init` it would be set even when stored downloads are hydrated when loading a page, which would be disruptive. Or is there a better way to do this?

noahmanger: @jmcarp do you have thoughts on this?

jmcarp: I think it would be less confusing to (1) expose a method on `DownloadItem` that focuses on the button, or (2) add a parameter to the `download` helper indicating whether the button should be focused after draw. Either approach would hide unimportant details of the `DownloadItem` implementation from devs who only want to work on tables--like the fact that `DownloadItem` has an attribute called `$body`, or that the relevant selector to focus on is called `.button`.

noahmanger: Makes sense. I'll take another look.

jmcarp:
When I checked this out locally, I noticed that exporting a table added two or more download items to the download panel. I fixed that issue and added some minor style cleanup in cc1edb49ec44371653731685a9efab8c12381ee5.

PR #1106, opened 2016-02-24
Drop Selenium tests.

jmcarp:
Our current Selenium tests are expensive to maintain but essentially
never catch real regressions. This patch removes all Selenium tests and
associated documentation; if we opt to restore end-to-end testing in the
future, best to start fresh.


.travis.yml
```diff
@@ -64,40 +35,13 @@ before_script:
...
-  - |
-      if [[ $BUILD_LEADER = 'YES' && $TRAVIS_PULL_REQUEST = 'false' ]]; then
-        if [[ $BUILD_AGGREGATE_STATUS = 'others_succeeded' ]]; then
-          echo "All builds passed; deploying..."
-          invoke deploy --branch $TRAVIS_BRANCH --yes
-        else
-          echo "One or more builds failed; skipping deploy"
-        fi
-      fi
-  - if [[ $BUILD_LEADER = 'YES' ]]; then codecov; fi
```
ccostino: There is a bunch of stuff in this Travis config file being removed that at a glance doesn't appear to be Selenium related - Postgres information and setup, Python setup, and some things related to Travis' config itself  Were these all here just to get Selenium setup and/or support it?

README.md
```diff
@@ -49,22 +48,12 @@ Install Python dependencies:
...
+    $ npm install
 
 ### Configuration
 
 The Flask app talks to an API for data. See [openFEC](http://github.com/18F/openFEC).
 
-The app has HTTP auth enabled. You will need to set environment variables with your desired username and password.
-Those vars are `FEC_WEB_USERNAME` and `FEC_WEB_PASSWORD`. There are other config environment variables that you
-can set, but that have defaults. You can see those in `openfecwebapp/config.py`.
-
```
ccostino: The basic auth is being removed as a part of this PR, thus negating the need for these environment variables?

README.md
```diff
@@ -91,10 +80,6 @@ To use styles served from a custom location (e.g., if developing against a local
 To be able to have links between this app and a local installation of the cms:
 
     export FEC_CMS_URL=http://localhost:8000
-    
-If you'd like the app to cache API requests it makes, use the `--cached` flag:
-
-    python __init__.py --cached
```
ccostino: This is also no longer supported?  Or did it not work in the first place?

ccostino:
The removal of the Selenium-related pieces and documentation looks fine to me, but I had several questions around the other seemingly un-related changes made.  Were these things all in place to support Selenium or are they cleanup items that have just been rolled into this PR?

README.md
jmcarp: This option hasn't existed for a long time--this is just doc rot.

.travis.yml
jmcarp: Yes, these were needed when we were using selenium and running multiple builds in parallel. We had to install postgres and run the api locally for the selenium tests, and we also had to wait for all parallel builds to pass before deploying. Both of those issues go away with the selenium tests.

README.md
jmcarp: Exactly!

.travis.yml
ccostino: Oh wow, so lots of extra setup and legwork just to support these that we no longer have to worry about with this change. :+1:

README.md
ccostino: Gotcha - thanks for clarifying!

ccostino:
Awesome, thanks @jmcarp!  It's always nice to get a bit of cleanup in place and remove cruft. :-)

PR #1109, opened 2016-02-26
Use election year filter for candidates.

### LINKED ISSUE #1002: Discuss behavior of "election years" filter on candidates view
The cycle filter on the browse candidates view is currently labeled "election years", but actually filters on two-year cycle periods instead of election years. These fields are different--for example, Presidential candidate Hillary Clinton has election years `[2008, 2016]` and cycles `[2008, 2010, 2012, 2014, 2016]`. In other words, filtering using the "2011-2012" filter *will* retrieve Hillary Clinton, even though she wasn't a candidate for any office in 2011-2012. As of #1000, the "Election year" column in the candidates table will actually display election years, rather than two-year periods.

I was discussing this with @noahmanger and wasn't sure exactly what the expect behavior of the "election years" filter is. Curious what others think--@PaulClark2 @jwchumley @LindsayYoung. @noahmanger, please chime in if the description above isn't clear.

### LINKED ISSUE #1078: On candidate page, election year filter showing candidates from other years
In candidate filters, when 2015-2016 filter is applied, results still show candidates from other years.
@emileighoutlaw I think you got a screenshot?

jmcarp:
[Resolves #1002]
[Resolves #1078]

Requires https://github.com/18F/openFEC/pull/1544.

Note: will need to make similar changes to the CMS after merging.

### DIFF
--- a/templates/partials/filters/election-years.html
+++ b/templates/partials/filters/election-years.html
@@ -1,3 +1,3 @@
 {% import 'macros/filters/years.html' as years %}
 
-{{ years.years('cycle', 'Election years') }}
+{{ years.years('election_year', 'Election years') }}
--- a/templates/partials/navigation.html
+++ b/templates/partials/navigation.html
@@ -10,7 +10,7 @@
         </li>
         <li class="t-bold site-nav__subtitle">Browse:</li>
         <li class="site-nav__item">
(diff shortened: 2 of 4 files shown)


LindsayYoung:
This looks good to me. Dev is down now so I am going to do local testing later. @ccostino feel free to grab this one if you can get to it first, if not I will get to it as soon as I can. The underlying API work is already merged.

ccostino:
Awesome, I'll take a look @LindsayYoung!

PR #1110, opened 2016-02-28
Feature/externalize feedback

jmcarp:
Allow users to file issues via personal github accounts:

screen shot 2016-02-28 at 2 58 01 pm
- Replace feedback widget with external library
- Run feedback instance with cloud.gov shim

TBD:
- Should feedback instance be run in each cloud.gov space?
- Should server shim also be externalized?


jmcarp:
@noahmanger: I see that this has accumulated some merge conflicts. Let me know when you're ready to review, and I'll resolve them.

noahmanger:
Ok, I took a look at this. I'd like @ccostino to take a closer look before we merge, but that should probably happen latter as this isn't a priority at the moment. 

Only question I have is why wouldn't we want to externalize the server-shim too? Seems a little strange to have a separate package.json and manifest.yml in this repo. And if it were externalized, would it mean we would need just a single instance that the CMS and web app could access?

jmcarp:
Yes, there's only going to be a single feedback instance that's shared by the CMS and webapp. The shim doesn't have to go in this repo, but it also doesn't belong in https://github.com/jmcarp/feedback, since it's specific to our particular cloud.gov configuration.

Reassigning to @ccostino.

PR #1125, opened 2016-03-11
Fix cycle labels and query parameters.

jmcarp:
Filters for candidates, committees, filings, and independent
expenditures currently use the same partial for cycles. This leads to
unexpected behavior, since the different resources expect different
query parameters: candidates are filtered on election year, and other
resources are filtered on two-year period. This patch fixes filter
labels and query parameters, and drops an unnecessary partial.


noahmanger:
LGTM. Can you fix the merge conflicts and then I can merge?

jmcarp:
Merged into `release/public-beta-20160316`.

PR #1115, opened 2016-03-03
Adjusting labels on typeaheads

noahmanger:

PR #1128, opened 2016-03-11
updating typeahead labels

noahmanger:
Clean version of updated typeahead labels to replace #1115

### DIFF
--- a/openfecwebapp/templates/partials/candidates-filter.html
+++ b/openfecwebapp/templates/partials/candidates-filter.html
@@ -6,7 +6,7 @@
 
 {% block filters %}
 <div class="filters__inner">
-  {{ typeahead.field('q', 'Candidate name or ID', 'Find an ID by searching for a candidate name', 'candidates', True) }}
+  {{ typeahead.field('q', 'Candidate name or ID', '', 'candidates', True) }}
   {% include 'partials/filters/election-years.html' %}
   {% include 'partials/filters/office-sought.html' %}
   {% include 'partials/filters/parties.html' %}
--- a/openfecwebapp/templates/partials/candidates-filter.html
+++ b/openfecwebapp/templates/partials/candidates-filter.html
@@ -6,7 +6,7 @@
 
 {% block filters %}
 <div class="filters__inner">
-  {{ typeahead.field('q', 'Candidate name or ID', '', 'candidates', True) }}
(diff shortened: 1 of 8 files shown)


jmcarp:
A few questions about the labeling pattern. It looks like we're using text like "name or ID" when we require users to pick an option from the typeahead suggestions, and just "name" when we also allow users to enter free text. First of all, if that's the expected pattern, I don't think we're following it consistently--see `candidates-filter.html`. Second, searching in a typeahead by candidate or committee ID should always find the relevant candidate or committee, so I'm not sure I understand the "name or ID" vs "name" labels.

Also, I may have made this mistake in the first place, but since the `typeahead.field` macro takes many arguments, let's always explicitly pass argument names for optional arguments: `dataset='candidates', allow_text=True`.

noahmanger:
Good question. @emileighoutlaw should chime in here, but here's my take:

The "Name or ID" label isn't meant to denote that free text entry isn't allowed. Rather, it's meant to be explicit that you can search by ID. We've received feedback that users want to filter by [candidate ID](https://github.com/18F/FEC/issues/230) and [committee ID](https://github.com/18F/FEC/issues/254), so wanted to call these out specifically. 

And then the reason that "Contributor name" / "(Disbursement) Recipient name" is just "Name" not "Name or ID" is because there won't necessarily be an ID for those entities. 

I get that this is potentially confusing though, and imagine we're going to need to continue to refine.

I'll update the naming of args.

jmcarp:
So because typeahead search always allows search by name *or* ID, should all typeahead filters be labeled to reflect that?

noahmanger:
I just worry that may be confusing when you have a field that searches across records that may or may not have an ID associated with them. But maybe it's more confusing this way.

jmcarp:
I'm going to vote in favor of not caring too much, if only because I don't think users actually read this text. So we can merge this as-is. But for the sake of argument, I don't think "or ID" implies that there must be an ID--I read it as correctly stating that you can search for the candidate or committee using the typeahead by name or ID.

noahmanger:
Gotcha. Yeah ok. The more I think about it the more I see that point. I'm down to make them all "Name or ID" but just want @emileighoutlaw 's blessing :)

emileighoutlaw:
I think calling them all "Name or ID" sounds good! Thanks for bringing me in on this one, y'all. :)

jmcarp:
Merged into `release/public-beta-20160316`.

PR #1132, opened 2016-03-14
Restore missing globals.

jmcarp:
h/t @noahmanger 

Fixes candidate and committee detail pages.


noahmanger:
Looks like we're missing `macros/tables.html`. I'm seeing that error on several committee pages (e.g. http://localhost:3000/committee/C00213512/ )

PR #1137, opened 2016-03-17
Removing duplicate row in the H + S totals table

noahmanger:
Noticed there was a duplicate on House and Senate totals tables:

screen shot 2016-03-17 at 1 02 49 pm

Can be merged after release.

### DIFF
--- a/openfecwebapp/templates/partials/committee-totals-house-senate.html
+++ b/openfecwebapp/templates/partials/committee-totals-house-senate.html
@@ -30,7 +30,6 @@
      (totals.0.refunded_individual_contributions, 'Individual refunds'),
      (totals.0.refunded_political_party_committee_contributions, 'Political party refunds'),
      (totals.0.refunded_other_political_committee_contributions, 'Other committee refunds'),
-     (totals.0.other_disbursements, 'Other disbursements'),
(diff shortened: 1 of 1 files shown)


openfecwebapp/templates/partials/committee-totals-house-senate.html
```diff
@@ -30,7 +30,6 @@
      (totals.0.refunded_individual_contributions, 'Individual refunds'),
      (totals.0.refunded_political_party_committee_contributions, 'Political party refunds'),
      (totals.0.refunded_other_political_committee_contributions, 'Other committee refunds'),
-     (totals.0.other_disbursements, 'Other disbursements'),
```
LindsayYoung: Looking at the current committee summaries and I think we want to keep the first `other disbursements` and switch out the duplicative `other disbursements` at the end with `total disbursements`.

http://www.fec.gov/fecviewer/CandidateCommitteeDetail.do (look up a house or senate member)

noahmanger: So, we show "total disbursements" above this table in the header. If we do want it here, we should also add "total receipts" too.

LindsayYoung: that makes sense!

PR #1145, opened 2016-03-29
Use global hbsfy transform in tests.

jmcarp:
By default, browserify only applies the requested transforms to source files, not to dependencies. This patch uses the global flag to apply the handlebars transform to dependencies too.

cc @noahmanger


noahmanger:
Thank you!

PR #1152, opened 2016-04-01
Search landing for Legal Resources

adborden:
Here's the first pass implementation for the [legal resources search ui](https://trello.com/c/g0MX7L9w/62-hook-ui-up-to-search-api). The goal here is to provide a basic frontend UI for us to validate some assumptions regarding the search strategy in general rather than a pixel perfect UI. I've gone off of the existing styles for now but based the layout off of the direction from @edmullen's [invision app](https://projects.invisionapp.com/share/5N6NL76QM#/screens/145835130).

## Stub data

We've stubbed the data for this initial PR while @anthonygarvan is working on the backend, forward. Soon we'll meet in the middle with a frontend that hits search via the openFEC API.

## Feature flag

We're using the same feature flag from https://github.com/18F/fec-cms/pull/260 to hid the Legal resources nav. It can be enabled by setting the `FEC_FEATURE_LEGAL` environment variable which will be set in our staging environment.

## Future routes

The current thinking is that this is an overall search (via `/legal/search`) and then the user will be able to focus the search on specific document types (`/legal/*`). I think that's the pattern I see with the campaign finance data search. Let me know if there's another pattern. These additional routes will come later as needed.

## Relationship to the existing search

I wanted to keep this work flexible-ish while we're collecting feedback. This UI felt different enough that I decided to use completely separate templates and avoid the search macro for now. I'm open to other ideas, but I feel like we'll know more about how the two are related as we get some feedback.

If the two remain split, renaming the existing search to "data search" might be more appropriate. IDK.

screenshot from 2016-03-31 18-05-35

### DIFF
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -45,6 +45,18 @@ def load_search_results(query, query_type='candidates'):
     return results['results'] if len(results) else []
 
 
+def load_legal_search_results(query, query_type='all'):
+    filters = {}
+
+    if query:
+        filters['q'] = query
+
+    url = '/legal/search'
+    results = _call_api(url, **filters)
+
+    return results['results'] if len(results) else []
+
+
 def load_single_type(data_type, c_id, *path, **filters):
     data = _call_api(data_type, c_id, *path, **filters)
     return result_or_404(data)
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -45,16 +45,38 @@ def load_search_results(query, query_type='candidates'):
     return results['results'] if len(results) else []
 
 
-def load_legal_search_results(query, query_type='all'):
+def load_legal_search_results(query, query_type='all', limit=20):
     filters = {}
+
+    def _transform_advisory_opinion(advisory_opinion):
+        source = advisory_opinion['_source']
+        return {
+            'id': source['AO_Id'],
+            'no' : source['AO_No'],
+            'name': source['AO_name'],
+            'summary': source['AO_Summary'],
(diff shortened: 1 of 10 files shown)


openfecwebapp/api_caller.py
```diff
@@ -45,6 +45,40 @@ def load_search_results(query, query_type='candidates'):
     return results['results'] if len(results) else []
 
 
+def load_legal_search_results(query, query_type='all', limit=20):
```
jmcarp: The `query_type` parameter doesn't seem to be used here. Could we drop it?

openfecwebapp/api_caller.py
```diff
@@ -45,6 +45,40 @@ def load_search_results(query, query_type='candidates'):
...
+    filters = {}
+
+    def _transform_advisory_opinion(advisory_opinion):
```
jmcarp: These nested functions don't seem to rely on variables defined in the outer context, so do they need to be nested? Prefer shorter, flatter functions, all things being equal: "flat is better than nested".

openfecwebapp/api_caller.py
```diff
@@ -45,6 +45,40 @@ def load_search_results(query, query_type='candidates'):
...
+            'name': source['AO_name'],
+            'summary': source['AO_Summary'],
+            'tags': source['AO_tags'],
+            'description': source['description'],
+            'doc_id': source['doc_id'],
+            'text': source['text'],
+        }
+
+
+    def _transform(data):
```
jmcarp: Just curious--what about applying some or all of these transforms on the api side? Should the api serve idiosyncratically named elastic-style fields like `_type` and `_source`?

Issue #1052, opened 2016-01-25
Bug: Open in new tab doesn't work on authorized committee pages


emileighoutlaw:
This bug is visible from committee pages.

If you try to open one of the other tabs in a new window (with `control + click`), it doesn't work. It opens the same tab in the new window. 

![tabs](https://cloud.githubusercontent.com/assets/10552702/12557876/db72a566-c35a-11e5-913b-f5e9464bd02f.gif)

evankroske:
I'm working on this. The issue is caused by a call to `preventDefault` from the event listeners on the tabs. The plan:
- [ ] Move tablist initialization from top-level scope to exported function.
- [ ] Write tests for tablist.
- [ ] Change tablist to use anchor links and popstate instead of pushState.

noahmanger:
Thanks, @evankroske ! Sounds like a great place to plug in. This is really old code, so it will be great to get it updated and tested.

PR #1156, opened 2016-04-04
Adds tests for tablist

evankroske:
1. Configures Browserify to process tablist.js.
2. Adds basic tests for tablist.

Part of step two of #1052.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -27,13 +27,15 @@ module.exports = function(config) {
 
     files: [
       'tests/unit/**/*.js',
-      'static/js/modules/**/*.js'
+      'static/js/modules/**/*.js',
+      'static/js/vendor/**/*.js'
     ],
 
     preprocessors: {
       'tests/unit/**/*.js': ['browserify'],
       'static/js/pages/**/*.js': ['browserify'],
-      'static/js/modules/**/*.js': ['browserify']
+      'static/js/modules/**/*.js': ['browserify'],
+      'static/js/vendor/**/*.js': ['browserify']
     },
 
     browserify: browserify,
--- a/tests/unit/vendor/tablist.js
+++ b/tests/unit/vendor/tablist.js
@@ -1,0 +1,26 @@
+'use strict';
+
+var chai = require('chai');
+var sinon = require('sinon');
+var sinonChai = require('sinon-chai');
+var expect = chai.expect;
+chai.use(sinonChai);
+
+var tablist = require('../../../static/js/vendor/tablist');
+
+describe('show', function() {
+  var fixture;
+  before(function() {
+    fixture = $('<div id="fixtures"></div>');
+    $('body').append(this.fixture);
+  });
+
+  beforeEach(function() {
+    fixture.empty().append(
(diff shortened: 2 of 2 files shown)


tests/unit/vendor/tablist.js
```diff
@@ -0,0 +1,58 @@
+'use strict';
+
+var chai = require('chai');
+var sinon = require('sinon');
```
jmcarp: That's true. We also don't need `sinonChai` here so far.

karma.conf.js
```diff
@@ -27,13 +27,15 @@ module.exports = function(config) {
 
     files: [
       'tests/unit/**/*.js',
-      'static/js/modules/**/*.js'
+      'static/js/modules/**/*.js',
+      'static/js/vendor/**/*.js'
```
jmcarp: @noahmanger knows more about this than I do, but I'm wondering if `tablist` has diverged enough from wherever we got it that we can take it out of `vendor`.

tests/unit/vendor/tablist.js
```diff
@@ -0,0 +1,58 @@
...
+
+    it('should show first tab if there\'s no query', function() {
+      fixture.empty().append(
+        '<div class="tab-interface">' + 
+          '<ul role="tablist" data-name="tab">' +
+            '<li><a role="tab" data-name="tab0" href="#section-0">0</a></li>' +
+            '<li><a role="tab" data-name="tab1" href="#section-1">1</a></li>' +
+          '</ul>' +
+          '<section id="section-0" role="tabpanel" aria-hidden="true">' +
+            '' +
```
jmcarp: What do these empty strings do?

tests/unit/vendor/tablist.js
evankroske: Done

karma.conf.js
evankroske: I'm not a lawyer, but I think that unless someone rewrites tablist.js, it will remain a derivative work of the original file.

tests/unit/vendor/tablist.js
evankroske: Nothing. Removed.

karma.conf.js
jmcarp: What I'm suggesting is that it's a ship of Theseus type situation. I don't remember replacing the module all at once, but almost none of the original code is left: https://github.com/18F/openFEC-web-app/blob/14d0daa91f877fd789f6936a0e2b1e08825ec8da/static/js/vendor/tablist.js.

\\_(ツ)_/¯

tests/unit/vendor/tablist.js
```diff
@@ -0,0 +1,58 @@
+'use strict';
```
evankroske: Disabled warning because Browserify wraps modules in function expressions.

noahmanger:
Sorry for the long delay @evankroske . This looks good to me. Re: the question of if it stays as in `/vendor/`, I don't care, but we can address that separately. Thanks for adding this!

PR #1140, opened 2016-03-23
Feature/filter data update

noahmanger:
Various changes to stay in sync with https://github.com/18F/fec-cms/pull/256 and https://github.com/18F/fec-style/pull/277
- Re-arranges filters and tags so that the tag bar is on top of the filter panel
- Creates a new datatable.html template and routes all datatable URLs to it, simplifying the steps to make a new datatable page

(Note: I pulled in the release to this branch, but not sure if we want to merge that in separately before this).

PR #1161, opened 2016-04-05
Feature/filter data update

noahmanger:
Clean try of https://github.com/18F/openFEC-web-app/pull/1140


openfecwebapp/__init__.py
```diff
@@ -100,6 +100,9 @@ def get_election_url(candidate, cycle, district=None):
 def asset_for(path):
     return url_for('static', filename=assets[path].lstrip('/'))
 
+def datatable_script(slug):
```
jmcarp: This doesn't seem to include logic that's specific to data tables. How about calling it something like `asset_for_page`?

noahmanger: 👍

ccostino:
We're set now, thanks @noahmanger!

PR #1164, opened 2016-04-05
Feature/styleguide adjustments

noahmanger:
Small class name adjustments to go with cleanup done in https://github.com/18F/fec-style/pull/287

Fixes https://github.com/18F/fec-style/issues/299


jenniferthibault:
A few things: 

Not seeing the styles come through on the candidate profile pages:

screen shot 2016-04-08 at 6 26 23 pm

Annd should switch to regular style here for 404 errors:

screen shot 2016-04-08 at 6 27 25 pm

Also not seeing the style come through for the feedback button
screen shot 2016-04-08 at 6 29 02 pm

jenniferthibault:
Whoops, found one more (that was harder to find!)
I seem to have lost the download buttons, which should be cta-alt like the calendar:

screen shot 2016-04-08 at 6 55 39 pm

noahmanger:
Got it! The matter of the styles not coming through in those places is likely due to needing to run `npm run build`.

noahmanger:
Here's how download buttons look for me. Is this right? 
image

jenniferthibault:
I was originally thinking outlined (`alt` style) but I think you're right. Full CTA style because this is the thing everyone wants 👍

PR #1166, opened 2016-04-06
Adding basics for individual contributions pages

### LINKED ISSUE #1147: Add Individual Contribution datatable page
Using the filters and columns currently available, make a new datatable page for individual contributions.

noahmanger:
I want to do a little refactoring to clean things up, but this gets it started.

Resolves #1147


noahmanger:
This is almost ready, and might be worth a preliminary review from @ccostino or @LindsayYoung .

In [this commit](https://github.com/18F/openFEC-web-app/pull/1166/commits/3ada60104deb04fef4dedfb321722b7cced73194) I deleted the old table partials which were used to set the column headers for each datatable page in favor of defining the headers in `constants.py` and programmatically generating them. Does that make sense as an approach? Is `constants.py` the right place?

openfecwebapp/constants.py
```diff
@@ -234,3 +234,17 @@
     ('contributions', 'Contributions'),
     ('other', 'Other'),
 ])
+
+table_columns = OrderedDict([
+    ('candidates', ['Name', 'Office', 'Election years', 'Party', 'State', 'District']),
+    ('candidates-office-president', ['Name', 'Party', 'Receipts', 'Disbursements']),
+    ('candidates-office-senate', ['Name', 'Party', 'Receipts', 'Disbursements']),
```
ccostino: Is `State` omitted from here accidentally or intentionally?

ccostino:
@noahmanger:  This seems sensible to me in looking it over, and given how similar things are defined in `constants.py` it looks like the appropriate place for those definitions as well.

openfecwebapp/constants.py
noahmanger: Good catch! It should totally be there for senate.

noahmanger:
I removed the route to operating expenditures so we can get this merged in before that API work is done.

Also, I filed a related PR into this branch: #1174  We could either merge it into here first, or merge this and then merge that into develop.

noahmanger:
Which is to say this is ready for review. Assigning to @ccostino .

noahmanger:
Ping on this @ccostino . Not part of the release, but would be good to get this (and #1174 ) in.

openfecwebapp/templates/partials/individual-contributions-filter.html
```diff
@@ -0,0 +1,60 @@
...
+  <div class="accordion__content">
+    {{ text.field('min_amount', 'Minimum contribution', {'data-suffix': 'or more', 'data-inputmask': '"alias": "currency", "rightAlign": false, "clearMaskOnLostFocus": true' }) }}
+    {{ text.field('max_amount', 'Maximum contribution', {'data-suffix': 'or less', 'data-inputmask': '"alias": "currency", "rightAlign": false, "clearMaskOnLostFocus": true'}) }}
+    {{ date.field('date', 'Receipt date', dates ) }}
+    <div class="message message--info message--small">
+      <span class="t-block">Receipts are reported periodically, according to the filer's reporting schedule. Receipts are updated as they’re processed— that time can vary.</span>
+    </div>
+  </div>
+</div>
+{% endblock %}
```
ccostino: @noahmanger:  Which page is this template used on?

openfecwebapp/templates/partials/operating-expenditures-filter.html
```diff
@@ -0,0 +1,42 @@
...
+    {% include 'partials/filters/disbursement-purpose.html' %}
+    {{ text.field('min_amount', 'Minimum expenditure', {'data-suffix': 'or more'}) }}
+    {{ text.field('max_amount', 'Maximum expenditure', {'data-suffix': 'or less'}) }}
+    {{ date.field('date', 'Disbursement date', dates ) }}
+    <div class="message message--info message--small">
+      <span class="t-block">Disbursements are reported periodically, according to the filer's reporting schedule. Disbursements are updated as they’re processed— that time can vary.</span>
+    </div>
+  </div>
+</div>
+{% endblock %}
```
ccostino: @noahmanger:  Same question as above - or are these in anticipation of future pages not currently available?

static/js/modules/columns.js
```diff
@@ -68,34 +236,111 @@ var filings = {
...
-    width: '20px',
+  receipt_date: dateColumn({data: 'receipt_date', className: 'min-tablet'}),
+  coverage_end_date: dateColumn({data: 'coverage_end_date', className: 'min-tablet hide-panel', orderable: false}),
+  total_receipts: currencyColumn({data: 'total_receipts', className: 'min-tablet hide-panel'}),
+  total_disbursements: currencyColumn({data: 'total_disbursements', className: 'min-tablet hide-panel'}),
+  total_independent_expenditures: currencyColumn({data: 'total_independent_expenditures', className: 'min-tablet hide-panel'}),
+  modal_trigger: modalTriggerColumn
+};
+
+var independentExpenditures = [
```
ccostino: This is just an array of objects representing all of the different columns for the independent expenditures, correct?

static/js/modules/columns.js
```diff
@@ -68,34 +236,111 @@ var filings = {
...
+          helpers.buildAppUrl(['committee', data.committee_id]),
+          'committee'
+        );
+      } else {
+        return '';
+      }
+    }
+  },
+  modalTriggerColumn
+];
```
ccostino: Thinking more about these arrays,  I wonder if it's worth trying to encapsulate the construction of the remaining columns in them with similar helper methods as the `dateColumn`, `currencyColumn`, etc. (or instantiating them in a similar fashion like lines 11 - 13 above) just to help with comprehension.  Do you think that's a worthwhile exercise in this case to help with code clarity?

static/js/pages/individual-contributions.js
```diff
@@ -0,0 +1,28 @@
...
+
+var $ = require('jquery');
+
+var tables = require('../modules/tables');
+var helpers = require('../modules/helpers');
+var columns = require('../modules/columns');
+
+var donationTemplate = require('../../templates/receipts.hbs');
+
+$(document).ready(function() {
```
ccostino: How many other `$(document).ready()` calls are made in the app, and are they (potentially) made multiple times on the same page?  If so, that can become problematic because there's no guarantee of the order in which they're run.  Oftentimes it's better to create objects that contain the functionality and provide an `init` method on them that takes care of what is being done in the `$(document).ready()` function(s).  With these `init` methods you can have one `$(document).ready()` to manage them in a central place that you are then in control of.

I searched through the code base and it doesn't appear like it's an issue if I understand the structure correctly, where most of these `$(document).ready()` functions are intended for a single page and wouldn't conflict with one another, but I did note one in `init.js` and `modules/toggle.js` as well so I wanted to call it out just in case.

We can review this in more detail if you'd like, I'm happy to talk about the concern if it's not clear and/or help investigate further to see if it is something we need to consider.

ccostino:
@noahmanger: I only had a couple of additional questions/comments after going through the rest of this more closely (and I think my question about a couple of the templates is related to the API endpoint that's not available yet).  Overall I like the organization of the code and the separation of things into more isolated functions within their own objects/modules, that will help with code clarity and further reuse. :-)

openfecwebapp/templates/partials/individual-contributions-filter.html
noahmanger: This is used by `receipts/individual-contributions`, which is the only new page this patch adds.

openfecwebapp/templates/partials/operating-expenditures-filter.html
noahmanger: Ah yes, this would be used by `/disbursements/operating-expenditures` but I removed that from the routes. Should we remove the template for now? Or keep it for when we add the route?

static/js/modules/columns.js
noahmanger: Yep, I agree. I started to do this, but then realized that many of the `committee` and `candidate` columns vary just _slightly_ across the different pages, due to different ways the API returns results.

I still think it's worthwhile, but decided to keep these as-is in the interest of time (and not accidentally breaking something).

static/js/modules/columns.js
noahmanger: Yep!

static/js/pages/individual-contributions.js
noahmanger: Ah, that makes sense. I'm all for changing it, but can we do it in a subsequent PR, since this branch didn't introduce anything new in this regard?

noahmanger:
Thanks for the review! I agree with your comments in places where things could be further refined or abstracted. I see this as the first of several steps towards breaking out the code into more reusable chunks.

static/js/pages/individual-contributions.js
ccostino: 👍 Totally cool with that!

openfecwebapp/templates/partials/operating-expenditures-filter.html
ccostino: It's fine to keep it since we are actively building this out, I just wasn't sure what was using it initially but remember now. :-)

ccostino:
Okay awesome, in that case I think we're good to go.  Thanks, @noahmanger!

Issue #1139, opened 2016-03-21
Use autopilot for zero-downtime deploys


ccostino:
We recently merged an updated to the [openFEC API](https://github.com/18F/openFEC/pull/1516) that uses the `autopilot` CloudFoundry plugin for zero-downtime deploys.  We'd like to incorporate the same functionality here in the openFEC web app.

h/t to @jmcarp for all of the initial work on the API side!

PR #1167, opened 2016-04-06
Improved web-app deployment with autopilot

ccostino:
This changeset mirrors the changes recently made to the openFEC API that improve the deployment process by using the CF Autopilot plugin.

This eliminates the need for the custom blue/green deploys and managing two apps within our Cloud Foundry spaces.  It also eliminates the problems we've been having with our `node_modules` being cached.

h/t to @jmcarp for the API implementation in https://github.com/18F/openFEC/pull/1516 and answering questions related to the setup!

Closes https://github.com/18F/openFEC-web-app/issues/1139

/cc @LindsayYoung and @jmcarp (if you have a moment again - thanks for your initial help and reviews!)


README.md
```diff
@@ -100,7 +100,47 @@ Compile Sass as changes are made:
...
+```
+cf target -s dev
+cf cups fec-creds-dev -p '{"SQLA_CONN": "..."}'
+```
+
+To stand up a user-provided credential service that supports both the API and the webapp, ensure that
+the following keys are set:
+
+* SQLA_CONN
+* FEC_WEB_USERNAME
```
jmcarp: The username and password variables aren't used anymore. This list of credentials is also different from the lists in the manifest files. Let's drop the comments from the manifest files so we don't have to maintain many sets of credentials and add the missing values here.

Beyond the scope of this patch, but it might be worthwhile to have the app refuse to start if it's missing credentials it needs to run--see https://github.com/18F/openFEC/blob/81aa0d02a1f7248b18df6fff80729a375b15653a/webservices/config.py#L52-L61 for a simple implementation on the API.

README.md
```diff
@@ -100,7 +100,47 @@ Compile Sass as changes are made:
...
+* FEC_WEB_PASSWORD
+* FEC_WEB_API_KEY
+* FEC_WEB_API_KEY_PUBLIC
+* NEW_RELIC_LICENSE_KEY
+
+Deploys of a single app can be performed manually by targeting the env/space, and specifying the corresponding manifest, as well as the app you want, like so:
+
+```
+cf target [dev|stage|prod] && cf push -f manifest_<[dev|stage|prod]>.yml [api|web]
```
jmcarp: Should this be `cf target -o [dev|stage|prod]...`? Might be worth noting that deploying with `cf push` involves brief downtime.

README.md
ccostino: Ah good to know, I wasn't aware of that.  I had taken what's currently in the API README for now just to get something in place given that the instructions had changed.  @emileighoutlaw are in the midst of updating that documentation but I'll be sure that we incorporate this information in those updates as well as reflecting them here.  There's no sense in perpetuating incorrect info, so thanks for flagging this!

README.md
ccostino: Yes it should - again, this is taken from the API README, but as I mentioned I'll make sure we update there in our open PR as well as here.

Good call on noting the brief downtime, too!

ccostino:
Thank you again @jmcarp for walking through the Flask app initialization improvements with me, much appreciated!

README.md
```diff
@@ -100,7 +100,50 @@ Compile Sass as changes are made:
 
 ### Deployment
 
-See directions in the 18F/openFEC repo.
+##### Likely only useful for 18F team members
```
adborden: Should be h4 `####`, right?

adborden:
:+1:

ccostino:
@adborden thanks for taking a look!  I opted to just remove the heading all together, an emphasized note makes more sense here I think. :-)

PR #1170, opened 2016-04-07
Winnows 404 content

emileighoutlaw:
Related to: https://github.com/18F/fec-style/issues/266#issuecomment-206014332, which has been inspiring me to look more closely at our error and alert messages.

This is a small place we can trim content to help our readers. 

Content-only PR. Anyone can review.


noahmanger:
Thanks! Can you make the change on the CMS as well?

Issue #1158, opened 2016-04-04
Content: Microcopy consistency on data filters


emileighoutlaw:
We've been looking at our microcopy for some time now in this spreadsheet: https://docs.google.com/spreadsheets/d/1-LjKQy3dOJI9FJmrhNEfgsr-uD7HzzH0qf1zpo3azTE/edit?pli=1#gid=1066620775

We've got great ideas for ways we can improve the user experience through consistency of language— now we need to implement those ideas. :)

PR #1171, opened 2016-04-07
Omits "funds"

emileighoutlaw:
"Funds" is a weird word. And in this case, the H5 is followed by a dollar sign, which I think is a clear explainer that we are talking about money. 

This PR proposes that rather than saying "Funds raised" and "Funds spent" we say only "Raised" and "Spent."

Related: https://github.com/18F/openFEC-web-app/issues/1158

Content only. Let me know if anyone has questions or concerns.


noahmanger:
:thu

PR #1174, opened 2016-04-08
Feature/refactor table columns

noahmanger:
This cuts down on the amount of duplicate code needed when creating a new datatable page by moving all column definitions to a separate partial. 

It also cleans up tables.js by moving some functions to columns.js and a new column-helpers.js

I'm sure there's more that can be done to further simplify things, but want to leave it here for now.

I branched off of `feature/new-datatable-pages` and wanted to PR into it to make the code easier to review, since it touches so many files. So we should merge this first, and then #1166

### DIFF
--- a/static/js/modules/column-helpers.js
+++ b/static/js/modules/column-helpers.js
@@ -1,0 +1,137 @@
+'use strict';
+
+var _ = require('underscore');
+
+var tables = require('./tables');
+var helpers = require('./helpers');
+var decoders = require('./decoders');
+
+var sizeInfo = {
+  0: {limits: [0, 199.99], label: 'Under $200'},
+  200: {limits: [200, 499.99], label: '$200—$499'},
+  500: {limits: [500, 999.99], label: '$500—$999'},
+  1000: {limits: [1000, 1999.99], label: '$1000—$1999'},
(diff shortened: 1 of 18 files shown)


static/js/modules/column-helpers.js
```diff
@@ -0,0 +1,137 @@
...
+  return {
+    min_date: dates.min,
+    max_date: dates.max
+  };
+}
+
+function buildTotalLink(path, getParams) {
+  return function(data, type, row, meta) {
+    data = data || 0;
+    var params = getParams(data, type, row, meta);
```
ccostino: Where is `getParams` defined?  It looks like this would result in a reference error.

noahmanger: This is just the arg it's expecting, not that's being passed, correct? 

I just moved this from here: https://github.com/18F/openFEC-web-app/pull/1174/files/106e363de5a403ba6c0d02d1c059a28189235278#diff-53c740c13f86f19e46e710c57d2d56cdL103

ccostino: Oh you're right, it's one of the arguments that comes in.  Sorry, I thought it was referring to a method defined elsewhere in this module, which I couldn't find.  Please disregard.

PR #1175, opened 2016-04-09
Legal search

adborden:
Here's the first pass implementation for the [legal resources search ui](https://trello.com/c/g0MX7L9w/62-hook-ui-up-to-search-api). The goal here is to provide a basic frontend UI for us to validate some assumptions regarding the search strategy in general rather than a pixel perfect UI. I've gone off of the existing styles for now but based the layout off of the direction from @edmullen's [invision app](https://projects.invisionapp.com/share/5N6NL76QM#/screens/145835130).

## Feature flag

We're using the same feature flag from https://github.com/18F/fec-cms/pull/260 to hid the Legal resources nav. It can be enabled by setting the `FEC_FEATURE_LEGAL` environment variable which will be set in our staging environment.

## Future routes

The current thinking is that this is an overall search (via `/legal/search`) and then the user will be able to focus the search on specific document types (`/legal/*`). I think that's the pattern I see with the campaign finance data search. Let me know if there's another pattern. These additional routes will come later as needed.

## Relationship to the existing search

I wanted to keep this work flexible-ish while we're collecting feedback. This UI felt different enough that I decided to use completely separate templates and avoid the search macro for now. I'm open to other ideas, but I feel like we'll know more about how the two are related as we get some feedback.

If the two remain split, renaming the existing search to "data search" might be more appropriate. IDK.

screenshot from 2016-04-08 18-31-18

cc @edmullen @anthonygarvan 

This depends on styles from https://github.com/18F/fec-style/pull/304

### DIFF
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -45,6 +45,18 @@ def load_search_results(query, query_type='candidates'):
     return results['results'] if len(results) else []
 
 
+def load_legal_search_results(query, query_type='all'):
+    filters = {}
+
+    if query:
+        filters['q'] = query
+
+    url = '/legal/search'
+    results = _call_api(url, **filters)
+
+    return results['results'] if len(results) else []
+
+
 def load_single_type(data_type, c_id, *path, **filters):
     data = _call_api(data_type, c_id, *path, **filters)
     return result_or_404(data)
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -44,39 +44,40 @@ def load_search_results(query, query_type='candidates'):
 
     return results['results'] if len(results) else []
 
+def _transform_advisory_opinion(advisory_opinion):
+    source = advisory_opinion['_source']
+    return {
+        'id': source['AO_Id'],
+        'no' : source['AO_No'],
+        'name': source['AO_name'],
+        'summary': source['AO_Summary'],
+        'tags': source['AO_tags'],
+        'description': source['description'],
+        'doc_id': source['doc_id'],
+        'text': source['text'],
+    }
+
+
+def _transform_legal_search_results(data):
+    results = {}
+    results['advisory_opinions'] = [_transform_advisory_opinion(i) for i in data if i['_type'] == 'ao']
+    results['regulations'] = [i for i in data if i['_type'] == 'regulation']
+    results['murs'] = [i for i in data if i['_type'] == 'mur']
+    return results
+
 
 def load_legal_search_results(query, query_type='all', limit=20):
     filters = {}
 
-    def _transform_advisory_opinion(advisory_opinion):
-        source = advisory_opinion['_source']
-        return {
-            'id': source['AO_Id'],
-            'no' : source['AO_No'],
-            'name': source['AO_name'],
-            'summary': source['AO_Summary'],
-            'tags': source['AO_tags'],
-            'description': source['description'],
-            'doc_id': source['doc_id'],
-            'text': source['text'],
-        }
-
-
-    def _transform(data):
-        results = {}
-        results['advisory_opinions'] = [_transform_advisory_opinion(i) for i in data if i['_type'] == 'ao']
-        results['regulations'] = [i for i in data if i['_type'] == 'regulation']
-        results['murs'] = [i for i in data if i['_type'] == 'mur']
-        return results
-
     if query:
         filters['q'] = query
         filters['limit'] = limit
+        filters['type'] = query_type
 
     url = '/legal/search'
     results = _call_api(url, **filters)
 
-    return _transform(results['results'] if len(results) else [])
+    return _transform_legal_search_results(results.get('results', []))
 
 
 def load_single_type(data_type, c_id, *path, **filters):
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -45,16 +45,38 @@ def load_search_results(query, query_type='candidates'):
     return results['results'] if len(results) else []
 
 
-def load_legal_search_results(query, query_type='all'):
+def load_legal_search_results(query, query_type='all', limit=20):
     filters = {}
+
+    def _transform_advisory_opinion(advisory_opinion):
+        source = advisory_opinion['_source']
+        return {
+            'id': source['AO_Id'],
+            'no' : source['AO_No'],
+            'name': source['AO_name'],
+            'summary': source['AO_Summary'],
+            'tags': source['AO_tags'],
+            'description': source['description'],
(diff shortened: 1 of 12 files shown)


openfecwebapp/api_caller.py
```diff
@@ -45,6 +45,40 @@ def load_search_results(query, query_type='candidates'):
...
+        return results
+
+    if query:
+        filters['q'] = query
+        filters['limit'] = limit
+
+    url = '/legal/search'
+    results = _call_api(url, **filters)
+
+    return _transform(results['results'] if len(results) else [])
```
jmcarp: Empty lists are falsy. Why check length here? Also, if `results['results']` is an empty list, do we need to replace it with a different empty list before passing to `_transform`? Sorry if I'm misreading.

openfecwebapp/routes.py
```diff
@@ -187,3 +187,15 @@ def elections(office, cycle, state=None, district=None):
...
+def legal_search():
+    query = request.args.get('search')
+    result_type = request.args.get('search_type') or 'all'
+    results = {}
+
+    # Only hit the API if there's an actual query
+    if query:
+        results = api_caller.load_legal_search_results(query, result_type)
+
+    return views.render_legal_search_results(results, query, result_type)
```
jmcarp: Is `result_type` used in the result templates?

jmcarp:
For some reason, I started commenting on an earlier version of this at #1152--sorry about that. I'm opting not to copy my comments here to avoid email clutter, but I think those comments apply here too.

openfecwebapp/api_caller.py
adborden: I think you're misreading, the length check is on `results`, a dict. We're testing that it has the `'results'` key (via the dict having length), and returning that. Maybe I could rewrite it to be more clear?

```python
return _transform(results['results'] if 'results' in results else [])
```

openfecwebapp/routes.py
adborden: Currently no, but it should be used as part of the HTML select element. I'll fix that.

adborden:
> For some reason, I started commenting on an earlier version of this at #1152--sorry about that. I'm opting not to copy my comments here to avoid email clutter, but I think those comments apply here too.

Sure, I'll make sure to address those as well.

adborden:
From https://github.com/18F/openFEC-web-app/pull/1152#discussion_r59123445
> Just curious--what about applying some or all of these transforms on the api side? Should the api serve idiosyncratically named elastic-style fields like `_type` and `_source`?

Great point, the API transform code does make more sense on the API side. We're working in a prototype mode trying to get the initial pieces together of the back end and front end talking. To move more quickly, I implemented the transform here to avoid having another round of reviews on the API side. This will get cleaned up in the next pull request.

adborden:
https://github.com/18F/openFEC-web-app/pull/1152#discussion_r59123414
> These nested functions don't seem to rely on variables defined in the outer context, so do they need to be nested? Prefer shorter, flatter functions, all things being equal: "flat is better than nested".

:+1:  I'll just move them out as private methods?

adborden:
https://github.com/18F/openFEC-web-app/pull/1152/files#r59123373
> The query_type parameter doesn't seem to be used here. Could we drop it?

It's a parameter I _do_ want to support in the API so I'll add it to the filters so the API is receiving that parameter even if it's not doing anything with it.

openfecwebapp/api_caller.py
jmcarp: Or `result.get('results', [])`?

openfecwebapp/templates/legal-search-results.html
```diff
@@ -0,0 +1,78 @@
...
+{% block title %}
+  Search results for "{{query}}"
+{% endblock %}
+
+{% block body %}
+<header class="page-header slab slab--primary">
+  <span class="page-header__title">Search results</span>
+    <form id="large-search" autocomplete="off" class="search__container js-search">
+      <div class="combo combo--search">
+        <select class="search__select" name="search_type">
```
noahmanger: This `select` should have the class `.select--alt-primary` and the `button` should be `.button--standard`.

adborden:
@noahmanger @jmcarp Thanks for the feedback, I pushed a new commit with those changes.

openfecwebapp/templates/partials/navigation.html
```diff
@@ -47,9 +47,15 @@
     <li class="site-nav__item">
       <a href="{{ cms_url }}/calendar/" class="site-nav__link">Calendar</a>
     </li>
+    {% if features.legal %}
+    <li class="site-nav__item">
+      <a href="{{ cms_url }}/legal-resources" class="site-nav__link">Legal resources</a>
```
noahmanger: Should this link go to `/legal/search` instead for now?

openfecwebapp/templates/partials/legal-search-results-advisory-opinion.html
```diff
@@ -0,0 +1,8 @@
+<tr>
+  <td><a title="{{ advisory_opinion.description }}" href="http://saos.fec.gov/aodocs/{{ advisory_opinion.no }}.pdf#search={{ query }}">{{ advisory_opinion.no }}</a></td>
```
noahmanger: Just curious why you're setting `title` to this here? It shows up as a huge block of text when hovering.

noahmanger:
Had a couple other questions, but nothing to stop it. So if you want to keep as is, lemme know, and I can merge.

openfecwebapp/templates/partials/legal-search-results-advisory-opinion.html
adborden: Yeah, the title is intentional for now. We don't have the real document to link to ATM and I didn't want to include the full text in the table. Using the title is just to give the user an idea of what the document text is so they have an idea why the document was a match in their search. Sort of a prototype state.

If you have other ideas, I'm all ears. Doesn't look like we'll have the document urls for this next sprint so it'll be in this hacky state for a while.

openfecwebapp/templates/partials/navigation.html
adborden: `/legal-resources` is served by the cms, currently just a search box that takes you to /legal/search. @edmullen has a mock-up of that landing page, so more content to land there soon.

openfecwebapp/templates/partials/legal-search-results-advisory-opinion.html
noahmanger: That was what I was guessing. Works for me!

openfecwebapp/templates/partials/navigation.html
noahmanger: 👍

PR #1176, opened 2016-04-11
Filter microcopy

emileighoutlaw:
This is the first round of edits to improve the consistency of our language on data filters. These should be the easy (i.e., not controversial) changes. Even so, I'd definitely like @noahmanger's or @jenniferthibault's eyes on this.

It'd be great to have Lindsay look at this (should she have capacity) but she has already looked at the [Mega Filter Spreadsheet](https://docs.google.com/spreadsheets/d/1-LjKQy3dOJI9FJmrhNEfgsr-uD7HzzH0qf1zpo3azTE/edit?pli=1#gid=99512504) that these changes stem from. 

Resolves: https://github.com/18F/openFEC-web-app/issues/1158

For this week's release.

### DIFF
--- a/openfecwebapp/templates/partials/candidate/financial-summary.html
+++ b/openfecwebapp/templates/partials/candidate/financial-summary.html
@@ -82,7 +82,7 @@
         </div>
         {% endfor %}
       {% endif %}
-      <p>Candidates receive and spend money through <span class="term" data-term="Committee">committees</span>. These are the <strong>combined financial totals</strong> for all of this candidate's <span class="term" data-term="Authorized committee">authorized committees</span>. You can learn more about each committee's fundraising and spending on its page.</p>
+      <p>Candidates receive and spend money through <span class="term" data-term="Committee">committees</span>. These are the <strong>combined financial totals</strong> for all of this candidate's <span class="term" data-term="Authorized committee">authorized committees</span>. Learn more about each committee's fundraising and spending on its page.</p>
     </div>
   </div>
 
--- a/openfecwebapp/templates/partials/candidate/financial-summary.html
+++ b/openfecwebapp/templates/partials/candidate/financial-summary.html
@@ -55,8 +55,8 @@
         </div>
         <div class="callout__action">
           <ul class="callout__sublinks">
-            <li><a href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle, tab='receipts') }}">Receipts</a></li>
-            <li><a href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle, tab='disbursements') }}">Disbursements</a></li>
+            <li><a href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle, tab='receipts') }}">Money received</a></li>
+            <li><a href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle, tab='disbursements') }}">Money spent</a></li>
             <li><a href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle, tab='fillings') }}">Reports filed</a></li>
           </ul>
         </div>
@@ -74,8 +74,8 @@
           </div>
           <div class="callout__action">
             <ul class="callout__sublinks">
-              <li><a href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle, tab='receipts') }}">Receipts</a></li>
-              <li><a href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle, tab='disbursements') }}">Disbursements</a></li>
+              <li><a href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle, tab='receipts') }}">Money received</a></li>
+              <li><a href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle, tab='disbursements') }}">Money spent</a></li>
               <li><a href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle, tab='filings') }}">Reports filed</a></li>
             </ul>
           </div>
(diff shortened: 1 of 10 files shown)


openfecwebapp/templates/partials/filings-filter.html
```diff
@@ -20,7 +20,7 @@
 
   <button type="button" class="js-accordion-trigger accordion__button">Filing date</button>
   <div class="accordion__content">
-    {{ years.years('cycle', 'Two-year period')  }}
+    {{ years.years('cycle', 'Years active')  }}
```
LindsayYoung: To me, "Years Active" doesn't make sense for filings since it is about particular records and not the activity of a group. In this case, cycle is derived from the year the filing applied to. Can we say "Report year" or something like that?

openfecwebapp/templates/partials/independent-expenditures-filter.html
```diff
@@ -20,7 +20,7 @@
 <div class="js-accordion accordion--neutral" data-content-prefix="filter" data-open-first="false">
   <div class="filters__inner">
     {{ typeahead.field('committee_id', 'Spender name or ID') }}
-    {{ years.years('cycle', 'Two-year period') }}
+    {{ years.years('cycle', 'Years active') }}
```
LindsayYoung: Again, "years active" makes me think years that the group is active, in this case we are looking at all events of a particular type in a two-year time span. Can we just say "years" or something like that?

emileighoutlaw:
Thanks for the notes @LindsayYoung. Both those changes should be in now. :)

openfecwebapp/templates/partials/candidate/financial-summary.html
```diff
@@ -55,8 +55,8 @@ <h4 class="callout__title">
         </div>
         <div class="callout__action">
           <ul class="callout__sublinks">
-            <li><a href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle, tab='receipts') }}">Money received</a></li>
-            <li><a href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle, tab='disbursements') }}">Money spent</a></li>
+            <li><a href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle, tab='receipts') }}">Receipts</a></li>
+            <li><a href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle, tab='disbursements') }}">Disbursements</a></li>
```
noahmanger: I get why you're changing this for consistency, but should we want to come up with some rules for when we want to say "money received" and "money spent", which seem easier to understand?

openfecwebapp/templates/partials/tables/candidates-table.html
```diff
@@ -3,7 +3,7 @@
     <tr>
       <th scope="col">Name</th>
       <th scope="col">Office</th>
-      <th scope="col">Election years</th>
+      <th scope="col">Years active</th>
```
noahmanger: I think this is should be "election years" still, as it's the year they filed to run. Right?

openfecwebapp/templates/partials/search-results-candidate.html
```diff
@@ -8,9 +8,9 @@ <h3 class="result__title">
     </h3>
     <span class="result__years t-data">
       {% if candidate.election_years[0] != candidate.election_years[-1] %}
-        Election years: {{ min(candidate.election_years) }}–{{ max(candidate.election_years) }}
+        Years active: {{ min(candidate.election_years) }}–{{ max(candidate.election_years) }}
```
noahmanger: I think this should also be "election years".

LindsayYoung: Yea, we changed the behavior in the last few weeks so that it really does use election year to filter on. I did a few tests to confirm and it seems to be filtering by the election year. For example, if you pick a non-presidential cycle and presidential filter, no candidates come up, which would be the desired behavior.

LindsayYoung: The "yea" was me agreeing with Noah, incase that was not clear from the comment

emileighoutlaw: Ah, that makes sense. Reversed this edit.

openfecwebapp/templates/partials/tables/candidates-table.html
emileighoutlaw: You are right! Fixed

openfecwebapp/templates/partials/candidate/financial-summary.html
emileighoutlaw: Yeah, I don't know— I need to think about this. Rather than deploying a change today, I reversed this edit, so I can consider it and come up with a well-reasoned plan.

emileighoutlaw:
Ready for you to take a second look whenever, @noahmanger. Thanks for your notes!

openfecwebapp/templates/partials/committees-filter.html
```diff
@@ -9,11 +9,11 @@
 <div class="js-accordion accordion--neutral" data-content-prefix="filter" data-open-first="false">
   <div class="filters__inner">
     {{ typeahead.field('q', 'Committee name or ID', False, dataset='committees', allow_text=True) }}
-    {{ years.years('cycle', 'Two-year period') }}
+    {{ years.years('cycle', 'Election year') }}
```
noahmanger: Sorry, but this one I think actually _should_ be "Years active" since it's based on the year they have financial activity.

emileighoutlaw:
Okay— I think this is right now. Thanks for sticking with me and helping me understand this better, @noahmanger and @LindsayYoung <3

openfecwebapp/templates/partials/committees-filter.html
emileighoutlaw: Not a problem — change is made

PR #1187, opened 2016-04-15
Beginning mega menu work

noahmanger:
Prepares the nav for the mega menu from https://github.com/18F/fec-style/pull/316
- Removes the old sub-lists
- Re-arranges some nav components
- Adds a global variable for the CMS url


noahmanger:
Ok this is ready for review as well.

adborden:
:+1: pending testing + merge of https://github.com/18F/fec-style/pull/316

package.json
```diff
@@ -67,7 +67,7 @@
     "d3": "3.5.5",
     "datatables.net": "1.10.10",
     "datatables.net-responsive": "2.0.1",
-    "fec-style": "~1.10",
+    "fec-style": "git+https://github.com/18f/fec-style#site-menu",
```
adborden: Should we leave this as is, and then push a separate change to bump `fec-style`? Would rather not have this pointing to a branch.

PR #1195, opened 2016-04-19
Fix initial data table fetch.

### LINKED ISSUE #1192: Bug: URL not being used by data table query
A friend wrote to me over the weekend to report a bug that's currently on production: When you go to a URL with a pre-populated filter query, the query is not used at first. If you un-select and re-select a filter, however, it will work. 

For example, this should link to a page of Obama for America disbursements: https://beta.fec.gov/data/disbursements/?recipient_name=C00431445&max_date=04-18-2016

But when you first go to that page it shows all disbursements until you modify the filters.

I'll look into the problem and it would be good to get a hotfix out soon.

jmcarp:
As described in #1192, a recent refactor to data tables caused a
regression such that filters in query parameters were ignored on initial
fetch. This happened because the `DataTable` class was instantiated
before the `filterSet` attribute was set; since data tables depend on
the `filterSet` attribute to create filters, and since fetching starts
as soon as the `DataTable` is instantiated, the `filterSet` attribute
must be set first. This patch restores the order of operations and adds
a comment explaining why it matters.

[Resolves #1192]

### DIFF
--- a/static/js/modules/tables.js
+++ b/static/js/modules/tables.js
@@ -403,6 +403,14 @@ function DataTable(selector, opts) {
   this.hasWidgets = null;
   this.filters = null;
 
+  // Set `this.filterSet` before instantiating the nested `DataTable` so that
+  // filters are available on fetching initial data
+  if (this.opts.useFilters) {
+    this.filterPanel = new FilterPanel();
+    this.filterSet = this.filterPanel.filterSet;
+    $(window).on('popstate', this.handlePopState.bind(this));
+  }
+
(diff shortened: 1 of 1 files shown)


noahmanger:
Thank you! I forgot to mention that I wanted to do this as a hotfix, so I went ahead and made the same change in a hotfix branch here: https://github.com/18F/openFEC-web-app/pull/1196

Also, I noticed that in your branch the tags weren't generated, which was solved by moving the `tagList` instantiation to the same spot as the `filterPanel` code. 

Closing this in favor of #1196

PR #1196, opened 2016-04-19
Moving filter panel and widgets code

noahmanger:
This is a hotfix version of @jmcarp 's PR: https://github.com/18F/openFEC-web-app/pull/1195

As he explained:

> As described in #1192, a recent refactor to data tables caused a
> regression such that filters in query parameters were ignored on initial
> fetch. This happened because the DataTable class was instantiated
> before the filterSet attribute was set; since data tables depend on
> the filterSet attribute to create filters, and since fetching starts
> as soon as the DataTable is instantiated, the filterSet attribute
> must be set first. This patch restores the order of operations and adds
> a comment explaining why it matters.

Additionally, this patch moves the `tagList` instantiation before the DataTable class to ensure the filter tags are generated by the URL as well.

### DIFF
--- a/static/js/modules/tables.js
+++ b/static/js/modules/tables.js
@@ -405,6 +405,8 @@ function DataTable(selector, opts) {
 
   this.$widgets = $(DATA_WIDGETS);
 
+  // Set `this.filterSet` before instantiating the nested `DataTable` so that
+  // filters are available on fetching initial data
   if (this.opts.useFilters) {
     var tagList = new filterTags.TagList({title: 'All records'});
     this.$widgets.find('.js-filter-tags').prepend(tagList.$body);
--- a/static/js/modules/tables.js
+++ b/static/js/modules/tables.js
@@ -403,19 +403,21 @@ function DataTable(selector, opts) {
   this.hasWidgets = null;
   this.filters = null;
 
+  this.$widgets = $(DATA_WIDGETS);
+
+  if (this.opts.useFilters) {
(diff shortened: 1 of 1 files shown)


jmcarp:
Did you want to drop the code comment and commit message? I figured it would be worthwhile to include a note in the code about why order matters here so we don't have more regressions later on. Any reason not to make your changes on top of the patch I sent in?

noahmanger:
Oops, yeah I'll bring back the comment as well. 

I did a clean hotfix because `develop` — which you branched off of — was ahead of `master`, so I wanted to branch off `master` instead. Or am I misunderstanding?

jmcarp:
Let me preface this by saying that my comments here are nit-picky and not a big deal. Anyway, unless the original patch conflicted with master, renaming the branch or cherry-picking would've worked, and we would've kept the more detailed commit message. Plus, I'm happy to make changes on work that I submit. I'm not suggesting a change here, but next time let me know, and I'll send revisions.

noahmanger:
That makes sense. It didn't conflict, but there were a number of commits I didn't want to merge in. I didn't think to cherry pick just your commit and wanted to save you the work since it was an easy fix. I'm happy to just have the code comment in there, but will definitely do that next time. Thanks!

ccostino:
Awesome, thank you both @jmcarp and @noahmanger!  This looks good and fixes the issue.  Duly noted for the next time in terms of workflow, too.

PR #1199, opened 2016-04-21
Improvements to the legal resources search

adborden:
The next iteration of the legal resources search as per https://github.com/18F/fec-eregs/issues/62.
- [x] Pagination on advisory opinions search page
- [x] Use `pdf_url` from API
- [x] Use total count from API, not limited to 10
- [x] Use snippet text from API with term highlights
- [x] Create advisory opinion search page at `/legal/advisory-opinions`
- [x] Review from the design team.

### DIFF
--- a/openfecwebapp/templates/legal-doc-search-results.html
+++ b/openfecwebapp/templates/legal-doc-search-results.html
@@ -14,7 +14,7 @@ Searching {{ document_type_display_name }} for "{{ query }}"
 <div class="main">
   <div class="container">
     <section class="main__content--full">
-      <div id="results-{{ result_type }}" class="content__section">
+      <div id="results-{{ result_type }}" class="content__section dataTables_wrapper">
         <div class="results-info results-info--top">
           <div class="results-info__left">
             <h2 class="results-info__title">{{ document_type_display_name }}</h2>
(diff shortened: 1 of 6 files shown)


adborden:
Accidentally committed `manifest_feature.yml` which wasn't meant for this PR, removed it.

openfecwebapp/templates/legal-doc-search-results.html
```diff
@@ -0,0 +1,64 @@
...
+{% import 'macros/legal-search.html' as legal_search %}
+{% set document_type_display_name = 'Advisory Opinions' %}
+
+{% block title %}
+Searching {{ document_type_display_name }} for "{{ query }}"
+{% endblock %}
+
+{% block body %}
+<header class="page-header slab slab--primary">
+  <span class="page-header__title">Legal resources // {{ document_type_display_name }}</span>
```
noahmanger: Either in this PR or separately, we should update this area to use the new breadcrumb pattern: 
image

(`/templates/macros/breadcrumbs.html`)

openfecwebapp/templates/partials/legal-search-results-advisory-opinion.html
```diff
@@ -1,8 +1,18 @@
...
-    <div class="t-serif" title="{{advisory_opinion.text}}">{{ advisory_opinion.text|truncate }}</div>
+    <div class="t-serif legal-search-result__hit">
+      &hellip;
+      {% for highlight in advisory_opinion.highlights %}
+        {{ highlight|safe }} &hellip;
+      {% endfor %}
+    </div>
+    {% if legal_include_display_all %}
+    <div class="legal-search-result__display-all">
+      <a href="/legal/advisory-opinions?search={{ query }}">Display all hits</a>
```
noahmanger: Just curious, why include this in each hit? More of a design decision than coding, so doesn't need to hold up merging, but it caught my eye.

noahmanger:
LGTM. I glanced at the python, since it looks like most is moving to the API. But the HTML is good and everything works.

Couple questions. Let me know if you just want to address those separately and then I'll merge.

openfecwebapp/templates/legal-doc-search-results.html
adborden: Cool, I'll address this in a [separate PR](https://github.com/18F/openFEC-web-app/issues/1204).

openfecwebapp/templates/partials/legal-search-results-advisory-opinion.html
adborden: Great point. I believe the intention here is that "Display all hits" doesn't actually take you to the advisory-opinion search, but it really just expands the row to show you all the hit text. I misunderstood the design and just included all the hit text concatenated with ellipsis. https://github.com/18F/fec-eregs/issues/76 will address this, but for now I think it's okay to go to the advisory opinion page. We can also remove this link if it's misleading.

cc @edmullen

noahmanger: Yeah, I think I'd vote for removing it (even just removing the `href` so you can still see how it looks with that link there), but don't feel too strongly.

PR #1203, opened 2016-04-27
Configure logging to stdout

adborden:
This will configure the logger to send output to stdout, and therefore `cf logs` and logs.cloud.gov.

### DIFF
--- a/openfecwebapp/app.py
+++ b/openfecwebapp/app.py
@@ -22,6 +22,7 @@ from openfecwebapp.env import env
 
 locale.setlocale(locale.LC_ALL, '')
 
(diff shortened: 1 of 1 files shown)


adborden:
I would suggest adding json formatting for `prod` so that lines to logs.cloud.gov don't get split.

adborden:
This would be the [change for json logging](https://github.com/adborden/openFEC-web-app/compare/configure-logging...adborden:json-logging), but those messages aren't getting indexed properly on the logs.cloud.gov side. Following up with the support team but we should probably merge this as-is for now. Better than nothing :)

cc @LindsayYoung

PR #1205, opened 2016-04-28
Updating alert messages throughout

noahmanger:
Updates the copy and implementation of messages throughout to be in line with https://github.com/18F/fec-style/issues/266#issuecomment-214425298

## 404 error
image

## 500 error
image

## Committee page with no data
image

## Search results
image

image

^ Note: we don't currently get a result count for how many results are on the other page we're linking to, so had to differ from the mockup in that way.

## Table with no results
image
This is the other area we had to differ. Because of the way this message is currently implemented, it's difficult to implement including the committee name and election cycle in the message, as sometimes the message shows up on an election, candidate or committee page. How does this more generic language work?

Alright, I think that's everything.


noahmanger:
CC @jenniferthibault and @emileighoutlaw

jenniferthibault:
I added a few small tweaks to quotation mark style and a suggestion on the table with no results that's totally up for discussion.

Other to do's / to discuss:

**404**
- Either remove the glossary icon for “committee” or add a glossary icon for “candidate” as well. (This is something i'm only seeing now, but is how it was before, so not your fault!)

**Elections search / multiple districts**
I missed this one before, sorry! The alert for a zip code with multiple congressional districts should probably be an informational alert instead of a warning. I couldn't find where this lived in the code to change it:
screen shot 2016-04-28 at 11 43 25 am

**Committee page with no data**
- Could we also make the selected election cycle bold, like you have the committee name? I think it will help to emphasize elements or categories that the user can change

noahmanger:
Thanks for the close look and for making the changes. I just made those edits. 

Re: the "no data in the table" alert, I'll think about it a bit more and see if I can find a way to make it work with variables. How would it work on election pages, such as an Electioneering Communications table that showed totals for multiple candidates? What would the language be?

jenniferthibault:
I totally defer to @emileighoutlaw judgement here for a final call!

noahmanger:
Bump on this. @jenniferthibault is this looking good to you?

openfecwebapp/templates/partials/search-results-alternatives.html
```diff
@@ -1,11 +1,11 @@
 <div class="message message--alert">
   {% if not results %}
   <h2 class="message__title">No results</h2>
-  <p>Sorry, no results match your search, <span class="t-bold">"{{ query }}."</span></p>
+  <p>Sorry, no results match your search, <span class="t-bold">“{{ query }}.”</span></p>
```
emileighoutlaw: I think this should say `<p>Sorry, no results match your search for <span class="t-bold">“{{ query }}.”</span></p>`

emileighoutlaw:
I made a tiny copyedit suggestion— otherwise looks good to me

noahmanger:
Ok! I made it so that we show specific language in the various table error messages. I went one step further to explain what data specifically we don't have in the table.

## Candidate pages
image

## Committee pages
image

image

image


## Election pages
image

noahmanger:
@emileighoutlaw how do those new messages look to you?

jenniferthibault:
Wow, hound really hates you on this one, Noah. 😁 

One last visual bug I'm seeing (that I SWEAR we already took care of somewhere?) is that the button options in all of the screens you just posted should have about double the margin left/right spacing between them. This is to prevent them from being misinterpreted as toggle options. Really, I swear we got this somewhere, but I can't find it now, or maybe I'm overlooking it. 
Thanks for your patience while I poke at the tiniest of details.

noahmanger:
Ah! Good eye. You're right, we got it in one place, but not this. Fixed:

image

openfecwebapp/templates/legal-search-results.html
```diff
@@ -2,7 +2,7 @@
 {% import 'macros/legal-search.html' as legal_search %}
 
 {% block title %}
-  Search results for "{{query}}"
+  Search results for “{{query}}”
```
ccostino: It looks like these are smart double-quotes (or smart-style double-quotes) - is this an intended change or should these still be plain text double-quotes?

openfecwebapp/templates/partials/search-results-alternatives.html
```diff
@@ -1,17 +1,19 @@
 <div class="message message--alert">
   {% if not results %}
-  <h3>Sorry, there are no results for your search "{{ query }}"</h3>
+  <h2 class="message__title">No results</h2>
+  <p>Sorry, no results match your search, <span class="t-bold">“{{ query }}.”</span></p>
```
ccostino: Smart-like quotes here as well; again, is this intentional?  If it is, we should be using the HTML entities that correspond to them instead: https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references

openfecwebapp/templates/search-results.html
```diff
@@ -3,7 +3,7 @@
 {% import 'macros/page-header.html' as header %}
 
 {% block title %}
-    Search results for "{{query}}"
+    Search results for “{{query}}”
```
ccostino: Another call-out for smart quotes, just in case!

static/js/modules/tables.js
```diff
@@ -486,7 +487,7 @@ DataTable.prototype.fetchError = function() {
 DataTable.prototype.hideEmpty = function(response) {
   if (!response.pagination.count) {
     this.destroy();
-    this.$body.before('<div class="message message--alert">No data found.</div>');
+    this.$body.before(missingTemplate(this.opts.hideEmptyOpts));
```
ccostino: Nice to see this change to a template instead!

static/js/pages/committee-single.js
```diff
@@ -383,6 +408,11 @@ $(document).ready(function() {
           columns: disbursementPurposeColumns,
           callbacks: aggregateCallbacks,
           order: [[1, 'desc']],
+          hideEmptyOpts: {
+            dataType: 'disbursements',
+            name: context.name,
```
ccostino: Why is hound complaining so much about this here?  It looks like `context` is fine elsewhere, which has me wondering if there a variable definition or function argument missing.

noahmanger: This is an old hound comment. I added it as a [global var](https://github.com/18F/openFEC-web-app/pull/1205/files#diff-93a270b5de8f2744399104a7ac896334R3) at the top of the file, but after the errors.

openfecwebapp/templates/legal-search-results.html
noahmanger: Ah. Unintentional. Fixing!

ccostino:
I just typed in `iilf` in the search box on the main page when you first go to betaFEC, which defaults to searching `Candidates` in the drop down menu.  However, the following screen suggests finding `Committees` instead:

screen shot 2016-05-18 at 5 23 55 pm

The display of the error message is correct, but shouldn't it suggest viewing `Candidate` results instead in this scenario?

noahmanger:
Are you referring to the error message? That link is a link to a "committee" search for that query instead.

ccostino:
Ah so the intention is that if no "candidate" results were found then the suggestion is for viewing potential "committee" results instead, and vice versa?

noahmanger:
Yep, exactly. Not the most elegant thing, but it's the way it's been for a while.

ccostino:
I think we're good to go now, thanks everyone!

PR #1207, opened 2016-04-28
Removing duplicate code

noahmanger:
Everywhere else we moved the filter pane initialization to the tables.js file, but forgot to remove it from here.


noahmanger:
Meant to target the release branch. Closing.

PR #1211, opened 2016-05-12
Feature/candidate flag ui

noahmanger:
- Adds the `five_thousand_flag` to all navigation URLs to the candidates page (resolves https://github.com/18F/openFEC/issues/1616)
- Fixes an issue where the candidate office pages weren't rendering
- Updates the language for the federal funds filter (resolves https://github.com/18F/openFEC/issues/1654)
- Rewords titles of datatable pages from "[type] Results" to just "[type]". So now "Candidates" instead of "Candidate results". I did this because on office pages, it sounded weird to have "Candidates for president results" instead of just "Candidates for president" (cc @jenniferthibault )
- Adds the rest of the data to the details panels for candidate pages (resolves #https://github.com/18F/openFEC/issues/1496)

Adding the rest of the data to the details panels involved making extra calls to the API on opening the panel. I modeled this after the code we had for the filings tables, but had to make some changes, so would appreciate extra eyes on that in particular. Building the correct links to committee and election pages that use the correct cycle variables can be tricky, but I'm pretty sure I got it.

To test the details panels, visit the following pages:
- http://localhost:3000/candidates
- http://localhost:3000/candidates/president (also where the federal funds filter lives)
- http://localhost:3000/candidates/senate
- http://localhost:3000/candidates/house

cc @adborden or @ccostino


noahmanger:
Just updated this with the rest of the detail panel work and updated the initial comment with more info on that.

openfecwebapp/templates/candidates-single.html
```diff
@@ -45,7 +45,7 @@ <h1 class="entity__name">{{ name }}</h1>
           </span>
           {% for c in committee_groups['P'] %}
           <span class="entity__term__data">
-            <a href="{{ url_for('committee_page', c_id=c.committee_id, cycle=c.related_cycle) }}">{{ c.name }}</a>
+            <a href="{{ url_for('committee_page', c_id=c.committee_id) }}">{{ c.name }}</a>
```
ccostino: I just wanted to make sure I understand the change - this will take you to the committee page itself instead of something tied to a cycle and was a desired change if I remember a recent conversation correctly, right?

openfecwebapp/templates/macros/filters/election-filter.html
```diff
@@ -25,11 +25,13 @@
       <option value="{{ year }}">{{ year }}</option>
     {% endfor %}
   </select>
+  {% if office in ['president', 'senate'] %}
```
ccostino: No need for `house` here?

static/js/modules/helpers.js
```diff
@@ -82,6 +82,31 @@ Handlebars.registerHelper('panelRow', function(label, options) {
...
 
+Handlebars.registerHelper('entityUrl', function(entity, options) {
+  if (options.hash.query) {
+    var query = {
+      cycle: options.hash.query.cycle || null,
+      election_full: options.hash.query.election_full || null
+    };
+  }
+  var id = entity.candidate_id || entity.committee_id;
+  var url = buildAppUrl([options.hash.type, id], query);
```
ccostino: Hound is getting confused here, but for good reason.  There's a mechanism known as variable hoisting in JavaScript; what this means is any variable defined in a method anywhere is going to be instantiated first at the beginning of the function at run-time, regardless of where it's used.  This, combined with JavaScript's block and scoping rules are what are triggering the alarms here and had me tripped up at first too.

What you can do is define the variables in the method first and set them after.  This helps account for the variable hoisting and ensures the proper scope is recognized.  At the beginning of the method, I would do this before anything else:

`var id, query, url;`

And then remove the `var` statements elsewhere in the method.  More information on variable hoisting can be found [here](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting) and [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#var_hoisting).

openfecwebapp/templates/macros/filters/election-filter.html
noahmanger: Nope. We don't show the time period for house races (because they're only two years).

openfecwebapp/templates/candidates-single.html
noahmanger: Yep, exactly. Otherwise links could try to take you to a future cycle for a committee page that would return a 404.

openfecwebapp/templates/macros/filters/election-filter.html
ccostino: Ah, gotcha!

ccostino:
Thanks, @noahmanger!  I had some initial feedback while I go through and test things out locally to look more closely, but regarding these URLs:

* http://localhost:3000/candidates/president
* http://localhost:3000/candidates/senate
* http://localhost:3000/candidates/house

Are they intended to be accessible anywhere on the site or not just yet?

noahmanger:
Not yet! They'll be in the new menu.

ccostino:
Oh okay, excellent! :-)

static/js/modules/helpers.js
```diff
@@ -118,6 +146,28 @@ function buildUrl(path, query) {
     .toString();
 }
 
+function getTimePeriod(electionYear, cycle, electionFull, office) {
+  var durations = {
+    P: 3,
+    S: 5,
+    H: 1
```
ccostino: These offsets seem a bit odd to me.  For instance, if I'm looking at a presidential election candidate, I'm seeing a full cycle of 2013 - 2016 being used.  Is this intended?  I am probably just misunderstanding the reasoning here, but wanted to call it out just in case.

noahmanger: Good question. Basically it's "[year after the last election] - [year of the next election]". Same logic as here:
image

static/js/modules/helpers.js
noahmanger: Ah! Good catch. Will make the change.

static/js/modules/helpers.js
ccostino: Okay, I had figured there was a reason/precedent for it, but wasn't sure.  Thank you for clarifying!

ccostino:
Okay, in looking into what it would take to support downloads for these I am noticing a couple of things that I am trying to work through.

The `resource` that serves as the entry point into processing in the API is different than the other `resources` we allow for exports in that its underlying `schema` is empty in the model.  This is because the `schema` associated with the model is a `page_schema`, not just a `schema`.  Therefore the `schema` property on the `resource` is just empty and doesn't exist.  Bypassing this yields a different underlying problem in that the class mapping used to tie all of these things together yields an object not expected by the code processing the download.

All of this leads me to a couple of questions that I'm trying to answer to enable support:

* Are we tying things to the correct model under the hood?
  * If so, how can we make models based on a `page_schema` work?
  * If not, then what do we need to do to change things appropriately?

So for the time being, exports should be disabled until we can unwind this a bit more.  Is the intention to have downloads working with these output in time for this sprint's end?

The remaining bits of this PR look okay to me if we want to go ahead and merge what's done so far, but I just want a bit more time to see if the export issue is solely on the API side or if there's a tie here as well that needs to be changed that would make this work.

noahmanger:
Thanks for looking into it. I'm ok releasing these pages with the exports disabled. I think the focus should continue to be on downloads for receipts and disbursements, but once those are addressed we get these going.

ccostino:
@noahmanger: Okay excellent, then in that case we can move forward with merging this as is.  @adborden, if you have a moment, did you still want to take a look?  If not I'll merge in about an hour or so.

ccostino:
Per follow-up conversations in Slack, this has been deemed good to go for now, so we are merging this in.  Thanks, @noahmanger!

PR #1212, opened 2016-05-12
Feature/add regulations to search

anthonygarvan:
This PR covers the removal of transformations in this app (they are now handled during ETL + api), as well as added templates to display regulations. Should me merged / deployed simultaneously with 18F/openFEC#1657.


ccostino:
As with the other PR, thanks again @anthonygarvan, great to run through a live pair review! No issues found during our session.

PR #1213, opened 2016-05-18
Remove .results-info--top

adborden:
This subclass is being dropped from fec-style. This should be merged first, before bumping the fec-style version.

Related: https://github.com/18F/fec-style/pull/342


adborden:
@noahmanger

PR #1217, opened 2016-05-20
Removing federal funds filter

noahmanger:
Temporary removal while we get the endpoint working. 

cc @ccostino


ccostino:
All good, thanks @noahmanger!

PR #1233, opened 2016-05-24
Adds support for typeahead filter control

noahmanger:
Adds support for the new `FilterControls` component introduced in https://github.com/18F/fec-style/pull/361


adborden:
:+1:

PR #1236, opened 2016-05-25
Changed flags from five_thousand_flag to has_raised_funds

jontours:
Flags changed, ready for review.

### DIFF
--- a/openfecwebapp/templates/partials/candidates-filter.html
+++ b/openfecwebapp/templates/partials/candidates-filter.html
@@ -18,8 +18,8 @@
       <legend class="label">Candidate fundraising</legend>
       <ul>
         <li>
-          <input id="cycle-checkbox-five-thousand" name="five_thousand_flag" type="checkbox" value="true">
-          <label for="cycle-checkbox-five-thousand">Has raised more than $0</label>
+          <input id="cycle-checkbox-has-raised-funds" name="has_raised_funds" type="checkbox" value="true">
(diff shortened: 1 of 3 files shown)


openfecwebapp/templates/partials/candidates-filter.html
```diff
@@ -19,7 +19,7 @@
       <ul>
         <li>
           <input id="cycle-checkbox-five-thousand" name="five_thousand_flag" type="checkbox" value="true">
```
noahmanger: This `name` attribute needs to be changed to `has_raised_funds` too.

ccostino: The `name` attribute needs to change to `has_raised_funds` as well, and I would also change the `id` attribute to reflect this (so make it `cycle-checkbox-has-raised-funds`).

openfecwebapp/templates/partials/candidates-filter.html
```diff
@@ -19,7 +19,7 @@
...
-          <label for="cycle-checkbox-five-thousand">Has raised more than $5,000</label>
+          <label for="cycle-checkbox-five-thousand">Has raised more than $0</label>
```
ccostino: In addition to the above change, the `for` attribute should change as well (`cycle-checkbox-has-raised-funds`).

ccostino:
Thanks, @jontours!

PR #1241, opened 2016-05-31
Adding advanced data page

noahmanger:
Creates the advanced data landing page at `/data/advanced/`. 

In the course of coding this I wanted to try a few things that departed slightly from the designs [here](https://github.com/18F/openFEC/issues/1681). Sharing here to discuss with @jenniferthibault and @emileighoutlaw . 

Rather than showing the "all data" link in the right column, I think it communicates a clearer relationship to have it above the list of all data sets that it includes (I also indented the sublist to make the hierarchy even clearer). Here's how it looks for candidates:

image

And committees:
image

This is particularly important when we get to Spending:
image

In this case, Operating Expenditures and Expenditures to Committees are included in the "All Disbursement" dataset. But the other three on the right are mutually exclusive, each of which represents an entire set in and of itself, hence the "all data" icon.

I also reversed the "filings" and "report summary" columns, since we don't have any report summary tables yet:
image

And last, I removed the "interactive overview" card from the bottom, as we have each of the cards above:
image

What do folks think? If this organization tweak makes sense, I'll apply it to the mega menu so that it's all consistent.


emileighoutlaw:
I like it a lot. The other thing you did here (intentionally or not) was add "candidates" after each of the candidate sub-categories. I think that's a really smart edit. Something was bugging me about the way  that section read, and I think it was that missing word— I never included it because I wanted as few words as possible, but I think having it there makes everything a little easier to scan.

jenniferthibault:
The only part that doesn't feel correct to me is using the "All" icon and idea for IEs ECs, and CCs. Prior, we've been using that signifier to mean "a group of data categories" which are then broken down into subcategories that we offer. As you said, IEs, ECs and CCs are complete data sets of themselves, they don't combine groups. The same reason we're not using it for Statements of candidacy or RFAI's. Does that make sense?

Could we also put the Advanced data icon in a circle, like this?
screen shot 2016-06-01 at 11 28 52 am

jenniferthibault:
Let me know when you're ready for a style fine-tune and we can pair 👯 
There's a few tiny things I'd like to finesse

noahmanger:
Ok this is ready to go whenever the fec-style PR is merged (tests are failing because the change isn't in yet).

jenniferthibault:
A small thing for @noahmanger :
- We should match the breadcrumb title to the page title

screen shot 2016-06-01 at 6 30 10 pm

A larger thing for @emileighoutlaw :
- **Before this gets merged**, we should replace the latin text with real content to each data section

screen shot 2016-06-01 at 6 46 34 pm

emileighoutlaw:
Per our release discussion, we're going to launch without the intro text, which I'll keep working on here: https://github.com/18F/openFEC-web-app/issues/1245

static/js/init.js
```diff
@@ -114,6 +115,9 @@ $(document).ready(function() {
     new Search($(this));
   });
 
+  // Initialize table of contents
+  new toc.TOC('.js-toc');
```
adborden: Didn't notice this before, but this looks different from our convention of passing in a set of jquery elements, rather than a selector string. I'm fine with either approach, but would prefer to keep it consistent.

openfecwebapp/templates/datatable.html
```diff
@@ -2,7 +2,7 @@
 {% import 'macros/page-header.html' as header %}
 
 {% set breadcrumbs_title=title %}
-{% set breadcrumbs=[('', 'Advanced data')] %}
+{% set breadcrumbs=[(url_for('advanced'), 'Advanced data')] %}
```
adborden: Do we want to disable this for now since we don't have the content for this page?

adborden:
Happy to merge this, pending the `fec-style` PR, but the breadcrumb link means that users can access it with the placeholder text. If we remove the breadcrumb link, the page will not be reachable to users and we could get this merged.

noahmanger:
Cool. I'm just going to remove the placeholder text, so we'll be good.

static/js/init.js
noahmanger: Good catch. Yeah, some of our components take a selector string, some take an actual $ element.

PR #1242, opened 2016-05-31
Adding report year filter

noahmanger:
Adds the `rpt_yr` filter to receipts, disbursements and individual contributions pages.

image

A few questions on this:

1. What should the `name` of this field be? Is it `rpt_yr` or something more readable like `report_year`?
2. Even if it's report year, I think we may want to label it something more friendly, but I'm not sure what that is (cc @emileighoutlaw )
3. Should values be presented as two-year periods (e.g. `2015-2016`) or individual years? 
4. What should the tooltip say? I think it ought to both explain what the filter does as well as why it's necessary to always have a year selected. Current language:
> The two-year period that includes the election which the transaction was reported for. Due to the large number of itemized transactions, you can only view one period at a time.

Requires https://github.com/18F/openFEC/pull/1429
Requiers https://github.com/18F/fec-style/pull/380

Resolves https://github.com/18F/openFEC/issues/1674


emileighoutlaw:
Thanks for tagging me in! 

I think `Election year` is the most human understandable option. If we use that, I think it should be for a single year, not a range. I think many folks will struggle to understand why something called `Election year` (singular) contains only multi-year ranges.

So it'd read like:
> **Election year**
> -2015
> -2014
> -2014 (and so on)

noahmanger:
Updated this to meet the needs of the way the filter works now. See https://github.com/18F/fec-style/pull/380 for a full explanation.

noahmanger:
Most of the stuff for this is in the FEC Style PR, but I'd like to know in particular if the validation stuff makes sense.

Also, I want to draw attention to a particular change. I replaced the call to `pushQuery` with `updateQuery`, which means that when the filters change, we update the URL, but not the browser's history. This is because with the new transaction period filters, we need to set filters by default if the page is loaded without them. With `pushQuery`, this creates an infinite history of pages, breaking the ability to use the browser back button.

With this change, you can no longer use “back” to move between queries; instead, “back” tacks you to the last distinct page you were on. 

Curious what folks (namely @jenniferthibault and @onezerojeremy ) think about this change. I think it's more natural: if you come to this page and change the filters, it updates the URL. But you can click "back" to go back to the page you were previously on.

jenniferthibault:
Smart! Using back to navigate through pages, not tag changes seems like the intended interaction to me as well.

onezerojeremy:
I think that's probably right, as well. Does seem like it'll feel more natural. :)


That infinite browser updates making back not work is always pretty frustrating for me.

ccostino:
@noahmanger:  Sorry it's taken a bit of time to get to reviewing this and answer your questions, but for your first and third question, the field in question is `transaction_year` in the API and it takes a single year value that should represent the end of a cycle (e.g., `2016` for 2015 - 2016).

I'll see if I can get this branch (and fec-style) updated with recent changes and account for the API support properly and push the updates to keep things moving, otherwise I'll work with you when you and I have a moment to catch up!

ccostino:
@noahmanger, don't forget to bring this up to date with `develop` to resolve the conflicts! :-)

LindsayYoung:
rpt_yr should be spelled out on the API as report_year 

Did we miss this somewhere? If it is a new introduction, lets change it on the API side. If not we can leave it so we don't have braking changes.

ccostino:
@LindsayYoung it is, I think @noahmanger was just referencing the column directly.  We expose it as `report_year` though.

noahmanger:
Fixed the merge conflict, but this is going to fail until fec-style has been updated. This can be reviewed though.

ccostino:
I think this is good to go as well once the bump in fec-style is accounted for (https://github.com/18F/fec-style/pull/380 is merged).  Thanks, @noahmanger!

ccostino:
Ready for merge, thanks @noahmanger!

Issue #1245, opened 2016-06-02
Advanced data page content


emileighoutlaw:
This issue is where I'll draft the introductory content for our advanced data index. 

To include:
- Candidates
- Committees
- Receipts
- Spending

![unnamed](https://cloud.githubusercontent.com/assets/10552702/15754350/d0726e78-28c5-11e6-9170-1840d4277c76.png)

We'll incorporate as fast as I can get this written and approved

jenniferthibault:
A small content question on this page. Looking at consistency between the breadcrumbs and the page title, `Advanced data` or `Advanced data index`, which should it be?

<img width="579" alt="screen shot 2016-06-06 at 10 16 43 am" src="https://cloud.githubusercontent.com/assets/11636908/15824974/d5fb4ea2-2bcf-11e6-851a-38608c05ba06.png">

emileighoutlaw:
Good question. I've been preferring `Advanced data` to `advanced data index`, so I'd like to go with that, unless someone else makes a strong case against it.

noahmanger:
Changed the title!

emileighoutlaw:
GitHub issue: https://docs.google.com/document/d/1nNcHO7dK6qvjnLbCpIrdChXw7k82wTKGfP36Y-7zurw/edit

I'll ping appropriate folks when I have a draft done :)

emileighoutlaw:
okay @noahmanger and @jenniferthibault — I'm coming in under the Friday wire, but this content draft is finished: https://docs.google.com/document/d/1nNcHO7dK6qvjnLbCpIrdChXw7k82wTKGfP36Y-7zurw/edit#

jenniferthibault:
Thanks Emileigh! I left a few clarifying questions in your doc.

emileighoutlaw:
@jenniferthibault is taking one more peek! :)

jenniferthibault:
Thanks for the bump! I had meant to offer some design solutions for your card idea for linking the reg/reporting side.

I think this is a good use of the pattern we established on the reg/reporting side where we link to reporting deadlines. How about this visual pattern?

![screen shot 2016-06-24 at 3 39 42 pm](https://cloud.githubusercontent.com/assets/11636908/16348668/0cf580e8-3a22-11e6-942a-e1b7d208558a.png)

noahmanger:
Oh I love that! Though we do have separate pages for PAC and party committee requirements

jenniferthibault:
[...]

PR #1258, opened 2016-06-06
Feature/20160608 fixes

noahmanger:
- Fixes an issue where the sticky menu on the advanced data page didn't stick
- Adjusts the title of the advanced data page and replaces hard-coded title with a template variable

### DIFF
--- a/openfecwebapp/templates/advanced.html
+++ b/openfecwebapp/templates/advanced.html
@@ -13,7 +13,7 @@
   </div>
   <div id="options" class="container">
     <div class="sidebar-container sidebar-container--left">
-      <nav class="sidebar sidebar--neutral sidebar--left side-nav js-sticky js-toc" data-sticky-container="options">
+      <nav class="sidebar sidebar--neutral sidebar--left side-nav js-sticky-side js-toc" data-sticky-container="options">
         <ul class="sidebar__content">
           <li class="side-nav__item"><a class="side-nav__link" href="#candidates">Candidates</a></li>
           <li class="side-nav__item"><a class="side-nav__link" href="#committees">Committees</a></li>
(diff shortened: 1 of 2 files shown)


openfecwebapp/templates/advanced.html
```diff
@@ -2,18 +2,18 @@
...
+  {{ title }}
 {% endblock %}
 
 {% block body %}
-  {{ header.header('Advanced data', '', show_search=False) }}
+  {{ header.header(title, '', show_search=False) }}
   <div class="main">
   <div class="container">
-    <h1 class="icon-heading--circle-table">Advanced data index</h1>
+    <h1 class="icon-heading--circle-table">{{ title }}</h1>
```
adborden: Not a big deal, but this does change the copy: Advanced data index -> Advanced data

noahmanger: Yep, that was the intention. @emileighoutlaw noted the desired change [here](https://github.com/18F/openFEC-web-app/issues/1245#issuecomment-223981578)

adborden:
Not a huge deal, but this branch looks like it's based on `develop`. If the fix is targeted for the release, it should be based on the release branch with something like `git checkout -b bug/release-fix origin/release/release-20160608`. That will avoid pulling in extra commits happening on `develop`.

adborden:
Oh, I should add, that of course you would need to create an extra PR to merge into `develop` if you want the fix to also exist on dev right away. Otherwise, the fix will eventually land on `develop` once the `release` branch is closed via git flow.

noahmanger:
Oh thanks. Yeah, I should have branched off of release. And yeah, it's ok to wait for this.

PR #1260, opened 2016-06-08
Replacing max_receipt_date with max_date

noahmanger:
And replacing hard-coded date with generated date. cc @LindsayYoung


adborden:
:+1:

PR #1272, opened 2016-06-15
Updating login

LindsayYoung:
To do Travis and local deploys

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -96,7 +96,14 @@ def deploy(space=None, branch=None, yes=False):
 
     # Log in if necessary
     if os.getenv('FEC_CF_USERNAME') and os.getenv('FEC_CF_PASSWORD'):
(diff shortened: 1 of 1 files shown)


adborden:
Is this necessary? I'm not familiar with all the implications of the switch to SSO, but my understanding is that bots and humans now login differently. Humans need `cf login --sso` while bots shouldn't.

The idea of checking the env variables was to enable us as humans to login however we want, and never set those env variables. Then the `invoke deploy` option will skip the auth step and just start running commands. Assuming you've already logged in yourself it should work fine. For bots, the env variables would be set, so the login/auth step would trigger.

adborden:
Oh, and [this comment](https://github.com/18F/openFEC/pull/1731#issuecomment-226638914) applies too, we want to add a `cf target` for the manual deploy case.

LindsayYoung:
this got fixed in a different pr

PR #1278, opened 2016-06-17
Fixes issue with undefined sort parameter

### LINKED ISSUE #1271: Disbursement table error
Web app is passing "-undefined" as a sort parameter into the API and that brakes the disbursements. We need to add a valid default sort or not pass a sort value at all. 

Here is the example, but, they seem to be everywhere:
https://beta.fec.gov/data/committee/C00043489/?tab=disbursements

ccostino:
Closes #1271 

This changeset addresses the issue with the undefined sort parameter being added to the API calls when viewing the Disbursements tab on a single Committee page.  Note that we also had to excplicitly set the purpose value to be lower case for this to work.

h/t to @noahmanger for the huge assist in helping track down where this was being constructed and the right values to set.

/cc @LindsayYoung

### DIFF
--- a/static/js/pages/committee-single.js
+++ b/static/js/pages/committee-single.js
@@ -147,7 +147,15 @@ var filingsColumns = columnHelpers.getColumns(
 
 var disbursementPurposeColumns = [
   {data: 'purpose', className: 'all', orderable: false},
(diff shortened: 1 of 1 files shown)


LindsayYoung:
I'll review now

PR #1290, opened 2016-06-20
Feature/top-lists

noahmanger:
- [x] Top candidates
- [x] Top PACs
- [x] Top party committees

Top candidates:
image

### DIFF
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -148,7 +148,25 @@ def landing_mock_data():
 def load_top_candidates(sort):
         response = _call_api(
             'candidates', 'totals',
-            sort_hide_null=True, election_year=2016, cycle=2016, election_full=True, sort=sort, per_page=5
+            sort_hide_null=True, election_year=2016, cycle=2016, election_full=False, sort=sort, per_page=5
+        )
+        if response['results']:
+            return response['results']
+        return {}
+
+def load_top_pacs(sort):
+        response = _call_api(
+            'totals', 'pac',
+            sort_hide_null=True, cycle=2016, sort=sort, per_page=5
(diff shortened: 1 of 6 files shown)


noahmanger:
Ok @jontours I added the code for top PACs and party committees. The test data only has one of each for 2016, but it seems to be working. Thanks for helping me get the API branch going!

jontours:
@noahmanger I did push another subset that has more than one party committee, if you want to see more interesting results.  It was tough writing this feature as the fec_vsum_f3x table was pretty sparse in the subset data.  

Also added a commitee_type filter to the api.  So for example, if you do a get on the 'party' endpoint and pass a parameter, "type=X', you will get back only X-type party committees, don't know if that will be needed, but may be nice to have for other unforeseen features

adborden:
:+1:

jontours:
This looks good for me except for candidates, I'd say it's ready to merge as the broken candidates is probably on the api.  Also, the formatting is wacky for me but could be my local set-up.

noahmanger:
Cool. I just pushed one more commit to fix a styling issue. But if that looks good then we can merge this. I'll make a separate PR into the release branch.

jontours:
This can be closed without merging, was referenced by newer pull request, which has been already merged.

jontours:
Ah wait, crap, this was for develop.  The feature branch should be merged back into develop as part of the deployment correct?  @adborden @ccostino

adborden:
Ah, yep good point @jontours. Yeah, I double checked that this branch made it into the release branch. We can merge this now, otherwise it will be merged back into develop later today when we finish the release and merge the branches. I don't think it matters either way. One less merge commit in the history if we wait :)

PR #1291, opened 2016-06-21
Feature/tooltip adjustments

noahmanger:
Makes the necessary markup changes to support the adjusted tooltip styles here https://github.com/18F/fec-style/pull/402

Also refactors the "unique" filter into a partial that's shared by individual contributions and receipts filter panels.


ccostino:
@noahmanger, it looks like this PR needs to be brought up to date with recent changes.

PR #1292, opened 2016-06-21
Address itemized downloads feedback

ccostino:
This changeset addresses a few things that were flagged as needing to be changed based on the recent feedback we received.  These changes include the following:

* Modifies the link in the export data tooltip

These modifications are a part of the full list in https://github.com/18F/openFEC/issues/1748

/cc @noahmanger, @LindsayYoung, @jontours

### DIFF
--- a/static/js/modules/tables.js
+++ b/static/js/modules/tables.js
@@ -36,7 +36,7 @@ var DOWNLOAD_MESSAGES = {
     'Exports are limited to ' +
     downloadCapFormatted +
     ' records—add filters to narrow results, or export bigger ' +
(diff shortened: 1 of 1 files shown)


noahmanger:
Can you include the rounding in this one too?

ccostino:
Yes - adding a WIP label on this, sorry, it was late last night.

ccostino:
Okay, I think I have the proper support in place, but I want to make sure my approach is correct.  Also, in the case that the number is less than 1,000, would it make sense to then round to the nearest hundredth or even tenth?  Or is that overkill and we should just focus on results greater than 1,000?

noahmanger:
I think that's ok to just round for over 1,000. I did a quick eyeball check on queries for around 30, 100 and 340 and the numbers were accurage.

Issue #1275, opened 2016-06-16
Write content for "About this data" on landing page


noahmanger:
In order for advanced users to understand what data is included in the totals shown on the landing page, write blurbs to explain what the data is. The blurbs should address:
- [x] The time period included. I believe this is everything reported from January 1, 2015 to May 25, 2016. 
- [x] The forms included. I believe this is Forms 3, 3P, 3X and 5, but someone should verify
- [x] Anything else necessary to understand what is or is not included

@PaulClark2 can you weigh in here? This is referring to the charts generated from the data you gave me in that spreadsheet a couple weeks ago.

![image](https://cloud.githubusercontent.com/assets/1696495/16124784/56ec8742-33a5-11e6-8173-e19cfe4da7b4.png)

emileighoutlaw:
Word. Thanks for opening this, Noah.

PaulClark2:
@noahmanger 
1- January 1, 2015 to May 25, 2016 is correct. 
2 - The data includes Forms 3, 3P, 3X, 5, 7 and 9
3 - I'll post the formulas

PaulClark2:
Money Spent = PAC and Party Adjusted Disbursements + Congressional 
Adjusted Disbursements + Presidential Adjusted Disbursements + Independent Expenditures (Form 5 Filers Only) + Communication Costs + Electioneering Communications 

Money Raised = PAC and Party Adjusted Receipts + Congressional Adjusted Receipts 
+ Presidential Adjusted Receipts + Total Contributions (Form 5 filers) + Total Donations (Form 9 filers)

**Congressional Formulas
(all Form 3 filers)**

**Disbursements**
Adjusted Disbursements equals Total Disbursements 
minus the sum of Transfers to Other Authorized Committees, Loan Repayments Made, Total Contribution Refunds and Other Disbursements

(Form 3 Line Numbers: Adjusted Disbursements = Form 3 Line 22 - (Line 18 + Line 19(c) + Line 20(d) + Line 21))

**Receipts** 
Adjusted Receipts equals Total Receipts 
minus the sum of Contributions from Political Party Committees, Contributions from Other Political Committees, Offsets to Operating Expenditures, Total Loan Repayments Made and Total Contribution Refunds

(Form 3 Line Numbers: Adjusted Receipts = Form 3 Line 16 - (Line 11(b) +  Line 11(c) + Line 14 + Line 19(c) + Line 20(d)))

**Presidential Formulas
(all Form 3P filers)**

[...]

PR #1293, opened 2016-06-21
Adding methodology content

noahmanger:
Adds the content from https://github.com/18F/openFEC-web-app/issues/1275 to the "About this data" sections.

image

A couple design thoughts: There's a lot of content here. I know we never squared away the rules for what goes in these boxes, and have talked about having a separate methodology page. I think that that makes sense to do in the future, but this is fine for now. 

In the interim, we could either keep this as-is or set a max-height on these accordions with an inner scroll. What do you think @jenniferthibault ?

### DIFF
--- a/openfecwebapp/templates/landing.html
+++ b/openfecwebapp/templates/landing.html
@@ -88,7 +88,7 @@
             <li>Offsets to operating expenditures</li>
             <li>Transfers from nonfederal accounts for allocated activities</li>
           </ul>
-          <p><span>The form-by-form breakdown for adjsuted receipts is:</span></p>
+          <p><span>The form-by-form breakdown for adjusted receipts is:</span></p>
           <ul class="list--bulleted">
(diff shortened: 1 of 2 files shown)


jenniferthibault:
I added one commit for a few quick typo fixes that seemed to come from copying markdown >> html. 

Setting a max-height with an inner scroll makes sense to me until we can determine a pattern of what type of content goes in here. Before moving to a separate methodology page, can we add analytics trackers to these accordion buttons to see how many people click to open them?

emileighoutlaw:
This looks really good. Thanks, y'all.

Would a Methodology page include everything under the current "Methodology" section? That makes sense to me.

noahmanger:
Ok updated with the click tracking. The fixed height requires this though https://github.com/18F/fec-style/pull/408

jenniferthibault:
Looks good, thank you!

Issue #1243, opened 2016-06-01
Add filter for "has received federal funds" to presidential page


noahmanger:
Now that we have the flag on the API, we need to add a filter to the front-end on the /candidates/president data page to allow users to filter to candidates who have received federal funds.

- [ ] Add filter
- [ ] Ensure flag shows up correctly in details panel

jontours:
Once this gets reviewed and merged: https://github.com/18F/openFEC/pull/1797, this issue can be closed.

noahmanger:
So I'm trying to run this agains the dev api, and I'm still not seeing the flag show up in the result: https://fec-dev-api.18f.gov/v1/candidates/totals/?api_key=rnPUWfzbUwoehuMKifk0aEbBLepJUefTB0ekXMkZ&sort_hide_null=true&election_year=2016&cycle=2016&election_full=true&federal_funds_flag=true&sort=-receipts&per_page=30&page=1&office=P

@jontours am I missing something?

jontours:
I just ran that pull request locally and replicated the argument parameters and got back results, with the flags included in the json. I would guess that the API didn't deploy automatically?

PR #1295, opened 2016-06-21
Adding presidential public funds flag filter

noahmanger:
Adds the "Has received presidential public funds" filter to the frontend:

image

## Dependency
It appears, though, that this flag isn't included in the response. @jontours can you add it?

Resolves https://github.com/18F/openFEC-web-app/issues/1243


jontours:
I can take a look at this now.

noahmanger:
Bingo! This works now. Thanks @jontours .

static/js/modules/helpers.js
```diff
@@ -117,6 +117,19 @@ Handlebars.registerHelper('electionUrl', function(year, options) {
   return new Handlebars.SafeString(url);
 });
 
+Handlebars.registerHelper('capitalize', function(value) {
```
xtine: Where's capitalize being used?

noahmanger: I originally used it instead of `convertBoolean` and then decided to keep it in as it might be useful down the road and I already wrote it. Should we remove it?

xtine: I would err on the side of removing it until it's needed again so it doesn't clutter up `helpers.js` if it doesn't ends up being used.

noahmanger: 👍  Removed.

PR #1309, opened 2016-06-23
Fix/accessibility fixes

noahmanger:
- Wraps data tables to ensure correct sizing
- Adds `aria-hidden` to `figure__decimals` so they won't be read by screen readers
- Replaces reaction radio buttons with regular buttons
- Fixes duplicate IDs on the bar chart components
- Removes a slash from a checkbox ID that was throwing an HTML_CodeSniffer error

Goes with https://github.com/18F/fec-style/pull/411

In response to https://github.com/18F/Accessibility_Reviews/issues/15#issuecomment-228167208

### DIFF
--- a/static/js/modules/bar-charts.js
+++ b/static/js/modules/bar-charts.js
@@ -15,15 +15,18 @@ var stripes =
   '</svg>';
 
 
-function GroupedBarChart(selector, data) {
+function GroupedBarChart(selector, data, index) {
   this.element = d3.select(selector);
   this.data = data;
+  this.index = index;
 
   this.margin = {top: 0, right: 20, bottom: 50, left: 40};
   this.height = 320 - this.margin.top - this.margin.bottom;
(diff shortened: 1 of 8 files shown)


static/js/modules/bar-charts.js
```diff
@@ -106,6 +109,7 @@ GroupedBarChart.prototype.buildChart = function() {
     .enter()
       .append('g')
       .attr('class', function(d) { return 'period ' + d.status; })
+      .attr('aria-labelledby', 'bar-chart-tooltip-' + this.index)
```
noahmanger: @nickbristow : Is this the right place to add `aria-labelledby`? Hovering on the `g` element is what triggers the tooltip.

ccostino:
Thanks, @noahmanger!

static/js/modules/bar-charts.js
nickbristow: @noahmanger sorry I got swamped. Yes this is correct!

PR #1312, opened 2016-06-24
Updates bulk URL

emileighoutlaw:
Jeff pointed out that our export button says 

"Exports are limited to 100,000 records—add filters to narrow results, or export bigger data sets with FEC bulk data exporter."

But we link to the data catalog which doesn't have any bulk data export capability. 

This updates the URL to go to this page: http://www.fec.gov/finance/disclosure/ftp_download.shtml

Resolves: https://github.com/18F/FEC/issues/356


noahmanger:
Thanks, Emileigh. @ccostino actually already fixed this in the release branch. So I'm going to close this and then it will be on develop once we merge the release back in.

emileighoutlaw:
Oh word. Thanks

Issue #1304, opened 2016-06-23
Optimize data table column widths 


noahmanger:
So that users can clearly read the data in the data tables, ensure that columns are sized appropriately. 

**Criteria for completion:**
The following pages should have columns with the proper widths:
- [x] Candidates
- [x] Candidates > President
- [x] Candidates > Senate
- [x] Candidates > House
- [x] Committees
- [x] Receipts
- [x] Individual contributions
- [x] Disbursements
- [x] IEs
- [x] ECs
- [x] CCs
- [x] Filings
- [x] Candidate pages
- [x] Committee pages
- [x] Election pages

*This issue consolidates the following issues:
- https://github.com/18F/openFEC-web-app/issues/1263
- https://github.com/18F/FEC/issues/268
- https://github.com/18F/FEC/issues/369
- https://github.com/18F/FEC/issues/314
- https://github.com/18F/FEC/issues/131
- https://github.com/18F/openFEC/issues/1759

noahmanger:
Ok! I've been spending some time on thi 

![image](https://cloud.githubusercontent.com/assets/1696495/16327478/21d9ceb8-3984-11e6-80d1-f99c3c58569a.png)

![image](https://cloud.githubusercontent.com/assets/1696495/16327479/27a10136-3984-11e6-8e0a-baa88f7568ae.png)

![image](https://cloud.githubusercontent.com/assets/1696495/16327480/2e75841e-3984-11e6-989d-7cf6cd4ded98.png)

Remaining to dos:
- [ ] Check how they look with open detail panel
- [ ] Check how they look at different screen sizes

noahmanger:
One thing I'd love to solve: We have several columns that include _really_ long words, which can't break onto multiple lines:
![image](https://cloud.githubusercontent.com/assets/1696495/16352787/8c9634ba-3a26-11e6-9ac7-2b7000b92a28.png)

In some cases, these words are the longest thing in the column by far, forcing the column to take up more space than the data needs, and also they tend to run into the sort icons (which are just background images). @jenniferthibault any ideas? @emileighoutlaw any thoughts on how we could shorten language while still maintaining clarity?

emileighoutlaw:
It looks like these columns are right below a big header that says "Disbursements" I be we can make "Disbursement amount" and "Disbursement date" just say "Amount" an "Date" without anyone being confused.

I'll keep thinking on the other two long column headers

noahmanger:
[...]

PR #1313, opened 2016-06-25
Feature/column-sizing

noahmanger:
This includes various and sundry improvements to the way that data tables present data. Combined with the style changes in https://github.com/18F/fec-style/pull/412 , it should help make our data tables more useful and more legible. 

- Re-arranges columns throughout for a more readable order (see my note about the rules for organizing columns below)
- Disables the `autoWidth` option on all data tables
- The style change turns off truncating cells, so the full cell content will always be visible
- Applies classes to columns to control sizing in CSS rather than in the Javascript. These size classes are then applied to try to constrain the widths of cells and force others to be larger
- Improves the logic around which columns are shown when the panel is open
- Only shows the "independent expenditures" column on the single committee filing tab if it's the kind of committee that makes IEs
- Fixes a bug where all filing table rows had the expand button, when only certain ones should 

For example:
image

image

image

Resolves https://github.com/18F/openFEC-web-app/issues/1304
Depends on https://github.com/18F/fec-style/pull/412

I still definitely want to pair with @jenniferthibault on this, but wanted to get the PR in before the weekend.

## Thoughts on organizing rules
As I was working on this I started sketching out a set of rules that could hopefully guide further decisions around how to present columns. Here's a draft:
- Number columns should always be right-aligned and should always be positioned furthest to the right, which is more readable when numbers are right-aligned and also allows for easier scanning
- If a row is a transaction, the amount of the transaction should always be furthest to the right and the committee reporting the transaction should always be furthest to the left (notably changes receipt and disbursement column order)
- On desktop, when a panel is open, the first 2-3 columns should be the only ones shown in order to prevent a jarring experience of columns appearing to be out of order
- On tablet, when a panel is open, only the 1-2 columns should be shown
- In general, the first column should be the candidate or committee who made transaction or submitted a report, on order to have a clear "subject" to the row
- Any table with chronological data should be sorted by date by default

These aren't hard and fast rules, but I think should only be broken with good reason.


noahmanger:
@jenniferthibault raises a good point default sorting by date: are there times when tables should be sorted by amount by default? I'm not sure, but could definitely see that being the case.

noahmanger:
This also now sets a max for the tables with bars in the cells that persists after paginating, so the bars decrease in size as you page through.

Resolves https://github.com/18F/openFEC/issues/1704

And with that this is ready for code review. @jenniferthibault gave the 👀  check. We still have somewhat of a lingering issue with figuring out a better solution for preventing column titles from overlapping with sort icons, but that doesn't need to hold this up.

static/js/modules/columns.js
```diff
@@ -117,104 +115,99 @@ var committees = [
...
-  dateColumn({data: 'transaction_date', className: 'min-tablet hide-panel-tablet'}),
+  currencyColumn({data: 'transaction_amount', className: 'min-tablet hide-panel column--med column--number'}),
+  dateColumn({data: 'transaction_date', className: 'min-tablet hide-panel column--med'}),
   modalTriggerColumn
 ];
 
 var disbursements = [
   {
-    data: 'recipient_name',
+    data: 'committee',
```
LindsayYoung: just double checking the change `recipient_name` should be replaced by `committee`

static/js/pages/committee-single.js
```diff
@@ -136,15 +136,21 @@ var occupationColumns = [
   }
 ];
 
+var columnKeys =   [
+  'pdf_url', 'amendment_indicator', 'receipt_date', 'coverage_end_date',
+  'total_receipts', 'total_disbursements', 'modal_trigger'
+];
+
+if (context.showIndependentExpenditures) {
```
LindsayYoung: So smart

PR #1314, opened 2016-06-28
Adds advanced content

emileighoutlaw:
I took a stab at adding the advanced data content. There were lots of `<div>`s on this page, so please let me know if I put these paragraphs in the wrong section. 

This PR:
- Adds content to "Candidates," "Committees," "Receipts," "Spending," and "Filings"
- Links terms in the new content to the glossary
- Makes "Filings and Reports" header "Filings," which is our preferred blanket term for all types of filings (this is also explained in the section's new subtext).

cc @noahmanger @jenniferthibault

### DIFF
--- a/openfecwebapp/templates/advanced.html
+++ b/openfecwebapp/templates/advanced.html
@@ -33,7 +33,7 @@
         <h2>Candidates</h2>
         <div class="option__content">
           <p>Search <span class="term" data-term="candidate">candidate</span> data, including money raised, money spent, cash on hand and debt.</p>
-          <p>Statements of Candidacy (Form 2) contain basic information about individuals running 
+          <p>Statements of Candidacy (Form 2) contain basic information about individuals running
           for federal office, including their names, addresses and authorized campaign committees.</p>
           <div class="grid--2-wide">
           <ul class="grid__item list--spacious t-sans">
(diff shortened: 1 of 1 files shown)


emileighoutlaw:
Partially addresses: https://github.com/18F/openFEC-web-app/issues/1245#issuecomment-228914265

openfecwebapp/templates/advanced.html
```diff
@@ -120,6 +133,12 @@ <h1 class="icon-heading__title">{{ title }}</h1>
       <div id="spending" class="option">
         <h2>Spending</h2>
         <div class="option__content">
+          <p>Spending includes all the types of <span class="term" data-term="disbursement">dibsursements</span> candidates and committees make.</p>
+          <p>Search all disbursements to candidates and committees by the recipient, purpose, amount and date.</p>
```
noahmanger: Sorry, I should have caught this sooner: this should be "Search all disbursements **from** committees..." right?

openfecwebapp/templates/advanced.html
```diff
@@ -97,6 +106,10 @@ <h1 class="icon-heading__title">{{ title }}</h1>
       <div id="receipts" class="option">
         <h2>Receipts</h2>
         <div class="option__content">
+          <p><span class="term" data-term="receipt">Receipts</span> are anything of value, like money, services or property. Search 
+          candidate and committee receipt data by contributor, amount and date.</p>
```
noahmanger: Sorry, another thing I should have caught: do we want to add something like "Receipts include contributions from individuals as well as from other committees" in order to make clear that "individual contributions" are one type of receipt?

openfecwebapp/templates/advanced.html
jenniferthibault: Looks like it should be `from` to me as well

jenniferthibault:
@noahmanger the formatting on this looks 👍  to me. Should we change `to` > `from` and get it into the release as is, then follow up with @emileighoutlaw on Tuesday to submit a patch if she agrees with your second comment ?

noahmanger:
Good call @jenniferthibault . Made the fix.

emileighoutlaw:
Y'all are totally right about this language change. Good catch on that, and thank you!

PR #1315, opened 2016-06-29
[MERGE WITH GIT FLOW] is_individual for contributor page 

LindsayYoung:
Moving the individual page form using entity type, which is not consistently populated, to is_individual which looks for unique individual contributions. We create that and it is populated for all the data.


noahmanger:
This still doesn't seem right. I'm seeing a bunch of committee contributions:
image

PR #1323, opened 2016-06-30
[MERGE WITH GITFLOW] Fixing incorrect bar data

noahmanger:
Fixes incorrect bar chart data as a result of me being a bad copy-and-paster.


adborden:
:+1: Not sure where this data is coming from though.

PR #1324, opened 2016-07-01
[MERGE WITH GITFLOW] Emptying tooltip on hide

noahmanger:
Fixes a bug on /data where hovering over the "Spending" chart and then hovering over the "Raising" chart would cause the page to crash.

### DIFF
--- a/static/js/modules/bar-charts.js
+++ b/static/js/modules/bar-charts.js
@@ -269,7 +269,9 @@ GroupedBarChart.prototype.showTooltip = function(x0, d) {
 };
 
 GroupedBarChart.prototype.hideTooltip = function() {
-  this.tooltip.style({'display': 'none'});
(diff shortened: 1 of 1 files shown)


static/js/modules/bar-charts.js
```diff
@@ -269,7 +269,9 @@ GroupedBarChart.prototype.showTooltip = function(x0, d) {
 };
 
 GroupedBarChart.prototype.hideTooltip = function() {
-  this.tooltip.style({'display': 'none'});
+  this.tooltip
+    .style({'display': 'none'})
+    .html('');
```
adborden: very weird. Can you use `.empty()` instead?

noahmanger: Hmmm doesn't appear so. It's not a jQuery object.

PR #1332, opened 2016-07-01
Feature/top lists

noahmanger:
Adding top lists to the release branch.

### DIFF
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -148,7 +148,25 @@ def landing_mock_data():
 def load_top_candidates(sort):
         response = _call_api(
             'candidates', 'totals',
-            sort_hide_null=True, election_year=2016, cycle=2016, election_full=True, sort=sort, per_page=5
+            sort_hide_null=True, election_year=2016, cycle=2016, election_full=False, sort=sort, per_page=5
+        )
+        if response['results']:
+            return response['results']
+        return {}
+
+def load_top_pacs(sort):
+        response = _call_api(
+            'totals', 'pac',
+            sort_hide_null=True, cycle=2016, sort=sort, per_page=5
+        )
+        if response['results']:
+            return response['results']
+        return {}
+
+def load_top_parties(sort):
+        response = _call_api(
+            'totals', 'party',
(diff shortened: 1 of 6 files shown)


openfecwebapp/api_caller.py
```diff
@@ -148,7 +148,25 @@ def landing_mock_data():
...
+        if response['results']:
+            return response['results']
+        return {}
+
+def load_top_pacs(sort):
+        response = _call_api(
+            'totals', 'pac',
+            sort_hide_null=True, cycle=2016, sort=sort, per_page=5
+        )
+        if response['results']:
```
ccostino: On the offhand chance that `results` is not a key that exists in the `response` dictionary, this would throw a `KeyError` exception and cause the page load to fail.  We could write this `if` statement and corresponding `return` statements like this:

 `return response.get('results', {})`

This will return either the `results` dictionary if it exists or the empty dictionary if it doesn't, and won't result in a `KeyError` being thrown.

noahmanger: Oh cool. So in that case the page would load but without data? Is this a pattern we want to use in our other API calls in this file?

ccostino: I see the other methods are also written this way as well, it might be worth modifying them just to make the code a bit more resilient to potential errors.  I ran into this scenario when I tested the code against a database that didn't have the requisite tables/views in place yet, so the `response` object didn't get filled out with the data that was expected.

noahmanger: Cool. Yeah, actually a bunch of different methods in this file are written the same way (that's where I got the approach from). How about we make a separate issue for this, since we'd also need to design the error state.

ccostino: Right, at least that's my understanding looking at most of the code in this file (in some cases, the `results` are just a list, not a dictionary): if nothing is there, return an empty object of some sort so the page will still render.

I agree that making a separate PR for this makes sense given that multiple methods could be improved and that a proper error state should be designed to handle that scenario.

ccostino:
Per conversation above, we'll revisit the error handling to improve it down the road (issue #1341)!

Issue #1334, opened 2016-07-05
Links to receipts and disbursements need transaction period 


LindsayYoung:
With out supplying the right 2-year period, it will default to the current two-year period and look for dates outside of the table. 

Example:

The link to all receipts on this page:
https://beta.fec.gov/data/committee/C00171140/?cycle=1984&tab=receipts

links to:
https://beta.fec.gov/data/receipts/?committee_id=C00171140&min_date=01-01-1983&max_date=12-31-1984&is_individual=true

instead of:
https://beta.fec.gov/data/receipts/?two_year_transaction_period=1984&committee_id=C00171140&min_date=01-01-1983&max_date=12-31-1984&is_individual=true

I think these links are on:
 - committee pages
 - election pages

noahmanger:
Ok, I have a fix for this here: https://github.com/18F/openFEC-web-app/pull/1336

noahmanger:
Resolved by #1336

Issue #1335, opened 2016-07-05
Filings table detail panel triggers are cut off


noahmanger:
On staging, the filings table on committee pages details triggers are cut off:
![image](https://cloud.githubusercontent.com/assets/1696495/16592678/aabe6d7c-4296-11e6-9574-8ba6e7a8d96a.png)

noahmanger:
Resolved by #1336

PR #1336, opened 2016-07-05
Fix/transaction period urls

noahmanger:
This does the following: 
- Adds `two_year_transaction_period` to all aggregate links of individual contributions on committee and election pages
- Instead of the aggregate links going to /receipts/ they now go to /receipts/individual-contributions/ 
- Fixes an issue where the transaction period filter `name` attribute was reverted back to the old name
- Fixes an issue with tooltip placement caused by some style adjustments 

Resolves https://github.com/18F/openFEC-web-app/issues/1334

### DIFF
--- a/openfecwebapp/templates/macros/filters/date.html
+++ b/openfecwebapp/templates/macros/filters/date.html
@@ -31,14 +31,14 @@
 {% macro partition_field(name, label, dates) %}
   {% set current_year = dates['year'][1] | date(fmt='%Y') %}
   <div class="filter js-filter js-select-filter js-filter-control" data-modifies-filter="{{ name }}" data-modifies-property="data-transaction-year" data-required-default="{{ current_year }}">
+    <label class="label t-inline-block" for="two-year-transaction-period">Transaction period</label>
     <div class="tooltip__container">
-      <label class="label tooltip__trigger-text" for="transaction-year">Transaction period</label>
       <button class="tooltip__trigger" type="button"><span class="u-visually-hidden">Learn more</span></button>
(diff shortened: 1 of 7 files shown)


noahmanger:
cc @LindsayYoung . I'm pretty sure I made the change everywhere it needs to be but would like your eyes.

Also cc @jenniferthibault and @onezerojeremy : I just realized that every time we link an aggregate of individual contributions (e.g. "all contributions between $200-$500 for a committee") we were still linking to the top level receipts page. Since we now have an individual contributions page, I figured we should be linking here instead. Does that make sense?

onezerojeremy:
Sounds smart to me!

jenniferthibault:
Sames!

noahmanger:
I also used this to set `autoWidth: false` on the committee filings table to fix the bug with the panel trigger buttons getting cut off.

Resolves https://github.com/18F/openFEC-web-app/issues/1335

noahmanger:
Ping @LindsayYoung on this. I'd like to get this into this release, but would like for you to review.

ccostino:
On the whole, this looks good to go to me!

One small thing though:  when looking at the `Other spending` tab on an election page (e.g., 2012), those links around the totals columns for independent expenditures and communication costs have the `two_year_transaction_period` flag in them.  Neither of those underlying tables/views require that (at least not yet in schedule e’s case), but it doesn’t look like it’s hurting anything either because the parameter is being stripped out in the underlying API request.

LindsayYoung:
Thanks Noah!

Issue #1268, opened 2016-06-13
Discription on the $200 bucket of aggregates by size 


LindsayYoung:
Any time we are looking at contributions by size, "Under $200" should say "$200 and under"

noahmanger:
So should our buckets be this?

- $0.01 - $200
- $200.01 - $500
- $500.01 - $1000
- $1000.01 - $2000
- Over $2000

noahmanger:
Paging @PaulClark2 . Does this sound right? Currently our buckets are:

- 0-199.99: 'Under $200'
- 200-499.99: '$200—$499'
- 500-999.99: '$500—$999'
- 1000-1999.99: '$1000—$1999'
- 2000+: 'Over $2000'

PaulClark2:
We are open to reevaluating these buckets. These buckets predate IEOPCs (Super PACs). It might make sense to break up the $2,000+ bucket or even consolidate some of the smaller dollar amount buckets while adding additional larger dollar amount buckets.

If we stick with our current buckets, the first bucket should be "$0.01 - $200.00: "$200 and under.""  So, my only change is the the first bucket. The others are arbitrary. For years we've used these:
$200 and Under:   $0.01 to $200.00
$200.01 - $499:    $200.01 to $499.99
$500 - $999:         $500.00 to $999.99
$1000 - $1999:     $1,000 to $1,999.99
$2000 and Over:   greater than or equal to $2,000.00

@jwchumley @amykort

LindsayYoung:
This issue is just for one label on the committee page. Though, doing a search for more instances is a good idea.

Here is where it is wrong in the contribution size label on the committee page-
https://beta.fec.gov/data/committee/C00411330/?tab=receipts

It is correct on the election page:
https://beta.fec.gov/data/elections/senate/VT/2012/

We have it documented correctly in the API:
https://api.open.fec.gov/developers/#!/schedules%2Fschedule_a/get_committee_committee_id_schedules_schedule_a_by_size

emileighoutlaw:
Just popping in to say that for the top and bottom buckets, I think we should say:

> $200 and under
> $2,000 and over

noahmanger:
Cool. Thanks all. We have a bigger question about what these buckets are over in #1321 , so for the time being I'll go with the existing buckets and @emileighoutlaw 's language rec.

PR #1337, opened 2016-07-05
Using the correct numbers and labels for size buckets

noahmanger:
As reported in https://github.com/18F/openFEC-web-app/issues/1268, we used different language and logic for the contribution size buckets on committee pages vs election pages. This makes them consistent.

On committee pages:
image

On election pages:
image

Resolves https://github.com/18F/openFEC-web-app/issues/1268

CC @LindsayYoung


LindsayYoung:
Thanks!

PR #1338, opened 2016-07-06
Adding reports data table pages

noahmanger:
This creates pages for presidential, house and senate, pac and party, and ie only reports data table pages. By default, these pages only show the most recent, unamended version of the report, and are sorted by coverage date. 

I have a separate issue started here with API changes that this will depend on: https://github.com/18F/openFEC/issues/1804

Sharing this early as a WIP for visibility into API needs.

Resolves https://github.com/18F/openFEC/issues/1790
## Update on this:
- This now includes all the filters and fields for reports pages
- I updated the filings page:
  - No longer includes financial columns
  - No longer includes "report type" filters, since those are available on reports pages
  - Report rows on the filings table share the same details panel template as on the reports pages, so now Form 5s include a details panel
- I added a new "range filter" for financial ranges, and replaced all instances of min and max financial fields with this new range macro (but no styles yet)

This should be all ready to go. Finishing up a few things in a related PR in fec-style, but not a blocker.


openfecwebapp/templates/datatable.html
```diff
@@ -48,7 +47,14 @@
...
 var context = {
-  office: '{{ office }}'
+  office: '{{ office }}',
+};
+</script>
+{% endif %}
+{% if form_type %}
+<script>
+var context = {
+  formType: '{{ form_type }}',
```
ccostino: Is the idea to have both of these properties be in the `context` object if the conditions allow for it?  It doesn't look like that'll be the case here, and I wonder if there may be a possibility of both conditional checks being true.  If that were to happen, `context` would be overwritten and only contain what is in the last `if` statement here.

If this is not the case, the extra comma at the end might be an issue, I thought JS complained about that after the last item in a literal array or object definition.

openfecwebapp/templates/macros/filters/range.html
```diff
@@ -0,0 +1,16 @@
+{% macro amount(name, label) %}
+<fieldset class="filter" id="{{name}}-field">
+  <legend class="label">{{ label }}</legend>
+  <div class="range">
+    <div class="range__input js-filter">
+      <label>From</label>
```
ccostino: Note that you can add a `for` attribute in a `label` element and set it to match the `id` of the input element.  This helps with [accessibility](https://pages.18f.gov/accessibility/forms/) as it ties the label to the form field. :-)

noahmanger:
cc @LindsayYoung

openfecwebapp/templates/datatable.html
noahmanger: So yeah, that's the idea: the `office` variable is only being passed through on candidate office pages, and the `form_type` variable is only being passed through on reports pages. But that's a good point: it's definitely brittle and liable to cause problems if we ever end up trying to pass both. I'll think of a better way to do it.

noahmanger: Ok, came up with a solution for this.

noahmanger:
Ready for review. @xtine and @jontours can you check this out?

jontours:
@noahmanger I was able to test the web-app locally with my PR for this deployed locally, all looks good to me.

openfecwebapp/templates/partials/independent-expenditures-filter.html
```diff
@@ -30,11 +31,8 @@
 
   <button type="button" class="js-accordion-trigger accordion__button">Transaction information</button>
   <div class="accordion__content">
-  {# {% include 'partials/filters/ie-report-types.html' %} #}
-
-    {{ text.field('min_amount', 'Minimum expenditure amount', {'data-suffix': 'or more', 'data-inputmask': '"alias": "currency", "rightAlign": false, "clearMaskOnLostFocus": true' }) }}
-    {{ text.field('max_amount', 'Maximum expenditure amount', {'data-suffix': 'or less', 'data-inputmask': '"alias": "currency", "rightAlign": false, "clearMaskOnLostFocus": true'}) }}
-
+    {# {% include 'partials/filters/ie-report-types.html' %} #}
```
xtine: Why is this still commented?

xtine:
@noahmanger: just had a question about a comment just for curiosity but otherwise looks good. 👍

noahmanger:
Fixed! Thanks.

Issue #1342, opened 2016-07-07
Broken link to Candidate/pac/party views when clicked from new top raisers/spenders list


jontours:
Found on stage, it seems the url is just off, the page that is linked to is missing the "data" parameter.

noahmanger:
Gah. I'm on it.

noahmanger:
Resolved by #1343

PR #1343, opened 2016-07-07
Fixing broken top-list urls

noahmanger:
This uses the `url_for` function to build the correct URLs to candidate and committee pages from the top lists. 

h/t @jontours for catching the bug.

Resolves https://github.com/18F/openFEC-web-app/issues/1342

### DIFF
--- a/openfecwebapp/templates/macros/overviews.html
+++ b/openfecwebapp/templates/macros/overviews.html
@@ -48,11 +48,11 @@
   </div>
 {% endmacro %}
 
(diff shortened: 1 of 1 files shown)


ccostino:
Oh awesome, definitely good catch!

jontours:
It works locally for me.  It's not apparent how "data" gets prepended to the path, but since I noticed the url_for function works for other parts of the app, such as candidate search results.  This should work, gonna go ahead and merge.

PR #1374, opened 2016-07-18
Fix fundraising comparison on election page

### LINKED ISSUE #1355: Fundraising comparison (in 2016 presidential election) showing unexpected candidates as default
https://beta.fec.gov/data/elections/president/2016/

These candidates are supposed to be listed/shown by amount raised so that the most interesting candidates will show up by default. I was looking at something else, and noticed that I didn't recognize any of these candidates, and the ones I expected to be there are halfway down the list.

Is the data wrong here? Or did something in the default setting break?

screen shot 2016-07-13 at 7 53 25 pm

cc @noahmanger @LindsayYoung

ccostino:
Depends on https://github.com/18F/openFEC/pull/1825
Fixes #1355 

This changeset adds a missing flag to the query that is used for pulling the candidates for the fundraising comparison section.  It was missing the `sort_null_hide` parameter.

/cc @noahmanger, @jenniferthibault

### DIFF
--- a/static/js/pages/elections.js
+++ b/static/js/pages/elections.js
@@ -484,7 +484,7 @@ $(document).ready(function() {
     .value();
   var url = helpers.buildUrl(
     ['elections'],
-    _.extend(query, {per_page: 0})
+    _.extend(query, {per_page: 0, sort_hide_null: true})
   );
   $.getJSON(url).done(function(response) {
     $table.dataTable(_.extend({}, defaultOpts, {


noahmanger:
Does this depend on an API change? I tried running it but still seeing the same problem

ccostino:
Oh, yes, it does, updating description to reflect that.

noahmanger:
I see it works with the API change so I'm going to go ahead and merge this.

Issue #1363, opened 2016-07-15
Implement new text input filter loading and completion state


noahmanger:
In order for users to better understand what data is showing when, implement the new text input filter loading and completion states as [designed here](https://github.com/18F/openFEC-web-app/issues/1328) and [here](https://github.com/18F/fec-style/issues/416)

![image](https://cloud.githubusercontent.com/assets/11636908/16701502/23323fec-452e-11e6-9a53-5b93a8e6747b.png)

![image](https://cloud.githubusercontent.com/assets/1696495/16970981/c02002cc-4dd4-11e6-847a-487188f4a502.png)

**Completion criteria:**
- [x] All text inputs filters have buttons with the appropriate icons (hint: use the `.combo` pattern `_form-styles.scss`
- [x] On hitting enter, blurring the input, or clicking the button, the button shows the loading indicator for a set amount of time
- [x] Upon success, the message shows the amount of results added or subtracted

Note, partially dependent on some UX design work happening here https://github.com/18F/openFEC/issues/1828

xtine:
![vu6d7phxbu](https://cloud.githubusercontent.com/assets/24054/17167277/1c053f00-5391-11e6-94a6-fec3e981ed43.gif)

Issue #1362, opened 2016-07-15
Implement new checkbox loading & completion states


noahmanger:
In order for users to better understand what data is showing when, implement the new checkbox loading and completion states as [designed here](https://github.com/18F/openFEC-web-app/issues/1328).

**Criteria for completion:**
- [x] Upon selecting a checkbox, show a loading indicator for 250ms or until the data is loaded, whichever is longer
- [x] Upon success, outline the checkbox target area in green and show a check mark icon
- [x] Fade out the check mark icon after 5 seconds
- [x] Upon unchecking a checkbox, show the loading indicator
- [x] Upon success, show the check mark, but not the green outline
- [x] Upon success for both deselecting and selecting, show a message stating how many results were added or removed

This should be the same for:
- [x] All checkbox filters on data table pages
- [x] All checkbox filters on the election pages
- [x] All checkbox filters on the calendar page

noahmanger:
I made some progress on this, but handing off to @xtine  to finish up.

xtine:
![uqsxixn2iy](https://cloud.githubusercontent.com/assets/24054/17167220/cca0774a-5390-11e6-97c9-41d927997fc4.gif)

Issue #1327, opened 2016-07-01
Show number of applied filters in accordion heading


noahmanger:
So that users know that a grouping of filters has filters applied within it, show the number of active filters in the accordion header area.

Example:
![image](https://cloud.githubusercontent.com/assets/1696495/16510948/ba583b18-3eff-11e6-9c29-ef742b29f394.png)

**Criteria for completion:**
- [x] When a filter is applied, the number increases by 1
- [x] When a filter is removed, the number decreases by 1
- [x] If no filters are applied, no numbers are shown
- [x] If filters are applied on page load, the numbers are shown

**Question:**
In a case like this:
![image](https://cloud.githubusercontent.com/assets/1696495/16510973/06ee2ea6-3f00-11e6-889b-e1951fe45a16.png)

Do we say the number of filters applied is 2 (Democrat and Republican) or 1 (Political party)? I _think_ it's more useful to show 2, as you'll see the number increment with every addition, but wanted to check.

cc @jenniferthibault @onezerojeremy

jenniferthibault:
Yes, I'd also say 2. Same as the way I would expect the number to increase by each state you select.

onezerojeremy:
Agreed. 2! State, or committee, or whatever. 

I think there is a weird thing here, which i came across when i was digging into the ixd: I think it's best articulated in thinking about the checkboxes for `individual` vs `committee` contributions. Effectively, this state, the default, 
<img width="185" alt="screen shot 2016-07-01 at 1 08 50 pm" src="https://cloud.githubusercontent.com/assets/474916/16528794/041a51cc-3f8d-11e6-887b-a5cebcf4d3c8.png"> 

is equivalent to this 
<img width="185" alt="screen shot 2016-07-01 at 1 08 56 pm" src="https://cloud.githubusercontent.com/assets/474916/16528800/0e815b7e-3f8d-11e6-9ffc-c1d27c665968.png">

This feels weird to me, but I haven't figured out if it's a big problem, or how to fix it.

PR #1377, opened 2016-07-19
Removing loading state / setting success state

noahmanger:
This adds support for removing the loading class and adding success / error classes to filters. 

I was thinking about where the best place for these calls to live was and I think it's here rather than having event listeners on the individual filter elements, but would like @xtine 's thoughts.

example

Depends on https://github.com/18F/fec-style/pull/435

cc @xtine

### DIFF
--- a/static/js/modules/tables.js
+++ b/static/js/modules/tables.js
@@ -528,22 +528,32 @@ DataTable.prototype.fetchSuccess = function(resp) {
     }
 
     if (type === 'text') {
-      $label = $('.button--loading');
+      // typeahead
+      if ($(updateChangedEl).hasClass('tt-input')) {
+        $label = $(updateChangedEl);
 
-      if ($(updateChangedEl).val()) {
-        filterAction = '"' + $(updateChangedEl).val() + '" applied.';
+        filterAction = 'Filter applied.';
       }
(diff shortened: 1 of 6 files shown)


xtine:
So far interactions work for:
- checkboxes (#1362)
- radio buttons
- text input (#1363)

Typeahead inputs do not have the loading state shown in the button icon but still shows the interaction that the checkbox so it works halfway.

Also this includes the applied filters on accordion headings for #1327

static/js/modules/tables.js
```diff
@@ -230,10 +240,97 @@ function updateOnChange($form, api) {
...
+  if (updateChangedEl) {
+    var $label;
+    var type = $(updateChangedEl).attr('type');
+    var message = '';
+    var filterAction = '';
+    var $filterMessage = $('.filter__message');
+
+    if (type === 'checkbox' || type === 'radio') {
+      $label = $('label[for="' + updateChangedEl.id + '"]');
+      $('.is-successful').removeClass();
```
noahmanger: Do you need to specify `.removeClass('is-successful')`?

noahmanger:
This loks great. Just had one little question. And we just need to figure out why the tests are failing.

PR #1378, opened 2016-07-19
Filter currency

emileighoutlaw:
Hey @noahmanger and @xtine! 

I noticed that the filters for disbursements and operating expenditures weren't currency inputs. Like so:

disbursements filter

I was going to file an issue, but I thought I'd try to fix it myself (I want to be better at this kind of thing!) All that's to say I'm 50/50 on whether I did this perfectly or completely wrong. 

<3 ,
Emileigh


noahmanger:
I totally missed this. Nice work!

PR #1382, opened 2016-07-28
Feature/data landing mobile

noahmanger:
Fixes various things about the data landing page so it displays better on smaller screens:
- Decimal-padding of numbers now adjusts when the screen size adjusts.
- Removes the empty data periods since usability testing revealed them to be confusing and not helpful and it gives the labels more breathing room on mobile
- Positions the tooltip differently on small screens vs medium and larger. The tooltip is now full-width, but the point adjusts like so:
  gif

Requires https://github.com/18F/fec-style/pull/441
Resolves https://github.com/18F/openFEC/issues/1743

cc @xtine

### DIFF
--- a/static/js/modules/bar-charts.js
+++ b/static/js/modules/bar-charts.js
@@ -14,15 +14,16 @@ var stripes =
       '</defs>' +
   '</svg>';
 
-
 function GroupedBarChart(selector, data, index) {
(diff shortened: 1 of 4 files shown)


noahmanger:
This will also need to be merged into the release branch.

xtine:
👍

PR #1391, opened 2016-08-03
Filter interactions for native dropdowns

xtine:
Fixes lag loading of data from native filter dropdowns and also adds interactive messaging.

screen shot 2016-08-03 at 12 50 15 pm

/cc @noahmanger


xtine:
Supposed to be in release instead. Closing.

PR #1395, opened 2016-08-04
Filter dropdown and typeahead loading states

xtine:
Supports loading interactions for filter dropdowns and typeahead inputs.

Supports:
18F/fec-style#451 (filter dropdown interactions)
18F/fec-style#452 (filter typeahead loading states)

### DIFF
--- a/static/js/modules/tables.js
+++ b/static/js/modules/tables.js
@@ -311,7 +311,7 @@ function filterSuccessUpdates(changeCount) {
 
     // Clicking on "Clear all filters" will remove all dropdown checkboxes,
     // so we check to make sure the message isn't shown inside the dropdown panel.
-    if (!$label.parent().parent().hasClass('dropdown__list')) {
+    if ($label.closest('.dropdown__list').length < 1) {
       // append filter change count message after input
       $label.after($('<div class="filter__message filter__message--success">' + message + '</div>')
         .hide().fadeIn());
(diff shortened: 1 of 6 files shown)


static/js/modules/tables.js
```diff
@@ -320,10 +309,17 @@ function filterSuccessUpdates(changeCount) {
       clearTimeout(messageTimer);
     }
 
-    $label.after($('<div class="filter__message filter__message--success">' + message + '</div>')
-      .hide().fadeIn());
+    // Clicking on "Clear all filters" will remove all dropdown checkboxes,
+    // so we check to make sure the message isn't shown inside the dropdown panel.
+    if (!$label.parent().parent().hasClass('dropdown__list')) {
```
noahmanger: What about using `$label.closest('.dropdown__list').length < 1` to be a little more concise?

static/js/modules/tables.js
```diff
@@ -259,17 +259,10 @@ function filterSuccessUpdates(changeCount) {
     var filterAction = '';
     var $filterMessage = $('.filter__message');
 
+    $('.is-successful').removeClass('is-successful');
+
     if (type === 'checkbox' || type === 'radio') {
       $label = $('label[for="' + updateChangedEl.id + '"]');
-      $('.is-successful').removeClass();
```
noahmanger: Where did all this stuff that was removed end up?

noahmanger: Where did all this stuff that was removed end up?

xtine: I just combined logic together and it seems more straightforward.

PR #1504, opened 2016-08-10
Move feature flag to credential service

adborden:
- Moves the feature flag to the credentials service, as per https://github.com/18F/fec-cms/pull/401
- Simplifies the feature flag variable using `config.features` directly in templates


annalee:
:shipit:

Issue #1361, opened 2016-07-15
Implement new cumulative line chart on data landing page


noahmanger:
So that users can more clearly understand big-picture trends for raising and spending, implement the new cumulative line chart on the data landing page.

This iteration is meant to address problems usability testing uncovered with the current bar charts:
- Users are confused by the irregular time scale (6 month periods for one year, quarters for the next)
- Users aren't easily discovering the tooltips

The final design details are still being worked out. 

**Criteria for completion:**
- [Pending final designs]

PR #1508, opened 2016-08-11
Feature/new line chart

noahmanger:
Implements the new chart for showing cumulative amount raised and spent. 
- Creates a new `LineChart` component that builds a chart with a line for each entity type
- The cursor line moves with the mouse on hover and can be controlled via the arrow buttons
- Moving the cursor changes the totals to the right
- The totals represent the amount of money for the ending coverage date *up to* the first of the month (excluding the money reported *on* the first)

demo

## Some notes:
After I got data from @PaulClark2 , I learned that given the very different rules that apply to all the committees and the irregular primary schedule, there's actually reports for all committees in just about every month, which allows us to show data points for every month. But the amount reported on any given date varies wildly depending on how prevalent of a filing deadline it is; this would be very confusing if we were showing the amount reported *in* that period, because some would be drastically different than others. But because we're showing the *cumulative* amount raised until that point, I think it works well.

I opted to have the cursor move on hover largely because it was the simplest solution, but if we want to have it move on click or drag, we can do that. As is, on mobile the cursor moves on tap (it should eventually be on drag).

I also haven't gotten to re-implementing everything else around this — the new methodology buttons and the accordions. I think it might make sense to follow up with that separately next sprint.

## Remaining to-do:
- [ ] Add "other spenders" back in
- [ ] Add spending data
- [ ] Clean up CSS and JS

Requires https://github.com/18F/fec-style/pull/460
Resolves https://github.com/18F/openFEC-web-app/issues/1361


static/js/modules/line-chart.js
```diff
@@ -0,0 +1,243 @@
...
+
+  var minYear = d3.min(data, function(d) {
+    return d.date.getFullYear();
+  });
+
+  var maxYear = minYear + 1;
+
+  // Build the scales
+  var x = d3.time.scale()
+    // .domain(d3.extent(data, function(d) {
```
xtine: what did this do?

static/js/modules/line-chart.js
```diff
@@ -0,0 +1,243 @@
...
+        return parseMY(d);
+      } else if (d.getMonth() % 4 === 0) {
+        return parseM(d);
+      } else {
+        return '';
+      }
+    };
+  }
+
+    axis.ticks(d3.time.month)
```
xtine: spacing

xtine:
When I click on "Methodology," the modal appears but the dimensions are bigger than my screen:
screen shot 2016-08-11 at 9 00 20 pm

static/js/modules/line-chart.js
noahmanger: Oops, I meant to remove that. It was defining the min and max of the x-scale by the min and max dates in the data set, but I think it's better to show the full two years. I'll remove it.

static/js/modules/line-chart.js
```diff
@@ -0,0 +1,240 @@
...
+        date: d.date,
+        amount: d[type]
+      };
+    });
+
+    var entityMax = d3.max(totals, function(d) {
+      return d.amount;
+    });
+
+    max = entityMax > max ? entityMax : max;
```
adborden: Can you use `Math.max` here?

noahmanger: Bloops. Yeah. 😎

xtine:
This is so great. One suggestion though: is it possible to set `overflow: hidden` on body when the modal is triggered? Modals look great on my screen but when I try to scroll the content of the modal the body underneath also scrolls.

adborden:
Oops, sorry, I had an old page open, just missed @xtine's comment.

noahmanger:
Oh oops. I just pushed a commit with `Math.max` and disabling scrolling on `body` when it's opened. I need to make a new PR into feature anyway.

PR #1511, opened 2016-08-12
adding indentation and sorting docs by date

anthonygarvan:
This addresses two of the four comments @porta-antiporta made concerning AOs. It sorts by date and adds a small indentation. It does not add the expand / collapse feature, which @jenniferthibault agreed was not necessary for this release considering other architectural compromises made (ie, all docs are known to be relevant to search and are limited in number), and it does not remove the right padding of the `Summary` paragraph, since that would require significant refactoring of the `cal-list__title` class in fec-style.


anthonygarvan:
screen shot 2016-08-12 at 9 31 36 am

jenniferthibault:
👍 on the baby indent! This will give us a good enough nudge to test from

PR #1512, opened 2016-08-12
adding indentation and sorting docs by date

anthonygarvan:
This adds a "baby indent" and sorts by reverse chronological order.

### DIFF
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -84,7 +84,9 @@ def load_legal_search_results(query, query_type='all', offset=0, limit=20):
             else:
                 grouped_aos[ao['no']] = [ao]
 
-        results['advisory_opinions'] = grouped_aos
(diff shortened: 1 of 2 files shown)


adborden:
A few things about the alignment
- the second AO section is indented too much
- the "Final Opinion" text should align with the AO heading text

screenshot from 2016-08-11 22-37-09

We should be re-using the styles from the [calendar page](https://beta.fec.gov/calendar/) if possible (and file a ticket to rename those styles to better describe the list component they affect).

I think we should avoid using a `table` here since semantically it's not really a table anymore.

adborden:
@jenniferthibault can you take a look at this? I think it's okay to go without the calendar styles and we can clean it up next release.

anthonygarvan:
Good catch on the indentation of the next section! Just fixed it. I believe the Final Opinion text should _not_ align with the AO text. My understanding is all documents should be indented.

PR #1516, opened 2016-08-12
Changing class from data-container__body to data-container__table

noahmanger:
Fixes padding on data-table pages vs legal resources search pages. Data tables are now wrapped in a `.data-container__table` class which only has padding on the left (and 0 padding on small screens). Search results have `.data-container__body` which has padding on both sides (including mobile).

cc @adborden


noahmanger:
Meant to target release.

PR #1522, opened 2016-08-15
Feature/table-switcher (efiling front-end)

noahmanger:
Begins the work on the table-switching component required to show efiling data:

demo

The way it works:
- For pages that have efiling data, init a new `TableSwitcher()` component right after initializing the data table
- The `TableSwitcher` takes options for the processed table and efiling table, and passes these through to the `DataTable` component on `change`
- The `DataTable` uses those options to change the API url the table is using and also tells the `FilterSet` to disable / enable any filters

The `TableSwitcher` is also initialized as a `ToggleFilter`, a new subclass of filters, so that it can make use of the built-in code for adding and subtracting tags, etc.

## Remaining to do:
In order to get the tables working right, I need some (hopefully small) API changes. If those aren't possible, we can look into front-end hacks, but they make more sense on the API end. 

From a front-end standpoint, though, here's some things I know we need:
- [x] Verify which filters will be disabled
- [x] Write filter disable code
- [x] Write tag disable code
- [x] Ensure that it loads efilings if it is in the query
- [x] Add explanatory message / tooltip to the filter


xtine:
so far so good, 👍

xtine:
note: don't merge until efiling text is updated :)

noahmanger:
Ok, this has been updated, but still pending language. 

Some changes:
- Instead of the `message` box below the filter, I'm sticking to our pattern and using a tooltip for the message. This required a small tweak to fec-style: https://github.com/18F/fec-style/pull/474
- In https://github.com/18F/openFEC-web-app/pull/1522/commits/a343e30e430a3c7dbbc36e6b83e952090a5250c7 I removed some of the complicated sorting logic I had put in place as a workaround for using two different date columns for the different tables. The new code is a lot simpler, but requires an API change now that we have timestamp data: https://github.com/18F/openFEC/pull/1890

I'd say go ahead and review @xtine and then we can merge once the language is ready.

static/js/modules/table-switcher.js
```diff
@@ -0,0 +1,29 @@
...
+
+function TableSwitcher(control, opts) {
+  this.$control = $(control);
+  this.opts = opts;
+
+  this.$control.on('change', this.handleChange.bind(this));
+}
+
+TableSwitcher.prototype.init = function() {
+  var opts = this.opts[this.$control.find('input:checked').val()];
```
xtine: Can you break out the array index into its own variable for readability?

noahmanger: 👍

noahmanger:
Closing in favor of https://github.com/18F/openFEC-web-app/pull/1528

PR #1524, opened 2016-08-15
updating URL for Final Opinion

porta-antiporta:
The current AO canonical page has a "Final Opinion" button that does not pull up the Final Opinion.
This PR addresses that and the edge cases where there is no Final Opinion.

### DIFF
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -93,10 +93,6 @@ def load_legal_advisory_opinion(ao_no):
     
     if not canonical_document:
         return None
-
-    # manipulating category for display purposes
-    canonical_document['category'] = canonical_document['category'].lower()
-    canonical_document['category'] = canonical_document['category'].capitalize()
     
     advisory_opinion = {
         'no' : ao_no,
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -88,8 +88,15 @@ def load_legal_advisory_opinion(ao_no):
 
     for document in documents:
         canonical_document = document
-        if (document['category'] == 'Final Opinion'):
+        if document['category'] == 'Final Opinion':
             break        
+    
+    if not canonical_document:
+        return None
+
+    # manipulating category for display purposes
+    canonical_document['category'] = canonical_document['category'].lower()
+    canonical_document['category'] = canonical_document['category'].capitalize()
     
     advisory_opinion = {
         'no' : ao_no,
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -86,7 +86,11 @@ def load_legal_advisory_opinion(ao_no):
     if not (documents and len(documents)):
         return None
 
-    canonical_document = documents[0]
+    for document in documents:
+        canonical_document = document
+        if (document['category'] == 'Final Opinion'):
+            break        
+    
     advisory_opinion = {
         'no' : ao_no,
         'date': canonical_document['date'],
@@ -94,6 +98,7 @@ def load_legal_advisory_opinion(ao_no):
         'summary': canonical_document['summary'],
         'description': canonical_document['description'],
         'url': canonical_document['url'],
+        'category' : canonical_document['category'],
         'documents': documents,
         'entities': [],
     }
--- a/openfecwebapp/templates/legal-advisory-opinion.html
+++ b/openfecwebapp/templates/legal-advisory-opinion.html
@@ -33,7 +33,7 @@
     <section class="content__section">
       <h2 class="t-ruled--bold t-ruled--bottom">Documents</h2>
       <div class="content__section">
-        <a class="button button--cta button--document button--lg" href="{{ advisory_opinion.url }}"> View {{ advisory_opinion.category }}</a>
+        <a class="button button--cta button--document button--lg" href="{{ advisory_opinion.url }}">{{ advisory_opinion.category }}</a>
       </div>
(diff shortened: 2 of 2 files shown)


openfecwebapp/templates/legal-advisory-opinion.html
```diff
@@ -33,7 +33,7 @@ <h4 class="sidebar__title">Cited in this opinion</h4>
     <section class="content__section">
       <h2 class="t-ruled--bold t-ruled--bottom">Documents</h2>
       <div class="content__section">
-        <a class="button button--cta button--document button--lg" href="{{ advisory_opinion.url }}">Final opinion</a>
+        <a class="button button--cta button--document button--lg" href="{{ advisory_opinion.url }}"> View {{ advisory_opinion.category }}</a>
```
adborden: Let's not do this because the copy is going to be wrong "Final Opinion" vs "Final opinion".

openfecwebapp/api_caller.py
```diff
@@ -86,14 +86,19 @@ def load_legal_advisory_opinion(ao_no):
     if not (documents and len(documents)):
         return None
 
-    canonical_document = documents[0]
+    for document in documents:
+        canonical_document = document
+        if (document['category'] == 'Final Opinion'):
```
adborden: Not a big deal, but you don't need the parenthesis for `if` blocks.

openfecwebapp/templates/legal-advisory-opinion.html
porta-antiporta: I also added "View" (to match the mockup).  Should I just send the category to all lower case? `View final opinion` or `View withdrawl of request`?

openfecwebapp/api_caller.py
```diff
@@ -86,14 +86,19 @@ def load_legal_advisory_opinion(ao_no):
...
+            break        
+
```
ccostino: In the case of none of the documents matching `Final Opinion`, going with the category of the final document will suffice?

Also,  can you guarantee the case here?  Would it be better to call `lower()` on the `document['category']` value and compare it to `final opinion`?

openfecwebapp/api_caller.py
anthonygarvan: Won't this throw an error if no final opinion is present? I think then canonical_document would be undefined and when you try to index it below it would thrown an error.

adborden: On Mon, 2016-08-15 at 11:02 -0700, Anthony Garvan <notifications@github.com> wrote:
> > @@ -86,14 +86,19 @@ def load_legal_advisory_opinion(ao_no):
> >      if not (documents and len(documents)):
> >          return None
> >  
> > -    canonical_document = documents[0]
> > +    for document in documents:
> > +        canonical_document = document
> > +        if (document['category'] == 'Final Opinion'):
> 
> Won't this throw an error if no final opinion is present? I think then canonical_document would be undefined and when you try to index it below it would thrown an error.

No error, it would just choose the last document in the set as the
Canonical document, so it works. But this would be nice to explicitly
test for.

--
Aaron D Borden
Human and Hacker

openfecwebapp/api_caller.py
anthonygarvan: yes, we can guarantee case here.

openfecwebapp/templates/legal-advisory-opinion.html
porta-antiporta: After discussing with @emileighoutlaw we are taking out "View" and only capitalizing the first letter.

porta-antiporta: I put it in api_caller mostly because i don't know how to do string manipulations in HTML file.

openfecwebapp/api_caller.py
porta-antiporta: Also to the other question @ccostino, yes if there is no `Final Opinion` document, we will are purposedly ending up on the last document in the array. (Hence the changes to use `category` as the button label in `legal-advisory-opinion.html`

openfecwebapp/api_caller.py
porta-antiporta: Added test.  It should never execute as there's a test for the `documents` array already;  but better safe than sorry.

openfecwebapp/api_caller.py
```diff
@@ -86,14 +86,26 @@ def load_legal_advisory_opinion(ao_no):
     if not (documents and len(documents)):
         return None
 
-    canonical_document = documents[0]
+    for document in documents:
+        canonical_document = document
+        if document['category'] == 'Final Opinion':
+            break        
+    
+    if not canonical_document:
```
adborden: This condition should never happen though...

openfecwebapp/api_caller.py
```diff
@@ -88,9 +88,16 @@ def load_legal_advisory_opinion(ao_no):
...
-        if (document['category'] == 'Final Opinion'):
+        if document['category'] == 'Final Opinion':
             break        
     
+    if not canonical_document:
+        return None
+
+    # manipulating category for display purposes
+    canonical_document['category'] = canonical_document['category'].lower()
+    canonical_document['category'] = canonical_document['category'].capitalize()
```
adborden: Since these are display purposes, it doesn't make sense to change them at the data structure level. These should be done in the view.

openfecwebapp/api_caller.py
porta-antiporta: Yep, should never. Just put it there in the event a change breaks:

`if not (documents and len(documents)):`

PR #1527, opened 2016-08-17
Use document specific count

adborden:
In case the API returns a different count for `total_all` vs `total_%(result_type)s`, go for the safe bet and use the document specific count.


anthonygarvan:
looks good 🎉

PR #1531, opened 2016-08-17
Search legal statutes

adborden:
Implementing search for legal statutes https://github.com/18F/fec-eregs/issues/210
- Add some tests around the legal search endpoints
- Refactor the legal search templates
- Removes the legal feature flag, introduces a statutes feature flag

Depends on https://github.com/18F/fec-style/pull/480


openfecwebapp/config.py
```diff
@@ -28,7 +28,7 @@
 )
 
 features = {
-    'legal': bool(env.get_credential('FEC_FEATURE_LEGAL', '')),
+    'legal_statutes': bool(env.get_credential('FEC_FEATURE_LEGAL_STATUTES', '')),
```
porta-antiporta: Would we want to just use the same flag name `features.legal`?  Minimizes updates to our readme docs.

adborden: @porta-antiporta mentioned in stand-up that we should leave the code as-is for now, with the new feature flag.

porta-antiporta: 👍 i would just add a request we update the readme to point to config.py to see what flags are available? (rather than updating the readme everytime we have a deprecate/add a flag?)

annalee:
nice work @adborden! :shipit:

openfecwebapp/config.py
adborden: I agree that we shouldn't upgrade the README everytime we change the feature flags but the README already links to `config.py` https://github.com/18F/openFEC-web-app#features

porta-antiporta: :homer-doh:

Issue #1397, opened 2016-08-05
Refine loading & success interactions


noahmanger:
Some follow up tasks to polish up the filter loading, success and error states from our meeting just now:
- [x] Remove error states on success
- [x] Fix overlapping `.filter__remove` button and error x’s
- [x] Change open state on dropdown button to up-pointing arrow
- [x] Change check mark in selected dropdown items to gray
- [x] Fix the occassional extra loading state that is triggered when losing focus on a text input
- [x] Move the "unique only" tooltip to be in the label, rather than next to the checkbox
- [x] Typeahead generated checkboxes should show loading and success indicators when checked / unchecked
- [x] Ensure that typeahead-generated checkboxes trigger a filter count increment / de-increment in the accordion header and in the filter panel

And some content-specific tasks:
- [x] Update the language in the success messages (tweaking "adding" / "removed" language)
- [x] The success message when there's 0 results added or removed should have special language
- [x] Add error message

xtine:
@noahmanger @jenniferthibault @onezerojeremy: while refining these interactions, I noticed that typeahead generated checkboxes can't be removed like dropdown generated checkboxes can be. What do you think about adding the ability to remove them as well?

onezerojeremy:
i think its a great idea! good catch!

noahmanger:
@emileighoutlaw do you have time to work on the few language tasks here?

emileighoutlaw:
The content pattern here is:
- "Added" and "removed" for filters, because that's the action the user takes. 
- "Expanded" and "narrowed" for results, which feels more technically accurate. "Expanded" isn't perfect, but I believe it's understandable, and — importantly — it allows us to use "narrowed" as its opposite. "Narrowed" is the important piece here, because what we want to give the impression of honing in on results, not taking data away. The word we use for that moment is the most fraught with risk for misunderstanding.
- No punctuation on either line, which lightens the tone a little bit from the very serious period. I think the line break is a sufficient ending

## Increase in results messages
**Filter added**
Expanded by 22,201 results

**Filter removed**
Expanded by 22,201 results

## Decrease in results messages
**Filter added**
Narrowed by 22,201 results

**Filter removed**
Narrowed by 22,201 results

## No change messages
**Filter removed**
No change in results

**Filter added**
No change in results


I'd love to hear any thoughts on any of those options.


## Error messages
I'm posting a question here and not a solution (sorry!) because I let conference brain get in the way of my remembering what we decided last Friday. 

[...]

PR #1532, opened 2016-08-17
Update filter messaging

xtine:
Based on the discussions on #1397, update filter success messaging to more/less:

screen shot 2016-08-17 at 3 47 58 pm screen shot 2016-08-17 at 3 48 02 pm

Include a message after filter errors:

screen shot 2016-08-17 at 3 43 38 pm

Goes along with 18F/fec-style#478

cc: @noahmanger

### DIFF
--- a/static/js/modules/tables.js
+++ b/static/js/modules/tables.js
@@ -612,15 +612,18 @@ DataTable.prototype.fetchError = function() {
 
   $('.filter__message').remove();
 
-  $('.is-loading')
+  $('.filters .is-loading')
     .removeClass('is-loading')
     .addClass('is-unsuccessful')
     .after($('<div class="filter__message filter__message--error">' +
       '<strong>We had trouble processing your request</strong><br>' +
-      'Please try again. If you still have trouble let us know</div>'))
+      'Please try again. If you still have trouble ' +
+      '<span class="js-filter-feedback">let us know</a></div>'))
     .hide().fadeIn();
 
   self.$processing.hide();
+
+  clearTimeout(messageTimer);
 
   messageTimer = setTimeout(function() {
     $('.is-unsuccessful').removeClass('is-unsuccessful');
@@ -628,7 +631,7 @@ DataTable.prototype.fetchError = function() {
     $('.filter__message').fadeOut(function () {
       $(this).remove();
     });
-  }, helpers.LOADING_DELAY);
+  }, helpers.ERROR_DELAY);
 };
 
 /**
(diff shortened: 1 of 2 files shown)


noahmanger:
Awesome! I'm going to beat @emileighoutlaw to the grammar note though and say it should be "fewer" not "less".

xtine:
Derp! Fixed:
screen shot 2016-08-17 at 4 13 29 pm

static/js/modules/tables.js
```diff
@@ -605,14 +610,28 @@ DataTable.prototype.fetchSuccess = function(resp) {
...
-  }, helpers.LOADING_DELAY);
+  $('.filter__message').remove();
+
+  $('.filters .is-loading')
+    .removeClass('is-loading')
+    .addClass('is-unsuccessful')
+    .after($('<div class="filter__message filter__message--error">' +
+      '<strong>We had trouble processing your request</strong><br>' +
+      'Please try again. If you still have trouble ' +
+      '<span class="js-filter-feedback">let us know</a></div>'))
```
noahmanger: Inconsistent tag (`a` vs `span`). And I _think_ this should be a button, since it's just opening something and not taking you somewhere.

emileighoutlaw:
🙌 @noahmanger for the fewer/less catch 😁 

Also, if you are interested in grammar gossip: there are rumors floating around that the grammatical distinction between less and fewer is going to be taken out AP soon. (which would be fascinating, if true)

static/js/modules/tables.js
xtine: I waffled between `a` and `button`, but seeing as how most of the triggers on the site are button, I went ahead and made the change. Thanks for the 👀

static/js/modules/tables.js
```diff
@@ -274,7 +275,7 @@ function filterSuccessUpdates(changeCount) {
         // show message after generated checkbox (last item in list)
         $label = $('.js-typeahead-filter li').last();
 
-        filterAction = 'Filter applied.';
+        filterAction = 'Filter applied';
```
noahmanger: Should this also be "Filter added"?

static/js/modules/tables.js
```diff
@@ -604,15 +609,38 @@ DataTable.prototype.fetchSuccess = function(resp) {
...
+  }
+
+  $('button.is-loading').removeClass('is-loading');
+
+  self.$processing.hide();
 
-  setTimeout(function() {
+  clearTimeout(messageTimer);
+
+  messageTimer = setTimeout(function() {
```
noahmanger: I just tried this out and @onezerojeremy and I agree that error messages shouldn't automatically disappear. They should only go away after another fetch. I can't remember if we spec'd it out this way or not, but it definitely feels weird to have it go away.

xtine: Should the error messages stay for each input, or only to the last input that triggered an error?

noahmanger: Just the last one that triggered the error.

PR #1535, opened 2016-08-18
[MERGE WITH GIT FLOW] Init GA based on server env

adborden:


noahmanger:
LGTM. I included a tiny copy edit our partners requested so it could get in in the same hotfix. Will merge and deploy once the tests pass.

openfecwebapp/templates/layouts/main.html
```diff
@@ -16,6 +16,7 @@
   {% endif %}
   {% if api_location %}
   <script>
+    ANALYTICS = {{ use_analytics|json }};
```
ccostino: One quick question here - why is the `|json` filter (I'm assuming filter since this is in the template) needed here?  Is this accounting for safe output?

noahmanger: Hm. It seems to work, but in the docs it says `tojson` . http://flask.pocoo.org/docs/0.11/templating/#standard-filters

ccostino: As @adborden found, it's our own custom filter: https://github.com/18F/openFEC-web-app/blob/3df7e7fad9ed1d2413a13270e026f619c7bdda40/openfecwebapp/filters.py#L33

adborden: Some discussion in the slack channel on this, but the net is
- We have a custom `json` template filter, we'll remove/replace it with `tojson` in the next sprint.
- `safe` isn't necessary but couldn't hurt as this is a boolean value there would never be anything for jinja to accidently escape.
- The `json` filter converts from python boolean `False` to javascript `false`

PR #1539, opened 2016-08-22
Adding support for efiling views of reports

noahmanger:
This adds support for toggling between efile data and processed data on the new reports pages. Here's what it does:
- Moves the efile toggle to a partial so filings filters and reports filters can share it
- Adds a column to reports tables for "receipt date"
- Initializes the `TableSwitcher` component for the reports tables

After taking a first pass at this I found a few inconsistencies between the end points that shouldn't be hard to change.


ccostino:
@noahmanger, is there a missing column heading or two here, or is this intentional?

screen shot 2016-08-25 at 3 41 34 pm

ccostino:
I should clarify, that above image is looking at the processed data - the eFiling one appears to have a similar header/missing column heading issue:

screen shot 2016-08-25 at 4 00 46 pm

Again, right side of the screen.

noahmanger:
Huh I'm not seeing it. I just made a change though, so try pulling down, running `npm run build` and checking.

ccostino:
Ah that might be why, needing to re-run `npm run build` - will do!

ccostino:
Okay that did it, thanks @noahmanger!

PR #1540, opened 2016-08-23
Feature/new text filter

noahmanger:
Supports the new `CheckboxFilter` and `TextFilter` . 

Individual types of filters are now designated with `data` attributes.

Replaces the individual contributor name `TypeaheadFilter` with a `TextFilter` now that they support multiple parameters.

Depends on https://github.com/18F/fec-style/pull/483


ccostino:
@noahmanger it looks like something caused a merge conflict here recently.

openfecwebapp/templates/macros/filters/range.html
```diff
@@ -1,21 +1,19 @@
...
-      <div class="range__hyphen">-</div>
-      <div class="range__input range__input--max">
-        <label for="max_{{name}}">To</label>
-        <input id="max_{{name}}" type="text" name="max_{{name}}" data-range="max" data-prefix="$" data-suffix="or less" tabindex="0">
-      </div>
-      <button class="button--go button--standard" type="button">
-        <span class="u-visually-hidden">Search</span>
-      </button>
+  <div class="range">
+    <div class="range__input range__input--min js-filter" data-filter="range">
```
xtine: Is there a reason why `.js-filter` is set on each range input individually? In 18F/fec-style#483, `this.$elm.find('button')` is undefined so currency filter submit isn't focusing correctly on enter press.

noahmanger: So I ran into issues because each instance of a `Filter` expects the same `name`. Because the `min` and `max` filters have different names, the URL wasn't being set properly. I might be missing something, but I _think_ this is right to keep them separate.

xtine: Gotcha. Once submit was fixed, it seemed to make everything else work even better. Great work!

xtine:
This looks ready to me! There still seems to be a conflict with `develop` before we can merge in.

noahmanger:
Resolved!

xtine:
👍!

PR #1544, opened 2016-08-24
Financial range filters

xtine:
Template and filter action message for new inline financial range styles.

Supports 18F/fec-style#484

### DIFF
--- a/openfecwebapp/templates/macros/filters/range.html
+++ b/openfecwebapp/templates/macros/filters/range.html
@@ -1,19 +1,21 @@
 {% macro amount(name, label) %}
 <fieldset class="filter" id="{{name}}-field">
   <legend class="label">{{ label }}</legend>
-  <div class="js-filter range">
-    <div class="range__input range__input--min">
-      <label for="min_{{name}}">From</label>
-      <input id="min_{{name}}" type="text" name="min_{{name}}" data-prefix="$" data-suffix="or more" tabindex="0">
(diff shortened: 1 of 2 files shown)


openfecwebapp/templates/macros/filters/range.html
```diff
@@ -3,23 +3,20 @@
   <legend class="label">{{ label }}</legend>
   <div class="range">
     <div class="js-filter">
-      <label for="min_{{name}}">From</label>
-      <div class="combo combo--filter--mini">
-        <input id="min_{{name}}" class="combo__input" type="text" name="min_{{name}}" data-inputmask="'alias': 'currency', 'rightAlign': false, 'clearMaskOnLostFocus': true" data-suffix="or more" tabindex="0">
-        <button class="combo__button button--go button--standard is-disabled" type="button">
-          <span class="u-visually-hidden">Search</span>
-        </button>
+      <div class="range__input range__input--min">
```
noahmanger: What about just adding the `.js-filter` class to this div? I don't think we need the extra one, and then `range__input` can be the class for styles.

noahmanger:
Just a couple little stylistic things on this and the fec-style one, and then I also caught this bug when I somehow triggered an error:

image

Other than that, looks great.

xtine:
Got error messages fixed:
screen shot 2016-08-24 at 10 38 56 pm

PR #1545, opened 2016-08-25
Bugfix/ao pagination

anthonygarvan:
Web app now consistently returns number of documents, even in case of AOs. Resolves [#211](https://github.com/18F/fec-eregs/issues/211). This PR also tests some of the grouping logic we implemented last sprint and refactors older tests to use `assert` keyword syntax.


annalee:
We paired on part of this and the tests look good! :shipit:

PR #1550, opened 2016-08-27
update statute style

porta-antiporta:
Makes the following changes
- fixes bug on Regulations style that wasn’t highlighting regulation name
- makes Statutes consistent with Regulations style

### DIFF
--- a/openfecwebapp/templates/partials/legal-search-results-statute.html
+++ b/openfecwebapp/templates/partials/legal-search-results-statute.html
@@ -1,18 +1,14 @@
 <div class="simple-table simple-table--responsive simple-table--display legal-search-results">
   <div class="simple-table__header">
-    <div class="simple-table__header-cell">
-      Name <span class="simple-table__header-subtext">| Hit</span>
-    </div>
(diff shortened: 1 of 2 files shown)


openfecwebapp/templates/partials/legal-search-results-statute.html
```diff
@@ -1,18 +1,14 @@
 <div class="simple-table simple-table--responsive simple-table--display legal-search-results">
```
porta-antiporta: @adborden couldn't find `legal-search-results` class in fec-style.  Not sure what this was for.

porta-antiporta:
fixes https://github.com/18F/fec-eregs/issues/239

openfecwebapp/templates/partials/legal-search-results-statute.html
adborden: Good catch, it's an old style and I think it can be removed.

adborden:
:+1: looks good, but I'm not going to hit the merge button until after the demo, but don't let that stop anyone else.

PR #1561, opened 2016-08-29
Adding the missing data attribute for the efiling filter

noahmanger:
Fixes the bug that was causing the broken filings and reports pages.


ccostino:
Thanks, @noahmanger!

Issue #1436, opened 2016-08-08
Known issue: Need remaining IE data


PaulClark2:
Currently missing a chunk of IEs by only having F3X
![image](https://cloud.githubusercontent.com/assets/11817734/17480621/bceb038c-5d46-11e6-8944-aaf416155971.png)

jwchumley:
Adding the non form 3x IEs is extremely important to entire group of content experts. Misleading without all of them.
Priority: High

noahmanger:
This fits in nicely with our "Add remaining data" epic, so I want to drill into what we need to do yet. 

As I understand it, there's a two remaining data sets we need to include
- Form 5s
- 24/48 hour reports

Correct?

And then there's two instances where this data is displayed:
- The aggregates that are shown on candidate, committee and election pages
- The itemized data itself on beta.fec.gov/data/independent-expenditures

When we were first launching the itemized IE data, we had a long discussion about what could and could not be shown together in the same view. So the good news is I think we have all the data already, which means it's just a matter of working out the interface decisions.

Here's an educated guess about what I think we're going for:

Form 5s: 
- Aggregates: Include alongside F3X aggregates
- Itemized: Include alongside F3X itemized data

24 and 48 hour reports:
- Do not include in aggregates, as it will be a double-count
- Can include on the IE data page, but it should be an either/or toggle, similar to efiling/processed, so that the records aren't interspersed with F3X records, which would result in duplicates.

Is that right? I know @PaulClark2 is out this week, so if @jwchumley or @amykort

jwchumley:
Pretty much yes.


There are some issues with form  5 filers that file 24/48 hour notices. They are supposed to follow up with a quarterly filing. IF they do then their summary figures are correct. IF they don't then their summary figures are incorrect or non existent.  We have some logic for this that we apply in very narrow circumstances but I think for now just including the notices and getting the form 5 summary data will be sufficient (Unless Paul has already talked to you about tracking quarterlies vs notices)


________________________________

noahmanger:
Ok, I have a working solution that looks something like this:

![image](https://cloud.githubusercontent.com/assets/1696495/18107171/cac54996-6ebb-11e6-991b-a220869ca2c3.png)

- This removes the previous filter that only showed `F3X` figures, so `F5` data is now included.
[...]

PR #1565, opened 2016-08-30
Adding data for form 5s and 24/48 hour notices to IEs

noahmanger:
- This removes the previous filter that only showed `F3X` figures, so `F5` data is now included.
- It adds a toggle filter for "Summary reports" vs "24 and 48 hour notices" (language needs to be changed, I'm sure). This filter defaults to show _only_ `F3X` and `F5` ("Summary reports") but can be toggled to show _only_ data from notices. 
- Aggregate links (e.g. from a total on a candidate page) default to only the "summary report" data

image

**Remaining work:**
- [x] @jwchumley verify the data is showing as expected
- [x] @LindsayYoung verify the data is showing as expected
- [x] Finalize language for the toggle (cc @emileighoutlaw )
- [x] Add a tooltip to the toggle if necessary (cc @emileighoutlaw )
- [x] Update the `ToggleFilter` component to output the correct text in the tag

Original issue discussion over here: https://github.com/18F/openFEC-web-app/issues/1436

Requires https://github.com/18F/fec-style/pull/530 for the tags


jontours:
@noahmanger in the interest of time and sanity I think I may move the 24/48 hour notices into a seperate endpoint.  Something like `/sched_e/notice` and then you can filter that endpoint down more based form type, i.e 'f3x', 'f5' etc.  Let me know if that is not workable.

noahmanger:
Oh yeah, that would totally work. We can just use the same front-end logic that we use for toggling between efilings and processed data.

noahmanger:
@jontours I just updated this so that the toggle toggles between `/schedules/schedule_e/` and `/schedules/schedule_e/notice` . But my local server is showing what looks to be some SQL errors. 

The big thing that I want to test and maybe you can just tell me: will the two endpoints have all the same filters? Will their responses be the same? Wondering if we'll need to account for that at all here.

jontours:
@noahmanger I did checkout this PR yesterday, I can pull the latest updates and see what's up.

noahmanger:
I just added a new `filing_form_type` filter that is enabled for "Summary reports" but disabled for notices:

image

image

noahmanger:
Ok @jontours I reverted this so it just uses a single endpoint. It uses the `is_notice` flag. Not sure if that's still the param though.

jontours:
Thanks @noahmanger, yes that is still the parameter used.

noahmanger:
Ok! This is all ready for review. I tested against the API branch and it's working as expected.

jontours (approved):
This looks good to go.  I noticed the descriptions were blank for notices, but after investigating through the API they some descriptions are null in the database.  I'll merge if it's ready to go.

noahmanger:
Let's do it. Thanks!

PR #1568, opened 2016-08-31
Fix/release fixes

noahmanger:
Fixes a couple things for the release:
- Fixes a missing closing parentheses that was causing /data/advanced/ to not load
- Comments out the link to /reports/ie-only/ because that table is still missing the committee name column (I chose to comment it out rather than remove it, but can remove it if anyone prefers)
- Adds missing `title` attributes to modal close buttons to make HTML code sniffer happy
- Ensures that the export button on reports data pages is always disabled


xtine:
👍

PR #1569, opened 2016-08-31
update copy to be U.S.C.

porta-antiporta:
Per https://18f.slack.com/archives/fec-partners/p1467229654000403 should be `U.S.C.` instead of USC.


ccostino:
Thanks, @porta-antiporta!

emileighoutlaw:
As a heads up, I'm starting to document these decisions in our growing FEC style guide. I'll do an audit pass for consistency when we hare in a less frantic sprint.

PR #1570, opened 2016-08-31
Fix/report-fixes

noahmanger:
Adds `document_description` to tables and details panels.

Adds a default character `--` to show up in the tables and detail panels if a given record does not have the associated field, which seems to happen on efiling due to the data not being processed. 

image

image

This is done in the `currency` handlebars helper by detecting if the value passed is not a number, and if so, returning `--`. This way, `0` will still show up as `$0`, but instead of a bunch of mysteriously blank fields, it's clearer that the field is just missing.

I know we've talked about the trickiness of displaying 0s vs. nulls, and I think this is a good solution, but @LindsayYoung let me know if you disagree.

### DIFF
--- a/static/templates/reports/candidate.hbs
+++ b/static/templates/reports/candidate.hbs
@@ -1,6 +1,6 @@
 <div class="panel__row">
   <div class="panel__heading">
-    <h4 class="panel__title">{{report_type_full}}
+    <h4 class="panel__title">{{document_description}}
     {{#if is_amended}}
       (This report has been amended)
     {{/if}}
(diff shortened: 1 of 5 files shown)


static/templates/reports/candidate.hbs
```diff
@@ -111,9 +111,6 @@
     {{#panelRow "Total loans received"}}
       {{{currency total_loans_received_period}}}
     {{/panelRow}}
-    {{#panelRow "Total loans received from candidates"}}
```
LindsayYoung: why are we removing this?

LindsayYoung: Also, I think the text should be candidate, instead of candidates.

noahmanger: I removed it because there's an inconsistency between house-senate reports and presidential reports in what the field is named, and I figured easier to just remove it for the time being since we haven't totally fleshed out what the rules are for what is included in these panels. But I can add it back if you think it's important.

PR #1571, opened 2016-08-31
Constrain document thumbnail grid to text-width

adborden:
Depends on https://github.com/18F/fec-style/pull/493

@jenniferthibault pointed out that the document thumbnail grid should not overflow the text width.
screenshot from 2016-08-31 14-47-49


jenniferthibault:
Thanks @adborden for keeping our grid pristine! 👌

PR #1572, opened 2016-09-01
Improved Date Filter Interactions

xtine:
Templates to support new date grid. Goes with 18F/fec-style#494


openfecwebapp/templates/macros/filters/date.html
```diff
@@ -47,11 +47,58 @@
...
+      <div class="range__input range__input--max" data-filter="range">
+        <label for="max_{{ name }}">Ending</label>
+        <input type="text" id="max_{{ name }}" name="max_{{ name }}" data-range="max" class="js-max-date" data-prefix="Ending" data-inputmask="'alias': 'mm/dd/yyyy'">
+      </div>
+      <button class="button--go button--standard" type="button">
+        <span class="u-visually-hidden">Search</span>
+      </button>
+    </div>
+    <div class="date-range-grid">
+      <div class="date-range-first">
```
noahmanger: Can we use BEM syntax? `.date-range--first`? (And for the others: `.date-range__months`, `.date-range__year`, etc.)

noahmanger: Maybe `.date-range-first` and `.date-range-second` should just be `.date-range` and select for them in JS using either data attributes or `nth-of-type`?

static/js/modules/tables.js
```diff
@@ -283,11 +283,16 @@ function filterSuccessUpdates(changeCount) {
...
         filterAction = 'Filter added';
       }
-      else if ($(updateChangedEl).parent().hasClass('range__input')) {
+      else if ($(updateChangedEl).closest('.range').hasClass('range--currency')) {
         $label = $(updateChangedEl).closest('.range');
 
         filterAction = 'Filter applied';
       }
+      else if ($(updateChangedEl).closest('.range').hasClass('range--date')) {
+        $label = $('.date-range-grid');
```
noahmanger: Let's scope this to the `updateChangedEl` in case there's multiple ones on a page.

openfecwebapp/templates/macros/filters/date.html
```diff
@@ -46,7 +46,6 @@
   </div>
   <div class="filter js-filter" id="{{ name }}" data-filter="date" data-name="{{ name }}" data-validate="true">
     <fieldset>
-    <legend class="label">{{ label }}</legend>
```
noahmanger: Removing the `<legend>` we also need to remove the `<fieldset>` which now that I think about it, shouldn't be necessary.

xtine: Now that I think about it, since it's a form, should we still keep the legend but hide the text for accessibility purposes?

noahmanger: I don't think we need the `legend` or `fieldset`. `fieldset` is just for grouping form elements that share the same name (like a group of checkboxes or radio buttons). So we don't need it here. And `legend` is only necessary to label a `fieldset`, so we don't need it either.

PR #1574, opened 2016-09-02
Adding printer-only logo

noahmanger:
Adds the print logo from https://github.com/18F/fec-style/pull/496

Because certain browsers don't render background images it has to be included as an `img`.


ccostino:
Thanks, @noahmanger!

PR #1575, opened 2016-09-03
Update reaction-box.html: spelling fix

busterroni:
Change "Uh oh. Something wennt wrong. Please try again." --> "Uh oh. Something went wrong. Please try again."

(removes second n from "wen**n**t")


porta-antiporta:
@busterroni thanks for the catch!

Issue #1466, opened 2016-08-08
Request: Link financial summary items to transactions


PaulClark2:
detailed summary page lines should link to transactions

jwchumley:
Priority: High

noahmanger:
I've got a working, interim concept going that I show over in #1588. Take a look and let me know if it might help this issue for the time being before we have the chance to a more thorough design.

noahmanger:
Per the discussion [here](https://github.com/18F/openFEC-web-app/pull/1588), the interim solution I tried out doesn't serve the need, so we'll revisit this when looking at the overall design of this section. Moving to the backlog.

noahmanger:
Closing in favor of https://github.com/18F/openFEC-web-app/issues/2125

PR #1588, opened 2016-09-16
Adding link to itemized receipts and disbursements from totals table

noahmanger:
This adds links to itemized disbursements and receipts from the committee financial summary tables:

image

image

This is an intermediate fix to address the need [originally raised in FEC feedback](https://github.com/18F/openFEC-web-app/issues/1466) to have links from these summary numbers to itemized transactions. We have several other issues addressing the larger design of this pattern, so this isn't meant as the ultimate solution, but hopefully it can ease some user pain in the short term.

I chose to present it as a separate link for a couple reasons:
- We don't have the ability to filter down to transactions for each of the sub-items in these tables
- Because the transactions will never add up to what was reported, presenting it as a separate link rather than linking from the actual dollar amount hopefully makes it slightly less confusing

I chose to just take care of this in between tasks. If this turns into a thornier issue requiring more discussion or some design time, we might need to punt. But on the chance that this resolves it, wanted to submit.

Paging @jwchumley to see what you think.

### DIFF
--- a/openfecwebapp/templates/macros/tables.html
+++ b/openfecwebapp/templates/macros/tables.html
@@ -1,7 +1,7 @@
 {% import 'macros/null.html' as null %}
 {% import 'macros/charts.html' as charts %}
 
-{% macro totals(header_name, header_value, data, report_year, header_description='') %}
+{% macro totals(header_name, header_value, data, report_year, transaction_type, id, year) %}
   <figure class="totals-table totals-table--charts js-chart-series chart-series--horizontal">
       {% if data|length %}
       <button tabindex="0" class="accordion__button accordion__button--spacious js-accordion-trigger">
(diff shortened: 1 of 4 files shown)


jwchumley:
Are the two tabs the links you are referring to?  I don't think I want to go with that. I'd rather wait until you link the transactions to the summary values.




Jeff

________________________________

noahmanger:
Each of those buttons links to the itemized data table page for each type. It's fine to wait, but it will just be a little bit until we're able to prioritize this feature.

jwchumley:
I see. I didn't recognize the buttons for what they were. I still don't like buttons as a permanent solution. @PaulClark2 what do you think about buttons instead of linking to numbers?

PaulClark2:
I'm not sure I understand what "buttons" you are referring to. The "all receipts"/"all disbursements" buttons? 

I agree that's not a solution that is very satisfying. If I'm want to know more about individual contributions, I don't want all receipts. I want individual contributions. 

I'd rather like to specific line numbers (transaction types for older data) and provide explanations about why the sum of the transactions might not equal the total reported on the summary line.

noahmanger:
Alright then! I'm going to close this and put this issue back in the backlog to be dealt with when we have more time to more holistically think through the design of these section.

PR #1592, opened 2016-09-21
Feature/reconcile h1s

noahmanger:
Adjusts classes and markup throughout to accommodate new header styles from https://github.com/18F/fec-style/pull/520


jenniferthibault:
@adborden surfacing the question @noahmanger left in https://github.com/18F/fec-style/issues/490#issuecomment-250344537 about which legal side templates you've started that this would impact. I can think of `legal-archived-mur.html` `legal-current-mur.html` and `legal-enforcement-landing.html` 

Are there other I'm missing?

openfecwebapp/templates/election-lookup.html
```diff
@@ -60,8 +60,8 @@ <h2 class="t-ruled--bottom t-ruled--bold">Search</h2>
     </div>
     <div class="content__section--extra">
       <div class="usa-width-one-half">
-        <div class="section__heading">
-          <h2 class="heading__title u-no-margin">Results:</h2>
+        <div class="heading--section">
```
noahmanger: Yes!

noahmanger:
Ok, addressed those issues, and I think I got all the places in legal templates that are currently in there, including the MUR pages.

jenniferthibault:
Quick check on next steps. Seem like these require a check by Aaron before any merging, is that right?

noahmanger:
Yeah, that would be good.

PR #1596, opened 2016-09-22
Feature/current mur canonical page

anthonygarvan:
Implemented modified design for canonical mur page.


adborden:
cc @jenniferthibault

openfecwebapp/templates/legal-current-mur.html
```diff
@@ -0,0 +1,109 @@
...
+        <ul class="sidebar__content">
+          <li class="side-nav__item"><a class="side-nav__link" href="#summary">Summary</a></li>
+          <li class="side-nav__item"><a class="side-nav__link" href="#documents">Documents</a></li>
+          <li class="side-nav__item"><a class="side-nav__link" href="#participants">Participants</a></li>
+        </ul>
+      </nav>
+  </div>
+  <div class="main__content--right">
+    <section id="summary" class="content__section">
+    <div class="main__content">
```
adborden: I don't think you need this class, it saves space on the right for the citations sidebar, which the current MUR page doesn't have.

openfecwebapp/templates/legal-current-mur.html
```diff
@@ -0,0 +1,109 @@
...
+                </td>
+              </tr>
+            {% endfor %}
+          {% endfor %}
+        {% endif %}
+      </tbody>
+    </table>
+    <div class="results-info results-info--bottom">
+ +        <a class="legal-top" href="#">Top of page</a>
+ +  </div>
```
adborden: Hmm, I don't think you meant to have these `+` signs here.

openfecwebapp/views.py
```diff
@@ -52,10 +52,16 @@ def render_legal_advisory_opinion(advisory_opinion):
...
+    if mur['mur_type'] == 'archived':
+        return render_template(
+            'legal-archived-mur.html',
+            mur=mur,
+        )
+    else:
+        return render_template(
+            'legal-current-mur.html',
+            mur=mur,
+        )
```
adborden: Ah, looking at this again, we _could_ simplify this to
```python
return render_template(
  'legal-%s-mur.html' % mur['mur_type'],
  mur=mur,
)
```

adborden:
:+1: This looks great, some minor edits. We'll just have to make sure to follow up before turning on the feature flag.

adborden:
Ah, that sentence was incomplete, I meant, we should make sure to follow up **with @jenniferthibault** before turning on the feature flag.

openfecwebapp/templates/legal-current-mur.html
jenniferthibault: And since there is no citations sidebar, the Summary div and underline should extend all the way to the container, like the other sections.

jenniferthibault:
Thanks for the opportunity to double-check. There's a couple'a things I noticed:

- the Document and Participant tables should use alternating background colors set by the left-most row. (Set up and explained to be by @adborden here: https://github.com/18F/fec-style/pull/510#pullrequestreview-376278 )
- Like we do in the citations sidebar, we should use the complete citation title in the column: For statutes, `30121` becomes `52 U.S.C. 30121` and for revulations, the format would be `11 C.F.R. 110.4(c)(2)`
 ( I think I have that right, but if I'm off please correct me!)

PR #1586, opened 2016-09-15
Canonical MUR page (archive MURs)

adborden:
Adds view for the canonical MUR page (for archived MURs). Resolves https://github.com/18F/fec-eregs/issues/227
- [ ] https://github.com/18F/fec-style/pull/510

@jenniferthibault since both archived and current MURs will have a name field, I moved the archive case indicator under the summary, but not really sure where it should go.

screenshot from 2016-09-15 14-37-23

PR #1597, opened 2016-09-22
Canonical MUR styling

jenniferthibault:
This is an addendum PR to @adborden's https://github.com/18F/openFEC-web-app/pull/1586

I have done at least one possibly controversial thing, so it definitely needs @adborden 's approval/review.

The changes here focus on the sidebar content. 

**Previously it looked like this:**
screen shot 2016-09-22 at 4 38 04 pm

Which I had two fusses with:
- It was hard to tell where one link ended and the other began
- the "Statutes"  and "Regulations" as `h5`s were really competing with the "Cited in this case" for reading order, and the visual hierarchy felt backwards

The first could be addressed by using the same class that's being used for [Registration & Reporting](https://beta.fec.gov/registration-and-reporting/essentials-political-party-committees/) sidebar lists, `class="u-padding--bottom"`

The second, I've addressed through much less graceful means. I turned the `h` into a `p` and added an existing emboldening class`class="t-bold"` This makes it look exactly like I want it to look, but .... is that ok to do? 😬 

**Happy (put possibly verboten) end state:**
screen shot 2016-09-22 at 4 53 35 pm


adborden:
:+1: I think this pattern is something we want to incorporate into `fec-style`. My understanding is that all headings within the sidebar should follow this. The default `h*` styles are never what we want in the sidebar. @jenniferthibault do you want to add some classes for this pattern or should I follow up with a second PR?

jenniferthibault:
That is true, all headings within the sidebar would follow this. Would that mean I make an ....`h7` or something? Or are you talking about a different type of class?

adborden:
A CSS class, like `side-bar__heading`.
```html
<div class="side-bar">
  <div class="side-bar__title">Citations</div>
  <div class="side-bar__content">

    <div class="side-bar__heading">Statutes</div>
    ...
  </div>
</div>
```

It's okay, i can follow up in another PR. It requires a change to `fec-style`.

adborden:
Merging this, HoundCI seems stuck.

PR #1604, opened 2016-09-27
Enforcement landing page

adborden:
Part of https://github.com/18F/fec-eregs/issues/224

Contains draft content from https://github.com/18F/fec-eregs/issues/233

Removes the filter panel on legal landing pages from https://github.com/18F/fec-eregs/issues/242
- [x] Depends on https://github.com/18F/fec-style/pull/523

### DIFF
--- a/openfecwebapp/templates/legal-enforcement-landing.html
+++ b/openfecwebapp/templates/legal-enforcement-landing.html
@@ -1,0 +1,43 @@
+{% extends "layouts/legal-doc-landing.html" %}
+{% import 'macros/document.html' as document %}
+
+{% block content %}
+<div class="section__heading">
+  <h1 class="heading__title">Enforcement matters</h1>
+</div>
+<div class="content__section">
+  <p>Enforcement.</p>
+</div>
+<div class="content__section">
+  <div class="grid grid--3-wide">
+    <div class="grid__item">
+      <h5><a href="">Administrative Fine Program</a></h5>
+      <p>Procedure and penalties for late and missed filings.</p>
+    </div>
+    <div class="grid__item">
+      <h5><a href="">Alternative Dispute Program</a></h5>
+      <p>Procedure and penalties for late and missed filings.</p>
+    </div>
+    <div class="grid__item">
+      <h5><a href="">Matters Under Review</a></h5>
+      <p>Procedure and penalties for late and missed filings.</p>
+    </div>
+    <div class="grid__item">
+      <h5><a href="">Alternative Dispute Program</a></h5>
+      <p>Procedure and penalties for late and missed filings.</p>
+    </div>
+    <div class="grid__item">
+      <h5><a href="">Matters Under Review</a></h5>
+      <p>Procedure and penalties for late and missed filings.</p>
+    </div>
+    <div class="grid__item">
+      <h5><a href="">Alternative Dispute Program</a></h5>
+      <p>Procedure and penalties for late and missed filings.</p>
+    </div>
+    <div class="grid__item">
+      <h5><a href="">Matters Under Review</a></h5>
+      <p>Procedure and penalties for late and missed filings.</p>
+    </div>
+  </div>
+</div>
+{% endblock %}
--- a/openfecwebapp/templates/legal-enforcement-landing.html
+++ b/openfecwebapp/templates/legal-enforcement-landing.html
@@ -1,9 +1,9 @@
 {% extends "layouts/legal-doc-landing.html" %}
 
 {% block content %}
-<div class="section__heading">
-  <h1 class="heading__title">Enforcement matters</h1>
-</div>
+<header class="heading--main">
+  <h1>Enforcement matters</h1>
+</header>
 <div class="content__section">
   <p>
     The FEC has exclusive jurisdiction over the civil enforcement the federal campaign finance law. 
@@ -43,9 +43,9 @@
   </div>
 </div>
 <div class="content__section">
-  <div class="heading--section">
-    <h2 class="t-ruled--bottom">Enforcement programs</h2>
-  </div>
+  <header class="heading--section">
+    <h2>Enforcement programs</h2>
+  </header>
   <h4>Matters Under Review (MURs)</h4>
   <p>
     FEC enforcement matters (called MURs) must, by law, remain confidential until they are closed. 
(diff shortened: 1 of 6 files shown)


adborden:
cc @jenniferthibault

jenniferthibault:
Looking here: http://127.0.0.1:3000/legal/enforcement-matters/

(Ignore the floating feedback button, that's an artifact of my screen cap tool!)

screencapture-127-0-0-1-3000-legal-enforcement-matters-1475012039070




There is so much "coming soon" to this page that I'm hesitant to offer design improvements yet. 

If it were shipping as is, the big issue to my 👀 is really awkward spacing, which I have some suggested fixes for. Since those are contingent to the exact copy of the page, I'd rather hold on them until final copy is in place, then resolve them. Hoping to not make design fixes twice.

Is that ok with you @adborden ?

adborden:
Yep, I agree. We'll do a more thorough design review when I update the content. The search bar is looking okay, though?

jenniferthibault:
I do have a few thoughts for the search bar, but they might fall under `fec-style` 😺 

I think `search--mini` should probably be used in cases where it's in a narrow column, like the filter panel. It feels proportionally wrong in a wide space like this. I nudged it back to `search--large` which feels _more_ proportionally correct, but I think I was looking for something with an in-between height, which we don't have. Is it worth making a `search--medium` ?

The other thing is the Example text below the search bar has an existing class (that I added) that captures the correct margin-top, but applies a less-useful margin-left. It seems like the margin-left is a specific use case for the `search--large` class when there is also a dropdown, like on the data landing page: https://beta.fec.gov/data/

But more often, we want to use that class without the margin-left. It seems like we should refactor that pattern to have no margin left as the base unit, then add a class/something to give it the left space on the data landing page instance. 

Here's the screenshot of both of those applied, but I know we'll probably want to retweak.

screen shot 2016-09-28 at 9 59 46 am


Am I out of scope/talking crazy? Let me know what you think

adborden:
Yep, makes sense! I'll follow up with the changes. Should the search bar be less than full-width like in your screenshot? Trying to stick the uswds grid, so if you have a specific width in mind, I'll use that.

jenniferthibault:
Nope, I don't have a specific width in mind, and yep, it's ok for it to not be full-width. 

For height on the medium sized search bar, could we try 4.5rem? This would also be the pattern we'd use for search bars on the other legal landing page templates that we talked about adjusting in https://github.com/18F/fec-eregs/issues/242. 

cc @noahmanger into this for visibility into pattern tweaks 👀

adborden:
@noahmanger updated this with the new heading styles.

adborden:
@jenniferthibault here is the screen capture.
screencapture-localhost-3000-legal-enforcement-1475776041317

adborden:
And fixed the spacing between the MURs search and the "Administrative Fine Program" header.

openfecwebapp/templates/legal-enforcement-landing.html
```diff
@@ -0,0 +1,95 @@
+{% extends "layouts/legal-doc-landing.html" %}
+
+{% block content %}
+<header class="heading--main">
+  <h1>Enforcement matters</h1>
+</header>
+<div class="content__section">
+  <p>
+    The FEC has exclusive jurisdiction over the civil enforcement the federal campaign finance law.
```
emileighoutlaw: You're definitely right. I made that change, and good catch @adborden

emileighoutlaw:
That coming soon text for audits does look really awkward. 

Perhaps I should send a screenshot to Amy and Cheryl and ask if we can cut it down to something minimal like:

> "Coming soon: Commission audits of political committees"

??

noahmanger (approved):
This looks good to me. Once @emileighoutlaw 's content point is resolved feel free to merge.

noahmanger:
Oops, recanting my approval for the time being while Jen finishes some things up.

jenniferthibault:
Sorry Noah, that's my bad. I was experimenting on my own without communicating. 

I did a few small things that will improve the design regardless of the design changes that will come from adding the legal citations the general counsel wants.

- first, I added a `<p>` class to the list items, like we did on the statutes landing page, to make sure they stay within the container.
screen shot 2016-10-06 at 8 06 38 pm

- Then, I made all "coming soon" text italic to set it off from the more settled content. Since there's so much of it in this layout, I found it helpful.
- Last, I moved the "Search or browse closed MURs" into the search container, where it felt closer to the action and more relevant. This is the only place I took a content liberty @emileighoutlaw and removed the word "current" because the search also goes across archived MURs. 

screen shot 2016-10-06 at 8 06 30 pm


I don't feel like this design is restful on the eyes overall right now, but it's ok enough to wrap this PR and look at what the new citations section will do to the layout, then get picky. 👌 

Thanks for your patience everyone, I know this PR and this page have been a bit all over the place.

openfecwebapp/templates/legal-enforcement-landing.html
```diff
@@ -62,15 +62,13 @@
...
-    <p>
-      Search or browse closed MURs.
-    </p>
     <div class="u-no-print">
       <div class="slab slab--neutral slab--inline">
         <div class="container">
           <div class="row">
             <div class="usa-width-one-whole">
               <form action="{{ url_for('murs') }}" autocomplete="off" class="search__container  js-search">
+                <label class="label">Search or browse MURs</label>
```
noahmanger: This `<label>` needs `for="search-input"` so it's associated with the input explicitly.

openfecwebapp/templates/legal-enforcement-landing.html
```diff
@@ -0,0 +1,103 @@
+{% extends "layouts/legal-doc-landing.html" %}
+
+{% block content %}
+<header class="heading--main">
+  <h1>Enforcement matters</h1>
+</header>
+<div class="content__section">
```
noahmanger: Instead of nesting `<p>` tags in `<li>` tags, let's just add `content__section--narrow` to this `div` like we do below. It will achieve the same effect.

noahmanger (changes requested):
Ok, commented on just a couple things @jenniferthibault . I can also make the changes too.

noahmanger (approved):

noahmanger:
After checking with @jenniferthibault I'm going to merge this now and content edits can follow.

jenniferthibault:
Picking up where this leaves off over in https://github.com/18F/openFEC-web-app/pull/1622

PR #1624, opened 2016-10-07
Adding the correct class to headings with action

noahmanger:
Fixes an issue where the buttons in headings were breaking onto new lines instead of being floated to the right like: 
image

cc @jenniferthibault


jenniferthibault:
Thank you for getting this!

PR #1629, opened 2016-10-12
[WIP] Aggregate table paging and export options

xtine:
Paging for Election, Candidate, and Committee aggregate tables:
screen shot 2016-10-11 at 6 38 42 pm
screen shot 2016-10-11 at 6 39 14 pm

### DIFF
--- a/static/js/pages/candidate-single.js
+++ b/static/js/pages/candidate-single.js
@@ -70,7 +70,7 @@ function initFilingsTable() {
     path: path,
     columns: filingsColumns,
     order: [[2, 'desc']],
-    dom: tables.simpleDOM,
+    dom: 't<"results-info"lpi>',
     pagingType: 'simple',
     hideEmpty: true
   });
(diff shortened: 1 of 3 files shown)


xtine:
@noahmanger: can you help with getting the export buttons in? not sure how they are included.

noahmanger:
Ok! I think I got these going. We only have them for the receipts and disbursement tables on committee pages, which this enables. 

The DataTable code is definitely pretty tangled up, and I'm afraid I'm not helping the matter much, so let me know if you have better ideas. 

If this can get into the release it would be terrific, but we shouldn't rush if it's not ready.

static/js/pages/candidate-single.js
```diff
@@ -70,7 +70,7 @@ function initFilingsTable() {
     path: path,
     columns: filingsColumns,
     order: [[2, 'desc']],
-    dom: tables.simpleDOM,
+    dom: 't<"results-info"lpi>',
```
xtine: I wasn't sure if `simpleDOM` was used elsewhere yet. But if those are the only places it's being used then I'll replace it. Thanks!

xtine:
I think we should fit this into the release as everything seems to work. Cleaning up the DataTables `tables.js` should be another task to break out and make more manageable.

noahmanger:
👍

PR #1631, opened 2016-10-12
Fix/add feature flag

noahmanger:
This moves the link for the press page in the footer nav behind a feature flag, since we won't be including it in this release.

It also adds the latest updates link, also behind a feature flag.

With the flag enabled, it looks like:
image

This just brings parity between the web app and the CMS. We should turn this flag on on staging at some point, but it's not necessary right now.


ccostino:
Thanks, @noahmanger!

Issue #1627, opened 2016-10-07
Create "Who's spending the most" chart


noahmanger:
In order for users to see the top spenders in a two-year period, create an interactive visual of the top raising candidates, PACs and party committees. 

---

Desktop design:
![image](https://cloud.githubusercontent.com/assets/1696495/19208156/a23843bc-8cab-11e6-8f55-dae21d837bd6.png)

Mobile design:
![image](https://cloud.githubusercontent.com/assets/1696495/19207988/1360e5dc-8caa-11e6-8f40-a2284a2e405b.png)

**Criteria for completion:**
- [x] The user can view top spenders for the selected 2-year period of different types using a basic `select`:
   - [x] Presidential candidates
   - [x] House and Senate candidates
   - [x] PACs
   - [x] Party committees
- [x] Candidate bars are color-coded by party (blue for Democrats, red for Republicans)
- [x] Each entity listing links to the page for that entity in the selected two-year period
- [x] The user can change the two-year period of the data viewed using a `select`
- [x] The chart shows the date range the numbers are drawn from
- [x] The user can page through the top 25 entries, 10 per page
- [x] The user can view a modal with the methodology 
- [x] The user can follow a link to raw data (destination TBD)
- [x] The user can enter a "reaction" using the reaction widget

---

This story is focused on creating a minimum viable feature that can be shipped using the `/totals/` endpoint as currently implemented. As such, it does not include some aspects of the feature that will ultimately be there, such as: 
-  A more complex filter interaction with the ability to fine tune the type of PAC or party committee you're viewing
- The toggle-able key with the color codes for minor parties.

noahmanger:
Resolved by #1638

Issue #1626, opened 2016-10-07
Create "Who's raising the most" chart


noahmanger:
In order for users to see the top raisers in a two-year period, create an interactive visual of the top raising candidates, PACs and party committees. 

---

Desktop design:
![design](https://cloud.githubusercontent.com/assets/1696495/19176052/0c0c49ac-8bf1-11e6-8097-f63d2873dd75.png)

Mobile design:
![image](https://cloud.githubusercontent.com/assets/1696495/19207988/1360e5dc-8caa-11e6-8f40-a2284a2e405b.png)

**Criteria for completion:**
- [x] The user can view top raisers for the selected 2-year period of different types using a basic `select`:
   - [x] Presidential candidates
   - [x] House and Senate candidates
   - [x] PACs
   - [x] Party committees
- [x] Candidate bars are color-coded by party (blue for Democrats, red for Republicans)
- [x] Each entity listing links to the page for that entity in the selected two-year period
- [x] The user can change the two-year period of the data viewed using a `select`
- [ ] The chart shows the date range the numbers are drawn from
- [ ] The user can page through the top 25 entries, 10 per page
- [ ] The user can view a modal with the methodology 
- [ ] The user can follow a link to raw data (destination TBD)
- [ ] The user can log a reaction using the reaction widget

---

This story is focused on creating a minimum viable feature that can be shipped using the `/totals/` endpoint as currently implemented. As such, it does not include some aspects of the feature that will ultimately be there, such as: 
-  A more complex filter interaction with the ability to fine tune the type of PAC or party committee you're viewing
- The toggle-able key with the color codes for minor parties.

noahmanger:
Question: where should the "Data" link take you:
- To the top-level receipts / disbursements page for itemized data (e.g. beta.fec.gov/data/receipts)
- To the reports pages for the kind of entity you're looking at (e.g. beta.fec.gov/data/reports/presidential)
- To the data table page with all the same data on it, but filterable (e.g. beta.fec.gov/data/candidates/president)

noahmanger:
Resolved by #1638

PR #1638, opened 2016-10-19
Feature/top raisers spenders

noahmanger:
This adds new pages for raising and spending breakdowns and the top raising / spending charts for them.

demo

**Progress:**
- [x] Create chart for top raisers
- [x] Create chart for top spenders
- [x] Add methodology modal 
- [x] Add links to raw data
- [x] Ensure it works with mobile
- [x] Ensure it's good in terms of a11y
- [x] Pair with Jen on visual styles
- [x] Add tests

Resolves https://github.com/18F/openFEC-web-app/issues/1626
Resolves https://github.com/18F/openFEC-web-app/issues/1627

Requires https://github.com/18F/fec-style/pull/548


noahmanger:
@jenniferthibault what color should bars be when there's no party designation for an entity? 
image

emileighoutlaw:
This content looks good to me. I capitalized "presidential" so it's the same as all the other drop-down options (and our thus-far-on-the-website convention of capitalizing the first word in drop-downs). 

I brought this to my critique group today, and they felt like the colon read weird, since it interrupts the flow of what would otherwise be a normal sentence. What would y'all think about taking it out?

noahmanger:
I'm fine with that.

static/js/pages/breakdowns.js
```diff
@@ -0,0 +1,204 @@
...
+    '</div></div>',
+    {interpolate: /\{\{(.+?)\}\}/g}
+);
+
+// Store candidate office letters for to look up when a chart category is a candidate
+var candidateCategories = ['P', 'S', 'H'];
+
+function TopEntities(elm, type) {
+  this.$elm = $(elm);
+  this.type = type === 'raising' ? 'receipts' : 'disbursements';
```
xtine: why setting type twice?

static/js/pages/breakdowns.js
```diff
@@ -0,0 +1,204 @@
...
+  this.cycle = this.$elm.data('cycle');
+
+  this.$table = this.$elm.find('.js-top-table');
+  this.$dates = this.$elm.find('.js-dates');
+  this.$previous = this.$elm.find('.js-previous');
+  this.$next = this.$elm.find('.js-next');
+
+  this.init();
+
+  $('.js-cycle').on('change', this.handleCycleChange.bind(this));
```
xtine: why is this out of the block scope?

xtine (commented):
Looks good, nothing jumps out at me but I'm just curious about a few minor things.

static/js/pages/breakdowns.js
noahmanger: Ah, found a simpler way to do this by just passing `receipts` or `disbursements` directly in as the param instead of deriving it.

openfecwebapp/api_caller.py
```diff
@@ -253,29 +253,29 @@ def landing_mock_data():
         }
     }
 
-def load_top_candidates(sort):
+def load_top_candidates(sort, office=None, cycle=2016, per_page=5):
```
ccostino: Nice work parameterizing these values with the defaults!  Same for the changes below.

openfecwebapp/api_caller.py
```diff
@@ -253,29 +253,29 @@ def landing_mock_data():
...
-def load_top_parties(sort):
+def load_top_parties(sort, cycle=2016, per_page=5):
         response = _call_api(
             'totals', 'party',
-            sort_hide_null=True, cycle=2016, sort=sort, per_page=5
+            sort_hide_null=True, cycle=cycle, sort=sort, per_page=per_page
         )
         if response['results']:
-            return response['results']
+            return response
```
ccostino: I'm curious about this change and the ones related to it both in here and in `routes.py`.  Technically it's fine, but what was the reason behind it?  Also, is `response['results']` itself a dictionary or just the parent `results`?

openfecwebapp/utils.py
```diff
@@ -118,3 +118,14 @@ def election_title(cycle, office, state=None, district=None):
...
+def page_info(pagination):
+    """Generate a string showing number of results out of how many
+    based on a pagination object from an API response
+    """
+    page = pagination['page']
+    per_page = pagination['per_page']
+    count = pagination['count']
+    range_start = per_page * (page - 1) + 1
+    range_end = (page - 1) * 10 + per_page
+    return str(range_start) + '-' + str(range_end) + ' of ' + '{:,}'.format(count)
```
ccostino: Probably a bit nit-picky here, but this might be better written (and more Pythonic) this way:

`return '{range_start}-{range_end} of {count}'.format(range_start=range_start, range_end=range_end, count=count)`

static/js/modules/top-entities.js
```diff
@@ -0,0 +1,204 @@
...
+    });
+  }
+  this.loadData(this.currentQuery);
+};
+
+TopEntities.prototype.handlePagination = function(direction) {
+  var currentPage = this.currentQuery.page || 1;
+  if (direction === 'next') {
+    this.currentQuery.page = currentPage + 1;
+    this.$previous.removeClass('is-disabled');
```
ccostino: I noticed that here and just a bit further below there is this handling of a disabled state for the `previous` control but don't see any for the `next` control if you reach the last page.  Is this intentional?

ccostino (commented):
A few things jumped out at me when looking this over; nothing major and mostly just clarification to make sure I understood what was happening or wasn't missing something.

Very nice job with the tests, @noahmanger!

static/js/modules/top-entities.js
noahmanger: Ah! Good point. No reason other than there's often ~50 or more pages so unlikely to reach the end.

openfecwebapp/utils.py
noahmanger: Oh nice. Still learning the pyhonic way, so good tip.

ccostino: Always happy to help there! :-)

openfecwebapp/api_caller.py
ccostino: You know what, I might just be over-thinking this.  it caught my eye for whatever reason but more as a curiosity than anything.

noahmanger: Oops forgot to respond to this one. This change is because we also need to pass the `response['pagination']` to the front-end, so I changed it so just the entire `response` object gets passed.

noahmanger:
Ok, those edits are in.

ccostino:
Awesome, thanks @noahmanger!

PR #1639, opened 2016-10-19
formatting penalty as currency

anthonygarvan:


noahmanger:
Not sure if this addresses what you need to do here, but just an FYI there's a `currency` template filter in filters.py that can be used on the frontend.

noahmanger:
Oh nice! Seeing that you're just using the currency filter I can review this instead of @LindsayYoung . Do you have an example page I can check out?

noahmanger:
Found one. Looks good.

Issue #1603, opened 2016-09-26
Update landing charts to use live data API


noahmanger:
So that users always see the most up-to-date information, update the data landing page charts to use the new API once https://github.com/18F/openFEC/issues/1953 is complete.

---

**Criteria for completion:**
- [x] The charts use data from the API
- ~~The charts allow users to navigate back to earlier two-year cycles using the arrows~~
- [x] The methodology is updated to acknowledge the data is live
- [x] The disclaimer is removed

LindsayYoung:
Got a version going that returns:
```
{
   "year": 2015,
   "Receipts": -2051.35,
   "month": 1,
   "type": "party",
   "cycle": 2016
},
{
   "year": 2015,
   "Receipts": -1392.96,
   "month": 2,
   "type": "party",
   "cycle": 2016
},
```
If would be easy to add a date if you want it. 
We could try to group them by date but I don't know how I would do it off the top of my head.

noahmanger:
Cool. Yeah, that's close. So does each object represent a monthly total? How do you know which month each applies to?

LindsayYoung:
Yea, the total is the aggregate for that date. The year and month represent the date- I can just add date, to make it more intuitive.

noahmanger:
Oh! I missed the `month` field.

LindsayYoung:
OK, added a date field, I was originally afraid it might be confusing since we need a day, to make it consistent, it is the first of the month, I will try to add documentation so people don't think it applies to that day.

noahmanger:
Could you make it the last of the month? If not, what about just a `MM/YYYY` format?

LindsayYoung:
I can do the last day of the month and if that is in the future I can override it with yesterday's date. I can get rid of the month and year as separate vars.

sound good?

LindsayYoung:
Now it looks like:
```
{
   "cycle": 2016,
   "type": "candidate",
   "receipts": 8252528.86,
   "date": "2015-09-30T00:00:00"
},
{
   "cycle": 2016,
   "type": "candidate",
   "receipts": 66310157.81,
   "date": "2015-12-31T00:00:00"
},
```

noahmanger:
Excellent. Thanks!

noahmanger:
We still need to update the methodology: 

![image](https://cloud.githubusercontent.com/assets/1696495/19581880/449df3a2-96e5-11e6-9cac-17b2024220e3.png)

[...]

PR #1641, opened 2016-10-20
Landing chart connection to api

xtine:
Updates the Raising and Spending data landing charts to use data from the API. #1603 

Requires 18F/fec-style#552

### DIFF
--- a/openfecwebapp/templates/landing.html
+++ b/openfecwebapp/templates/landing.html
@@ -53,7 +53,6 @@
     <section class="content__section" id="raising">
       <h3 class="u-no-margin">Raising</h3>
       <p>The total amount raised, by group, until specific points in time.</p>
-      <p>This chart is a static snapshot. Search the most up-to-date data in the <a href="{{ url_for('advanced') }}">advanced data section</a>.</p>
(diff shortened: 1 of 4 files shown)


xtine:
@noahmanger: I'll need your help to cycle to previous 2 year periods.

noahmanger:
@xtine re: cycling to previous period, let's not worry about that for now. Let's just get this in and then add that on.

xtine:
With 18F/openFEC#1990, live datasets look more accurate:

screen shot 2016-10-20 at 3 01 42 pm
screen shot 2016-10-20 at 3 36 34 pm

While growth lines pattern looks good, the numbers are a bit off from the hardcoded values (beta.fec.gov/data). @LindsayYoung can you double check if there is a discrepancy from either side?

noahmanger:
The data seems to be very off, but I believe this is an API issue as I'm looking at the results there. 

For instance, by Aug 31, 2016, the data Paul gave us says $1.7b for candidate spending, and the API is showing ~$3.3b. 

But this looks good, so I'll merge and open a new issue for validating data.

PR #1649, opened 2016-10-24
Add fec_url to reports pages

noahmanger:
cc @jontours


jontours:
Verified this branch works for all the resources.

PR #1653, opened 2016-10-24
Make "view xxx" content more accessible

emileighoutlaw:
"View" generally isn't an ideal word for interface copy, because it can feel exclusionary to folks who use screenreaders and aren't "viewing" the text. 

Additionally, what we mean by "view" can vary depending on the context. This PR attempts to make our content more accessible and descriptive by changing the word "view" into one of the following:

- `Show` when we mean “display” information on the same page
- `Open` when we mean “opening” a document or taking users to a new page
- `Access` when we mean “making contact with” or “gaining access to" information
- `Check out` in the rare instance when we are offering alternative search pathways

 This PR also contains some edits to improve the consistency of our results ranges. Sometimes we had "Viewing XX-XX of XXX results." Sometimes we had "Showing XX to XX of XXX results." Sometimes we had "XX-XX of XXX results."

This PR moves to make all results ranges consistent with `XX-XX of XXX results`, which is the most concise and fastest to scan.

Content only. Anyone can review, but my gut is telling me that @noahmanger should take a gander

### DIFF
--- a/openfecwebapp/templates/advanced.html
+++ b/openfecwebapp/templates/advanced.html
@@ -179,7 +179,7 @@
 
 <div class="slab slab--neutral">
   <div class="container">
-    <h2>More ways to view this data</h2>
+    <h2>More ways to explore data</h2>
     <div class="grid grid--4-wide">
       <div class="grid__item">
         <a href="{{ url_for('election_lookup') }}">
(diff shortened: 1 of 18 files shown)


LindsayYoung:
It's fun working with writers!

Issue #1650, opened 2016-10-24
Landing charts not drawing line for communication filers


noahmanger:
Something is happening in the data transformation response and the data for "other" category entities are getting lost. I'll look into it. Needs to be fixed before release.

noahmanger:
Resolved by #1657

Issue #1643, opened 2016-10-20
Update snapshot on landing page charts


noahmanger:
Now that we have an API endpoint for the charts, we need to remove the mock data that populates the snapshot data on the server-side:
![image](https://cloud.githubusercontent.com/assets/1696495/19581850/0dfd7d36-96e5-11e6-8ca8-0cc271ad7514.png)

noahmanger:
Resolved by #1657

PR #1657, opened 2016-10-25
Chart fixes

noahmanger:
This fixes a couple bugs that need fixing before releasing the live data charts.
1. It removes the mock server-side data that was populating the chart on initial page load and instead just moves the dotted line cursor to the last data point.
2. It restores the line for communication spenders by hard-coding the entity types rather than deriving them from the data. This is because the first data point didn't include an entry for `other` which was preventing the line from getting drawn.

cc @LindsayYoung @ccostino 

Resolves https://github.com/18F/openFEC-web-app/issues/1650
Resolves https://github.com/18F/openFEC-web-app/issues/1643

### DIFF
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -235,24 +235,6 @@ def result_or_404(data):
         abort(404)
     return data['results'][0]
 
-def landing_mock_data():
-    return {
-        'raising': {
-            'total': 5409774242.62,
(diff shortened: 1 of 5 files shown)


ccostino:
It looks like Hound is stuck again for some reason, but the changes all look fine to me.  Thanks, @noahmanger!  It's awesome seeing these come together. :-)

PR #1660, opened 2016-10-25
Oppose/support

emileighoutlaw:
I noticed that we were inconsistent in whether we said "oppose/support" or "support/oppose" on our profile pages. This PR makes consistent so that all pages read "support/oppose," which I believe is more in line with readers' mental models.

This also fixes one of our lowercase "president" instances, which we're capitalizing now.

Content only— anyone can review.


LindsayYoung:
Thank you!

PR #1661, opened 2016-10-26
Reverting to static data

noahmanger:
Because we haven’t finished the API fixes for the charts, this reverts
to static data by:
- Loading data from static json file
- Adding methodology note
- Adding disclaimer

cc @LindsayYoung


LindsayYoung:
Thank you Noah. Hoping we can get a mini release of this with live data soon!

PR #1664, opened 2016-10-27
[MERGE WITH GITFLOW] Hotfix/chart placeholder

noahmanger:
Adds static placeholder images and links to breakdown pages instead of static data for line charts.

image

---

UPDATE: FEC Style dependency is all set so this just needs to be merged with gitflow.


ccostino:
Thanks, @noahmanger!

PR #1673, opened 2016-11-01
[MERGE WITH GITFLOW] Hotfix/adding live data

noahmanger:
This adds the code to use live data for the overview charts. It does excludes lines for the "communication filer" category as that data is still being worked out.

cc @LindsayYoung @xtine


ccostino:
Thanks, @noahmanger, @emileighoutlaw, and @LindsayYoung!

Issue #1654, opened 2016-10-24
Design preferred order for participants list on MURs pages


jenniferthibault:
Pointed out by the FEC enforcement folks during their review:

When multiple respondents are listed on a case, often the counsel are listed first, with the primary respondents and complainants (the two most interesting and useful participants) listed very last.

Here's an example where this happens: https://fec-stage-proxy.18f.gov/data/legal/matter-under-review/5903/

In grooming the feedback, @anthonygarvan mentioned that we can choose this order, so we should set it!

Steps to completion:
- [x] Review all options for participant relationships
- [x] Order them by relevance / usefulness / priority
  - (maybe we can learn this through a card sort activity with someone at the FEC?)
- [x] Transfer or create issue for implementation

jenniferthibault:
cc @nickykrause — this is an information design task that we might be able to knock out relatively easily. Would you like first crack? Or if you're full with testing wireframes this sprint I don't mind trying.

nickykrause:
@jenniferthibault: I'd like to, but I suspect I will be too busy this week with trying to get more realistic content in the wires and designing/doing the testing on them (assuming people sign up for testing, anyway...). It's hard to know how much time I'll have until we know how recruiting goes

jenniferthibault:
I think I can take this, which would unblock a MUR-focused dev task.

@anthonygarvan or @vrajmohan is it possible to surface a list of all the possible participant relationship categories from the database or code? Or should I attempt to dig this up manually?

vrajmohan:
We can get it from the database. I'll update this ticket with the list over the weekend.

vrajmohan:
1.  Previous Respondent
2.  Representative
3.  Treasurer
4.  Primary Respondent
5.  Respondent
6.  Complainant
7.  Law Firm
8.  Respondent's Counsel
9. Opposing counsel
10.  Previous Treasurer

jenniferthibault:
Thanks Vraj, I've reached out to Cheryl to help guide the order. As soon as I hear back I'll report in!

jenniferthibault:
Cheryl connected me with Stephen and Dawn, who helped clarify many things about MUR participants for me.

For everything MUR-related, we should always list the Respondent before the Complainant, because in the MUR process, the MUR is about the alleged actions of the respondent, not the complainant. For that reason, MURs are routinely referred to by the name of the first respondent of most prominent respondent.

That means we should:

[...]

PR #1682, opened 2016-11-04
Feature/mur participant order

vrajmohan:
Fixes https://github.com/18F/openFEC-web-app/issues/1654

### DIFF
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -12,7 +12,6 @@ from openfecwebapp import config
 from collections import OrderedDict
 
 MAX_FINANCIALS_COUNT = 4
-RESPONDENT_ROLES = ['Respondent', 'Primary Respondent', 'Previous Respondent']
 
 
 session = requests.Session()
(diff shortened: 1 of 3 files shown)


openfecwebapp/api_caller.py
```diff
@@ -268,3 +259,47 @@ def load_top_parties(sort, cycle=2016, per_page=5):
...
+        "Respondent",
+        "Previous Respondent",
+        "Treasurer",
+        "Previous Treasurer",
+        "Complainant",
+        "Respondent's Counsel",
+        "Opposing counsel",
+        "Representative",
+        "Law Firm",
+    ]
```
jenniferthibault: @vrajmohan I just checked with @emileighoutlaw to be sure on this, and she confirmed: Could we adjust the capitalization of these fields to have only the first word capitalized?
`Primary respondent` 
`Previous respondent` 
`Previous treasurer`
`Respondent's counsel`
`Law firm`

We realize the Document tables in MURs will need this as well, and @emileighoutlaw is going to file a separate issue for that for the next sprint.

jenniferthibault: Just realized that how these fields display might not be controlled on the front end. If this requires a change to the API....then not required for this release.

anthonygarvan:
looks good!

PR #1687, opened 2016-11-07
Fixing the date formatting on MURs

noahmanger:
cc @emileighoutlaw 

image


emileighoutlaw:
Since this appears in a chunk of content, how would you feel about changing this to 

> {{ date_property | date('%B %e, %Y') }}

so that it displays as: `March 13, 1978`

noahmanger:
I like it.

PR #1693, opened 2016-11-08
Adding means filed to filings panels

noahmanger:
This adds the means filed field to the report panels:

image

It will show "eFiling" and "Paper filing".

@emileighoutlaw is that the right labelling?


emileighoutlaw:
We're going with "electronic filing" whenever possible. I just pushed that change to this PR.

emileighoutlaw:
How would y'all feel about changing `Means filed` to `Filing method` ? 

Sounds a little less... mean :)

xtine (approved):

noahmanger:
I like filing method! I was struggling to come up with the right word. Can you make a pr?

PR #1695, opened 2016-11-09
Reports capitalization

emileighoutlaw:
Content-only PR. Makes 24-Hour Report and 48-Hour Report capitalization consistent.

Resolves: https://github.com/18F/fec-cms/issues/586


jenniferthibault:
Since this changes javascript and python files, I think it requires technical review. I'm not sure how case sensitive those fields are.

emileighoutlaw:
If this requires a technical review, maybe @ccostino or @LindsayYoung wants to review?

openfecwebapp/templates/partials/independent-expenditures-filter.html
```diff
@@ -32,8 +32,8 @@
         <span class="button--alt">Regularly scheduled reports</span>
       </label>
       <label for="notice" class="toggle">
-        <input type="radio" class="toggle" value="true" id="notice" name="is_notice" data-tag-value="24 and 48 hour reports">
-        <span class="button--alt">24- and 48-hour reports</span>
+        <input type="radio" class="toggle" value="true" id="notice" name="is_notice" data-tag-value="24- and 48-Hour Reports">
```
LindsayYoung: Maybe it is supposed to be "24-" but I thought I would double check

emileighoutlaw: Yep! It is supposed to be "24-"

PR #1696, opened 2016-11-09
Hybrid / Super capitalization

emileighoutlaw:
Partially resolves: https://github.com/18F/fec-cms/issues/587

Makes Hybrid PAC and Super PAC capitalization consistent.

Content only.


jenniferthibault:
This has changes to a python file, and I'm not sure if case-sensitivity impacts those fields. This should have a technical review.

PR #1699, opened 2016-11-10
Electronic filing

emileighoutlaw:
One of our UI consistency decisions was to use the phrase "electronic filings" rather than the more jargon-y "e-file" or "eFile" whenever possible.

Content only.

Partially resolves: 18F/fec-cms#585


emileighoutlaw:
Ping! This one has been open for a while — since it's content only, maybe @jenniferthibault wants to merge?

noahmanger:
I see that this changes the tooltip content but not the actual text of the button. Is that intentional?

emileighoutlaw:
Nope, not intentional! I think it's fixed now. Thanks @noahmanger :)

PR #1708, opened 2016-11-23
[Merge with gitflow] remove ethnio

LindsayYoung:
we are over our limit


ccostino:
Thanks for helping take care of this, @LindsayYoung!

PR #1719, opened 2016-12-01
Removing feature flags from footer nav items

noahmanger:
This removes the feature flags on the footer nav items so that they will show up:

image


ccostino:
Thanks, @noahmanger!

PR #1720, opened 2016-12-01
Format citations according to type

vrajmohan:
Use the following conventions:
- 11 CFR 112.1(a)
- 52 U.S.C. §30123
- Periods and section symbol for U.S.C.
- No period or section symbol for CFR
- Linking only the part of the citation that follows CFR or U.S.C.

Fixes 18F/openFEC#2069

https://github.com/18F/openFEC/issues/2069

### DIFF
--- a/openfecwebapp/templates/legal-current-mur.html
+++ b/openfecwebapp/templates/legal-current-mur.html
@@ -92,8 +92,14 @@
                 <td> {{ entry.respondent }}</td>
                 <td>
                 {% for citation in entry.citations %}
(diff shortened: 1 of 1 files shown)


emileighoutlaw:
Even though this is merged, I think there's still a stray space between the statutes and the section number in citations, so it's displaying as `52 U.S.C. § 30123` rather than `52 U.S.C. §30123`

Would that be an easy thing to fix or a hard thing to fix @vrajmohan ?

vrajmohan:
Yes, on it. Thanks for noticing it.

PR #1724, opened 2016-12-02
Highlighting correct parent on nav items

noahmanger:
Chipping away at bugs, this was an easy one: now, if you're on a page in the legal resources section, the "Legal resources" nav item will be highlighted. 

To do this, routes need to have a template variable called `parent` passed in, with either `data` or `legal` as the value. `navigation.html` will then highlight the correct item in the menu.

Resolves https://github.com/18F/fec-eregs/issues/170

### DIFF
--- a/openfecwebapp/routes.py
+++ b/openfecwebapp/routes.py
@@ -26,6 +26,7 @@ def search():
     else:
         return render_template('landing.html',
             page='home',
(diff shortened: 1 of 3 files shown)


ccostino:
Looks good, thanks @noahmanger!

PR #1726, opened 2016-12-09
Feature/ao-filters

noahmanger:
This adds the basic makings for the AO filters. 

image

To do this, I changed the implementation of the filter panel slightly so that the general filter panel code is a block in `legal-doc-search-results.html` that individual doc search pages can populate with the filters of their choosing.

There are now three macros for legal filters:
- `keyword_search` which creates the combo input + button (used by non-AO pages currently)
- `text_filter` which creates a basic text input
- `date_filter` which creates the min and max date range

Remaining to do:
- [x] Add input masks to the date filter in order to format dates correctly

cc @anthonygarvan

### DIFF
--- a/openfecwebapp/templates/layouts/legal-doc-search-results.html
+++ b/openfecwebapp/templates/layouts/legal-doc-search-results.html
@@ -19,7 +19,21 @@
 </header>
 
 <section class="main__content--full data-container__wrapper">
-  {{ legal.filter('Search %s' % (document_type_display_name), result_type, query) }}
+  <div id="filters" class="filters is-open">
+    <button class="filters__header js-filter-toggle" type="button">
+      <span class="filters__title">Search {{ document_type_display_name }}</span>
+    </button>
+    <div class="filters__content">
+      <form id="category-filters" action="{{ url_for(result_type) }}">
+        <div class="filters__inner">
+          <div class="filter" id="search-field">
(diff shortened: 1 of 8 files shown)


noahmanger:
Alright this should be good to go @anthonygarvan . Feel free to either take over this branch for your work or merge and then do your stuff.

emileighoutlaw:
A couple quick content notes: 
- It feels a little odd to me to have a the verb "search" with `keywords` but not with other search boxes
- `Issued date` feels a little weird when I say it aloud 

How about:

> ### Search advisory opinions
> **Keywords** _(Should this be plural? Can people search by more than one?)_
> [_________]

...

> **Date issued**

...


What do you think, @noahmanger? If you agree, I'll push those changes to this PR.

noahmanger:
That  makes sense. And yeah, keywords should be plural.

emileighoutlaw:
Awesome. Pushed those commits

anthonygarvan:
@noahmanger are min / max dates supposed to be datepickers? If so it's not showing up for me locally.

noahmanger:
@anthonygarvan no date picker for now, but it should show up as a masked (`mm/dd/yyyy`) input.

noahmanger:
Closing in favor of @anthonygarvan 's new one.

PR #1732, opened 2016-12-13
MM-DD-YYYY to MM/DD/YYYY

emileighoutlaw:
Noticed this stray date formatting inconsistency.

Content only

### DIFF
--- a/static/js/modules/columns.js
+++ b/static/js/modules/columns.js
@@ -45,10 +45,10 @@ var receiptDateColumn = {
     var parsed;
     if (meta.settings.oInit.path.indexOf('efile') >= 0) {
       parsed = moment(row.receipt_date, 'YYYY-MM-DDTHH:mm:ss');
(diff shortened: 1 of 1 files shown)


xtine:
Is this trying to change dates formatted like this:
screen shot 2016-12-13 at 3 56 16 pm
to a MM/DD/YYYY for Receipts datatable page? If so, I think this change doesn't actually do that and needs a change specifically in this line: https://github.com/18F/openFEC-web-app/blob/f369659a864d422c4def8d302158558a90712483/static/js/modules/columns.js#L428

noahmanger:
This change fixes it on the filings page, but to fix it elsewhere required changing this Handlebars helper in fec-style https://github.com/18F/fec-style/blob/master/js/helpers.js#L17 which I just added as part of this PR https://github.com/18F/fec-style/pull/581. 

So I'll merge this.

emileighoutlaw:
Thanks y'all 💟

PR #1733, opened 2016-12-14
Moved search.js to fec-style repo and updated refs

patphongs:
Search.js controls the select box toggle between Candidate and Committee searches. This function had to be moved from openfec-web-app in order to reference in other places such as the homepage.


noahmanger:
So, I just re-read `Search.js` and I think we don't actually need it anywhere anymore. All it does is change the placeholder text. The value of the `select` is what determines the kind of search that's submitted. And the typeahead itself is initialized separately.

So I think we don't need to add this to fec-style and can instead just use this PR to remove it outright.

What do y'all think?

noahmanger:
To clarify: we don't use placeholders any more, which is why we don't need the placeholder value to change.

patphongs:
@noahmanger search.js is still currently used to support the toggling between candidates and committees searching. The typeahead relies on search.js to emit the ```searchTypeChanged``` event in order to change the API call that is made. If we want to remove placeholder text completely, we can remove the additional logic that search.js performs to handle placeholder text. 

I have re-added search.js in my recent comment in fec-style commit https://github.com/18F/fec-style/pull/582/commits/1845ec111bbb5e31666d1ad7ed19abd2a95131bf

noahmanger:
Oh! Yeah you're totally right. I spaced on that.

PR #1742, opened 2016-12-15
Feature/support advanced ao filters

anthonygarvan:


openfecwebapp/api_caller.py
```diff
@@ -49,14 +49,21 @@ def load_search_results(query, query_type='candidates'):
...
+def load_legal_search_results(query, query_type='all', ao_no=None, ao_name=None,
+                                offset=0, limit=20):
     filters = {}
     if query:
         filters['q'] = query
         filters['hits_returned'] = limit
         filters['type'] = query_type
         filters['from_hit'] = offset
 
+        if ao_no and ao_no[0]:
```
vrajmohan: This idiom is unclear to me. Why does `if ao_no` not suffice? What are we guarding against?

anthonygarvan: _sighs_. So, the standard HTML form sends back empty keywords, but we are interpreting them as a list, so we get back results that look like `['']`: a list with a single empty string as an element.

anthonygarvan:
@vrajmohan I think this is good to go now, pls merge if it's ok with you.

static/js/pages/legal.js
```diff
@@ -1,10 +1,15 @@
 'use strict';
 
 /* global require */
+require('jquery.inputmask');
+require('jquery.inputmask/dist/inputmask/inputmask.date.extensions.js');
+require('jquery.inputmask/dist/inputmask/inputmask.numeric.extensions.js');
```
noahmanger: Responding to @vrajmohan 's question on my previous PR,`inputmask.numeric.` isn't required. Copy-and-paste mistake. :)

anthonygarvan: thanks! removed.

PR #1748, opened 2016-12-15
Prepare AO landing page

noahmanger:
I forgot to remove the mock data from the new AO landing page before it was merged in, so this removes that and also comments out the pending AO section. I could remove that entirely if we'd prefer to not have commented code, but that should be done soon. 

The remaining task here is to add the last two months of AOs to this page. I'm having local API trouble so can't do this at the moment, but should be able to add on Friday.

### DIFF
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -53,11 +53,13 @@ def load_legal_search_results(query, query_type='all', ao_no=None, ao_name=None,
                                 ao_min_date=None, ao_max_date=None,
                                 offset=0, limit=20):
     filters = {}
-    if query:
-        filters['q'] = query
+    if query or query_type == 'advisory_opinions':
         filters['hits_returned'] = limit
         filters['type'] = query_type
         filters['from_hit'] = offset
+
+        if query:
+            filters['q'] = query
 
         if ao_no and ao_no[0]:
             filters['ao_no'] = ao_no
--- a/openfecwebapp/views.py
+++ b/openfecwebapp/views.py
@@ -70,7 +70,7 @@ def render_legal_mur(mur):
 
 def render_legal_ao_landing():
     today = datetime.date.today()
-    ao_min_date = today - datetime.timedelta(weeks=52)
+    ao_min_date = today - datetime.timedelta(weeks=26)
     results = api_caller.load_legal_search_results(query='', query_type='advisory_opinions', ao_min_date=ao_min_date)
     recent_aos=OrderedDict(sorted(results['advisory_opinions'].items(), key=lambda item: item, reverse=True))
     return render_template('legal-advisory-opinions-landing.html',
--- a/openfecwebapp/views.py
+++ b/openfecwebapp/views.py
@@ -67,6 +67,17 @@ def render_legal_mur(mur):
         parent='legal'
     )
 
+def render_legal_ao_landing():
+    today = datetime.date.today()
+    ao_min_date = today - datetime.timedelta(weeks=20)
+    results = api_caller.load_legal_search_results(query='', query_type='advisory_opinions', ao_min_date=ao_min_date)
+    recent_aos=results['advisory_opinions']
+    return render_template('legal-advisory-opinions-landing.html',
+        parent='legal',
+        result_type='advisory_opinions',
+        display_name='advisory opinions',
+        recent_aos=recent_aos)
+
 
 def to_date(committee, cycle):
     if committee['committee_type'] in ['H', 'S', 'P']:
--- a/openfecwebapp/views.py
+++ b/openfecwebapp/views.py
@@ -9,6 +9,7 @@ from flask.ext.cors import cross_origin
 from webargs import fields
 from webargs.flaskparser import use_kwargs
 from marshmallow import ValidationError
+from collections import OrderedDict
 
 import github3
 from werkzeug.utils import cached_property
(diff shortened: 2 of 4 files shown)


noahmanger:
Ok, I added an API call to show the last 26 weeks of AOs (an arbitrary cut-off that returns a reasonable amount), ordered in reverse order by their number. This required making a change to the `load_legal_search_results` function so that it would work if there's no `query` param. I'm feeling a _little_ out of my depth on the Python front, so if I'm doing anything weird or wrong, of course please let me know. 

image

openfecwebapp/views.py
```diff
@@ -67,6 +68,17 @@ def render_legal_mur(mur):
         parent='legal'
     )
 
+def render_legal_ao_landing():
+    today = datetime.date.today()
+    ao_min_date = today - datetime.timedelta(weeks=26)
+    results = api_caller.load_legal_search_results(query='', query_type='advisory_opinions', ao_min_date=ao_min_date)
+    recent_aos=OrderedDict(sorted(results['advisory_opinions'].items(), key=lambda item: item, reverse=True))
```
ccostino: I see what you're trying to do here and you're very much on the right track!  This caught my eye though because of the sort key you've defined: `lambda item: item`.  It looks like this is producing the behavior you want, but I want to make sure that's actually the case and explain a bit of what's happening here.  If this is diving too far into the weeds or if this explanation is unwarranted, totally sorry and please feel free to disregard!

I'm assuming that `results['advisory_opinions']` has another dictionary for its value given the `.items()` call on it.  With that being the case, let's use a simpler `dict` as an example:

```python
test_dict = {'one': 1, 'two': 2, 'three': 3, 'four': 4,}
```

When you have a `dict` and you call either the `keys()`, `values()`, or `items()` methods on it, an iterable object that behaves *like* a `list` but is **not** actually a `list` is returned.  Continuing with the example, here is what is returned with those calls:

```python
test_dict.keys()  # returns dict_keys(['four', 'three', 'two', 'one'])
test_dict.values()  # returns dict_values([4, 3, 2, 1])
test_dict.items()  # returns dict_items([('four', 4), ('three', 3), ('two', 2), ('one', 1)])
```

Notice that the `test_dict.items()` call returns an iterable containing a `list` of `sets` (or `tuples`) that comprise of the key in the first position and the value in the second for each item in the `dict`.

Going back to the `sorted` statement and the key defined there, each individual `item` is being used as the sort key, which means it is looking at each key-value pair as a whole.  This isn't inherently bad as it does work!  But, what is it actually sorting against and what is it returning?  Using the example `dict`, here's the output of a similar `sorted` statement:

```python
sorted(test_dict.items(), key=lambda item: item, reverse=True)
# returns [('two', 2), ('three', 3), ('one', 1), ('four', 4)]
```

It returns a `list` of `tuples` (`sets`) that make up the key-value pairs, sorted in reverse order by the keys by the looks of it.  We can confirm this with a couple of similar statements:

```python
# Remove the reverse=True flag:
sorted(test_dict.items(), key=lambda item: item)
# returns [('four', 4), ('one', 1), ('three', 3), ('two', 2)]

# Remove the items() call on the dictionary and just sort the dictionary itself:
sorted(test_dict, key=lambda item: item, reverse=True) 
# returns ['two', 'three', 'one', 'four']

# Try sorting on the values explicitly:
sorted(test_dict.values(), key=lambda item: item, reverse=True) 
# returns [4, 3, 2, 1]
```

So we can see that it's the keys that are being sorted on even when going against the key-value pair.  Now, in the original statement, having the `list` of `tuples` returned is actually the desirable output because it is being fed into an `OrderedDict` object.  Each key-value pairing in the `list` is split apart in the `OrderedDict` construction so that the first piece becomes the key and the second becomes its associated value, thus producing the result you desire.  For completeness sake, here's what is actually returned with that call (using the example `dict` again):

```python
test_ordered_dict = OrderedDict(sorted(test_dict.items(), key=lambda item: item, reverse=True))
# OrderedDict([('two', 2), ('three', 3), ('one', 1), ('four', 4)]) is returned
```

It's an object that contains an iterable (a `list`-like object) of the `tuples`, but you can reference things as you would in a normal `dict`, e.g., `test_ordered_dict['one']`.  Of course, the desired result of this is that if you loop/iterate over the items in an `OrderedDict` they are in the order you expect them to be.

Again, hopefully this was useful and helpful, but I understand if I dove too far into something folks may already understand!

ccostino:
@noahmanger, I updated this PR to target the release branch instead now that it's available.

openfecwebapp/api_caller.py
```diff
@@ -53,12 +53,14 @@ def load_legal_search_results(query, query_type='all', ao_no=None, ao_name=None,
                                 ao_min_date=None, ao_max_date=None,
                                 offset=0, limit=20):
     filters = {}
-    if query:
-        filters['q'] = query
+    if query or query_type == 'advisory_opinions':
```
noahmanger: So, it's the behavior I believe we want in _this_ scenario, but I'll defer to @anthonygarvan or @vrajmohan as to if this is wise on the whole.

openfecwebapp/views.py
noahmanger: Thanks for explaining! I think I get it, and I'm pretty sure it's what we want. `api_caller.load_legal_search_results()` returns an `OrderedDict` for `results['advisory_opinions']` that looks like:

```
OrderedDict([
  ('2016-09', [{'category': 'Final Opinion',...}]),
  ('2016-07', [{'category': 'Final Opinion',...}])
])
```

Which we then want to loop through on the front-end. So it seems like it makes sense to just turn that into a new `OrderedDict` sorted by the first value of each pair.

ccostino: You're welcome!  It does sound like what you want based on what you've described - if the recent AOs should be in reverse chronological order then it looks like you're already good to go with things as they are. :-)

anthonygarvan:
looks good!

PR #1757, opened 2016-12-22
making dates required to avoid parse error

anthonygarvan:
This makes the date fields required. If they are submitted as empty, webargs throws a parse error that cannot be handled by setting missing=None (since it is not missing, but actively set to an empty string). Long term solution here is to replace native submit functionality with a js submit.


noahmanger:
Thanks. Yeah this helps prevent the worst case scenario. I have a separate issue for enhancing these with js submit functionality.

PR #1760, opened 2016-12-24
Give name for candidates w/o financial data

busterroni:
This should solve 18F/FEC#560. This is my first contribution to this repository and first time working with Flask so while I'm confident this works, additional testing should be done to make sure.


noahmanger:
Thanks for doing this! Looks great.

Issue #1738, opened 2016-12-15
Update filings and reports tables with new amendment information 


noahmanger:
So that users can always understand the version status of a filing, show the amendment status and history in the filings and reports tables.  

---

🖌  [Panel design](https://cloud.githubusercontent.com/assets/11636908/21234683/a22ce0c0-c2c1-11e6-9484-8786a018672b.png), [table design](https://cloud.githubusercontent.com/assets/11636908/21234682/a2257574-c2c1-11e6-8d99-794d26c9d7df.png) | 🗣  [Discussion](https://github.com/18F/openFEC-web-app/issues/1715#issuecomment-267388367)

**Completion criteria:**
- [x] Add "Version" column to the /filings page table
- [x] Add "Original" or "Version [#]" to the document description on /filings page table 
- [x] Update the reports details panel to reflect the designs, including proper date formats
- [x] Update filters on `/filings` and `/reports/**` pages to include a filter for "Version" and "Document status" (per comment below)
- [x] Add the table and panel info on the /reports/ pages as well
- [x] Update all links to the `/reports/**/` pages to not include `?is_amended=false` (this is a frequent request from RAD)
- [x] Show default value when version status is unknown (per https://github.com/18F/openFEC-web-app/issues/1806)

noahmanger:
Circling back on the open question here. 

**Reports pages:** I think we need to include the same version column and "original" vs. "amendment" word in the report title. 

**Filters:** I think we should replace the current amendment-y filters on both filings and reports with the same two filters:

> **Version**
> [ ] Most recent version 
> [ ] Past version

> **Document status**
> [ ] Original
> [ ] Amendment 

This would replace both this filter on the reports pages:
![image](https://cloud.githubusercontent.com/assets/1696495/21703252/95427a78-d365-11e6-9450-2dc52d369eda.png)

And this one on filings:
![image](https://cloud.githubusercontent.com/assets/1696495/21703267/b37c0770-d365-11e6-9ae8-cc2eee61da35.png)

Additionally, we've gotten a lot of feedback from folks requesting that the links to reports pages default to showing all reports (whereas now it defaults to `is_amended=false`).

cc @emileighoutlaw for language and @jontours to make sure the logic makes sense for the filter.

noahmanger:
One last change here. I'm adding another criteria for completion which is to ensure that this all shows up when viewing raw data as well, which depends on https://github.com/18F/openFEC/issues/2147

noahmanger:
[...]

PR #1768, opened 2017-01-03
Filings Amendment Status

xtine:
Update Filings table with version and amendment information. https://github.com/18F/openFEC-web-app/issues/1738

screen shot 2017-01-03 at 3 00 09 pm

Relies on https://github.com/18F/fec-style/pull/599

### DIFF
--- a/static/js/modules/columns.js
+++ b/static/js/modules/columns.js
@@ -281,11 +281,17 @@ var filings = {
     orderable: false,
     render: function(data, type, row) {
       var doc_description = row.document_description ? row.document_description : row.form_type;
+      var show_version = true;
       var is_original = false;
       var amendment_num = 1;
       var pdf_url = row.pdf_url ? row.pdf_url : null;
       var csv_url = row.csv_url ? row.csv_url : null;
       var fec_url = row.fec_url ? row.fec_url : null;
+
+      // because of messy data, do not show if not e-filing or null amendment indicator
+      if(row.means_filed !== "e-file" || row.amendment_indicator === null) {
+        show_version = false;
+      }
 
       if (row.amendment_indicator === 'N') {
         is_original = true;
@@ -297,6 +303,7 @@ var filings = {
 
       return reportType({
         doc_description: doc_description,
+        show_version: show_version,
         is_original: is_original,
         amendment_num: amendment_num,
         fec_url: fec_url,
--- a/static/js/modules/columns.js
+++ b/static/js/modules/columns.js
@@ -34,10 +34,10 @@ var versionColumn = {
   className: 'hide-panel hide-efiling column--med min-desktop',
   render: function(data) {
     if (data === true) {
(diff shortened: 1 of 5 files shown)


xtine:
@jontours: I need some logic help on the Original/Amendment <Version Number> description. I'm still getting confused because there are some filings that have `is_amended: false` but also has `amendment_indicator: A` and an `amendment_chain`, so is that Original or an Amendment? Also not totally sure how to get amendment version number based on `amendment_chain` and recent/previous file numbers.

xtine:
@noahmanger: I've updated the amendment versioning logic that should be correct for all e-filings, but the data might not be totally correct for paper. Should I leave out the document versions for paper filings until the data is cleaned/verified?

noahmanger:
@xtine so for paper, what would it show?

xtine:
@noahmanger: it would just have the document name without original/amendment next to it like how it currently is.

jontours:
Some of the paper filings are messy, so messy that not only can we not tell if it's an original or amended filing (without looking at the PDF), but we have not amendment chain for them.  After talking to @xtine we thought it may best to keep them as is until we can clean the data (we being us and the FEC).

jontours:
Annddd it's even messier than that it seems, after digging around in the data some there are `e_file` rows that have null amendment indicators.  So @xtine maybe it's best to just not provide any of the new styling info if `amendment_indicator` is null.

xtine:
I've updated it to only show e-filings and hiding any `null` values to avoid showing misleading data. It shows less data but I'd rather only show accurate data than more potentially incorrect data.

noahmanger:
Are you hiding only the amendment status stuff or hiding the records entirely?

xtine:
Just the amendment status stuff.

static/js/modules/columns.js
```diff
@@ -268,20 +281,39 @@ var filings = {
...
+      if(row.means_filed !== 'e-file'|| row.amendment_indicator === null) {
+        show_version = false;
+      }
+
+      if (row.amendment_indicator === 'N') {
+        is_original = true;
+      }
+
+      if (row.amendment_chain) {
+        amendment_num = row.amendment_chain.length;
```
noahmanger: I think this should be `row.amendment_chain.length - 1`? The numbers seem like they're inflated by 1.

static/js/modules/columns.js
```diff
@@ -29,6 +29,19 @@ var amendmentIndicatorColumn = {
   },
 };
 
+var versionColumn = {
+  data: 'most_recent',
+  className: 'hide-panel hide-efiling column--med min-desktop',
+  render: function(data) {
+    if (data === true) {
+      return '<div class="icon-circle--check-outline--inline--left"></div>Most recent version';
```
noahmanger: Small thing, but I see that elsewhere we usually do `<i>` for icons instead of `<div>`

noahmanger:
This is great.

PR #1771, opened 2017-01-04
Adding basic support for raw data

noahmanger:
This adds the ability to toggle between raw and processed schedule E data.

It was pretty easy to get this going roughly, but I hit some missing API features that we need, which I wrote up here https://github.com/18F/openFEC/issues/2135.

cc @jontours in case you want to use this to test out the API. 

This took <20 minutes to throw together, so it'll be easy to do for the other pages as well. 

---

### Update
I added the toggle to the receipts and disbursements pages as well to help test the API. 

---

### Update 2
In retrospect, my "this is so easy" statement was a little premature. But this is finally ready for review. A few notes in addition to the obvious:

- Because the raw and processed panels each exist in the DOM at the same time (but one is hidden), I had to add a way to specify different IDs for the various filters generated by the same macros. I used `id_suffix` as an optional param on these macros to do this
- Because the raw endpoints use a different type of pagination, the table switcher now passes in a `paginator` option in order to change between our `SeekPaginator` and `OffsetPaginator` depending on what's needed

Requires https://github.com/18F/fec-style/pull/607


noahmanger:
@jontours I just updated this to add the toggle to the receipts and disbursements pages.

jontours:
I just verified this works with independent expenditures, noticed only one problem.  For `support_oppose_indicator` the parameter being passed to the api via the url param is: `support_oppose_indicator`.  But that's it, no variable to filter on, it should be, e.g `support_oppose_indicator=S` for supports.

jontours:
Just verified functionality for schedule a, all is well except the receipt date column.  The values aren't showing, the field in the api is `contribution_receipt_date`.

noahmanger:
@jontours can we rename the field int he API to `receipt_date` so that it mirrors the regular `schedule_a` endpoint? It's significantly easier if the names match 1:1.

jontours:
I believe it's `contribution_receipt_date` on the regular schedule a endpoint as well.

noahmanger:
Ah ha. So it is using that value, but apparently it's null on the efile endpoint:
image

jontours:
I found that the receipt data for some reason was an integer on the api model (fixed and pushed to the repo), it's now working correctly and there are non-null `contribution_receipt_dates`. Be sure to pass in `sort_hide_null=true` as well.

noahmanger:
Ah there we go. Thanks!

jontours:
I took a look at the new filters and verified most work, but wanted to make note of something, or solicit feedback on these filters: `contributor_employer` and `contributor_occupation`.  For the processed these are tsvectors which are essentially vectorized in the database to enable quick search in the database.  It's easy to do that for the processed because we generate the tables and the columns therein.  For the raw resources it's not quite as simple.  However, to start cautiously I was able to add a column to raw resource that updates as rows are updated or inserted in the efiling table.  Since it's a big table and currently not sure how performance will be affected

jontours:
Also gonna leave a note here before I forget that paging doesn't seem to be working for the raw view.  HItting the button generates another http request, but doesn't add the `page=n` parameter.

noahmanger:
I noticed that too. Need to figure out what's going on there. It works for the filings tables, so not sure why it's not here.

noahmanger:
Ok, this is now ready for review. @xtine can you take a look?

xtine:
Looks good, nothing jumps out at me and it works.

PR #1777, opened 2017-01-04
Deleting enforcement landing template

noahmanger:
In order to allow the right type of page nesting, I moved this template to fec-cms, so this just deletes it from the web app.


ccostino:
@noahmanger, is the corresponding template in the CMS the [enforcement](https://github.com/18F/fec-cms/blob/develop/fec/home/templates/home/legal/enforcement.html) one?

noahmanger:
Exactly.

PR #1780, opened 2017-01-10
Feature/consume additional ao fields

anthonygarvan:
This adds UI and consumes the API for the additional AO fields.


openfecwebapp/api_caller.py
```diff
@@ -73,6 +74,19 @@ def load_legal_search_results(query, query_type='all', ao_no=None, ao_name=None,
...
+        if ao_is_pending:
+            filters['ao_is_pending'] = True
+
+        if ao_requestor:
+            filters['ao_requestor'] = ao_requestor
+
+        if ao_category:
+            filters['ao_category'] = ao_category
+
+        print(ao_requestor_type)
```
vrajmohan: Can be removed?

PR #1781, opened 2017-01-11
deploy disclaimer

LindsayYoung:
forewarned is forearmed

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,11 @@ Compile Sass as changes are made:
 
 We use Travis for automated deploys after tests pass. If you want to deploy something it is much better to push an empty commit with a tag than doing a manual deploy.
 
+    git checkout master
+    git commit --allow-empty -m 'Message indicating what the commit is for'
+    git tag -a -m 'Message about the tag' <tagname> <commit id generated above with commit above>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -130,7 +130,11 @@ Compile Sass as changes are made:
 
 ### Deployment
 
-*Likely only useful for 18F team members*
+*Likely only useful for 18F FEC team members*
+
+We use Travis for automated deploys after tests pass. If you want to deploy something it is much better to push an empty commit with a tag than doing a manual deploy.
```
ccostino: Should we include a line as an example, e.g.:

```bash
git checkout master
git commit --allow-empty -m 'Message indicating what the commit is for'
git tag -a -m 'Message about the tag' <tagname> <commit id generated above with commit above>
git push --follow-tags origin master
```

Does that look useful and right?

noahmanger: I like adding this example.

ccostino: @LindsayYoung, looks like this is still outstanding and we should probably get it merged - would adding this example be okay?

LindsayYoung: Yes looks good

PR #1788, opened 2017-01-13
Bugfix/check final opinion by default

anthonygarvan:
This also includes a bugfix that was breaking non-AO searches.


vrajmohan:
Looks good

PR #1789, opened 2017-01-13
Feature/raw ie data

noahmanger:
This is a duplicate of #1771 but into develop instead of the release. This has already been reviewed so just needs to be merged.


ccostino:
Thanks, @noahmanger!

PR #1791, opened 2017-01-14
[WIP] Front end MUR filters

dogweather:
This implements #1735


.python-version
```diff
@@ -0,0 +1 @@
+3.5.2
```
dogweather: @ccostino @anthonygarvan What do you think - can we lock this down? I think it'd provide some stability.

tests/unit/test_legal_search.py
```diff
@@ -69,7 +68,43 @@ def test_search_advisory_opinions(self, load_legal_search_results):
...
+                'search_type': 'murs',
+            }
+        )
+        _call_api.assert_called_once_with('legal',
+                                          'search',
+                                          from_hit=0,
+                                          hits_returned=20,
+                                          murs_no='6995',
+                                          q='presidents',
+                                          type='murs')
```
dogweather: I'm doing TDD so I don't have to manually check the API call. Also to allow refactoring. 

What do you all think of the test name? `api_invoked_correctly_for_mur` — I like to name tests w/ a little more info, to describe what they're testing for. (As opposed to, say, `test_mur`.)

dogweather:
Case Number text filter is working.

.python-version
ccostino: @dogweather I agree, but we need to see which version is supported in cloud.gov and also make sure Travis matches as well.  I'm running 3.5.2 locally right now (though 3.5.3 was just released) and have not run into any issues, but we need to ensure parity and support across all systems and environments we use.

I have asked to see what cloud.gov currently supports via the Cloud Foundry Python buildpack.  I know Travis supports 3.5+ too.  I do think it'd make more sense to make these changes in a separate PR, though, since there is a bit more work involved.

ccostino: Just confirmed that cloud.gov is using the latest Python buildpack (via the `cf buildpacks` command, which returns [1.5.14](https://github.com/cloudfoundry/python-buildpack/releases/tag/v1.5.14) and is the current latest release).

Like I said, I think it makes more sense to move this work out into its own PR.  How do others feel about that?

noahmanger:
Closing this since this will be implemented with React eventually.

PR #1798, opened 2017-01-17
Hide null amendment status versions

xtine:
This is a fix to not show amendment status versions if `most_recent: null`, because we are unsure if it is actually the most recent or past version.

screen shot 2017-01-17 at 2 43 44 pm

https://github.com/18F/openFEC/issues/2158


noahmanger:
We'll need to merge this into the release.

noahmanger:
@xtine do you need to pull in the change here for the build to pass?

PR #1812, opened 2017-01-29
[WIP] JS-enhanced legal filters

noahmanger:
I was curious to see how complicated it would be to get the javascript-enhanced filters going for the legal searches, so I started poking around and ended up getting a pretty decent proof-of-concept going in not-too-long. 

The main challenge with this code base is that it's really confusing how all the different components connect and fit together. So this PR is to get the basic pieces in place and show how they connect as more of a starting point for discussion as much as anything. I'm totally open to refactoring parts and also think it'd still be good to break things out into smaller tasks. But FWIW, here's what I threw together. 

image

The real heavy lifting here is done in a new [LegalSearch component](https://github.com/18F/openFEC-web-app/compare/feature/js-enhanced-filters?expand=1#diff-96a9a3ddc63afa9fccd5aaf15bfa659f). This component:
- Initializes the `FilterPanel` component, which in turn initializes each of the filters
- Initializes the `FilterTags` component, creating an area for the filter tags
- Calls the API and renders the results (which requires a new Handlebars template for the result)

Each filter is an instance of a `Filter` component, with different types for text, checkboxes, dropdowns, etc. These components:
- Update the count of active filters
- Trigger events that create tags and listen for the removal of tags. 

A few other things that I had to do:
- In order for `FilterPanel` to initialize individual filters, those divs need to have the `js-filter` class and then a `data-filter="[type of filter"]` attribute. This is built in to the text, checkbox and dropdown filter macros we use on the data table pages, so I just used those macros. 
- I moved the AO document types to constants.py, where we store the constants for other filters
- Tweaked the layout of the `legal-doc-search-results` template a bit to add a div for the filter tags

But it's still far from complete. Here's what I know needs to be done, and likely there's more:
- The rendering of the results isn't quite right. I got a basic handlebars template in place, but I know it's not consistent with the other template
- Improve the error/success/loading state stuff
- Get the pagination / results count component working

cc @anthonygarvan @xtine @dogweather . Hope this helps. Let's discuss how we want to move forward.


noahmanger:
Another thought on this, another way to go is to not worry about any of the AJAX fetching and rendering of results and only do use the accordions and possible the javascript enhanced text fields, which allow you to search for multiple terms by making checkboxes. 

You would still need to submit the form by clicking the button, but I _think_ it would allow us to implement filter panels as designed without accruing additional tech debt. Just a thought. Curious what others think.

static/js/modules/legal-search.js
```diff
@@ -0,0 +1,105 @@
...
+var URI = require('urijs');
+var _ = require('underscore');
+var urls = require('fec-style/js/urls');
+var FilterPanel = require('fec-style/js/filter-panel').FilterPanel;
+var filterTags = require('fec-style/js/filter-tags');
+
+var ao_template = require('../../templates/legal-search/ao.hbs');
+
+// Just lifted this from calendar-helpers.js
+function getUrl(path, params) {
```
dogweather: Are we using some kind of implicit namespacing system which prevents these from becoming globals?

openfecwebapp/constants.py
```diff
@@ -251,6 +251,39 @@
     ('S', 'Senate')
 ])
 
+ao_doc_types = OrderedDict([
```
dogweather: Anyone know why these constants aren't uppercase, in the usual Python style?

openfecwebapp/templates/legal-search-results-advisory_opinions.html
```diff
@@ -1,45 +1,55 @@
...
+    {{ text.field('ao_name', 'AO name', value=ao_name)}}
+    {{ text.field('ao_no', 'AO number', value=ao_no)}}
+    {{ legal.date_filter('ao_min_date', 'ao_max_date', 'Issued date', ao_min_date, ao_max_date)}}
+    {{ legal.checkbox_filter('ao_is_pending', 'Show only pending requests', ao_is_pending)}}
+  </div>
+  <button type="button" class="js-accordion-trigger accordion__button">Requestor</button>
+  <div class="accordion__content">
+    {{ text.field('ao_requestor', 'Requestor Name', value=ao_requestor)}}
+    {{ legal.dropdown('ao_requestor_type', 'Requestor Type',
+    [ {'value': 0, 'text': 'Any'},
```
dogweather: Can this be refactored to use the constant defined in Python? Just noticing that we've got this data structure in the code twice.

static/js/pages/legal.js
```diff
@@ -1,14 +1,10 @@
 'use strict';
 
 /* global require */
-require('jquery.inputmask');
-require('jquery.inputmask/dist/inputmask/inputmask.date.extensions.js');
 
 var $ = require('jquery');
-var FilterPanel = require('fec-style/js/filter-panel').FilterPanel;
+var LegalSearch = require('../modules/legal-search').LegalSearch;
```
dogweather: I'm guessing that this is how we avoid polluting the global namespace.

noahmanger:
Closing this since we're going another route with https://github.com/18F/openFEC-web-app/pull/1871

Issue #1614, opened 2016-09-28
Implement itemized debts page


noahmanger:
So that users see the debt obligations owed to or by committees, implement a data table page for itemized schedule D data.

---

[API endpoint.](https://api.open.fec.gov/developers/#!/party-coordinated_expenditures/get_schedules_schedule_d)

**Filters and columns:**
These are what I understand the filters and columns we need to be:
 
- Committee ID (typeahead filter)
- Creditor name (text filter)
- Purpose (column only)
- Date (date range without time-period select) (I think this is the date of the filing it was included on, but not positive)
- Amount paid this period (financial range filter)
- Amount incurred this period (financial range filter)

**Details panel:**
Include all the fields available on the Schedule D form:
![image](https://cloud.githubusercontent.com/assets/1696495/22170969/d7b31c7a-df39-11e6-9dde-e2b4e2e9c8ee.png)

The details panel should also link to the original image, just like receipts and disbursements.

**Completion criteria:**
- [x] Debts page available at `/debts`
- [x] Data is filterable by available filters
- [ ] Details panel contains additional data and links to original image
- [ ] Link to page is included on "Advanced data" page

noahmanger:
Updated the original issue with more details about filters and columns. @jwchumley @PaulClark2 can verify those are right (and @jontours can help with API questions).

xtine:
I have the columns working in a datatable, but currently none of the proposed filters are a parameter so I can't get the filters working right now. @jontours: can the filters be added as parameters?

jontours:
I can take a look.  There should be a couple of filters, but not all of them are implemented.

PaulClark2:
`Date (date range without time-period select) (I think this is the date of the filing it was included on, but not positive)`  -- This is correct. The date in the SD table is the report receipt date. It is not the date the debt or obligation was incurred.

PR #1824, opened 2017-02-06
Feature/debts page

xtine:
Itemized debts page for schedule D data. https://github.com/18F/openFEC-web-app/issues/1614

Filters are not working yet, waiting for API to update parameters so they can be used.

screencapture-127-0-0-1-3000-debts-1486346967646

### DIFF
--- a/static/js/modules/columns.js
+++ b/static/js/modules/columns.js
@@ -1,11 +1,11 @@
+'use strict';
+
 /**
 * modules/columns.js
 * -------------------
 * This file contains objects to support column data for datatable pages.
 *
 */
-
-'use strict';
 
 var _ = require('underscore');
 
--- a/static/js/modules/columns.js
+++ b/static/js/modules/columns.js
@@ -529,6 +529,28 @@ var reports = {
   }
 };
 
+var debts = [
+  {
+    data: 'committee',
+    orderable: false,
+    className: 'all column--large',
+  },
+  {
+    data: 'creditor_debtor_name',
+    orderable: false,
+    className: 'all column--large',
+  },
+  {
+    data: 'nature_of_debt',
+    orderable: false,
+    className: 'all column--med',
+  },
+  // dateColumn({data: 'date', className: 'min-tablet hide-panel column--med'}),
+  currencyColumn({data: 'amount_incurred_period', className: 'min-desktop hide-panel column--number'}),
+  currencyColumn({data: 'payment_period', className: 'min-desktop hide-panel column--number'}),
+  modalTriggerColumn
+];
+
 module.exports = {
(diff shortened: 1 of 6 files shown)


static/js/modules/columns.js
```diff
@@ -506,6 +513,31 @@ var reports = {
   }
 };
 
+var debts = [
+  {
+    data: 'committee',
+    orderable: false,
+    className: 'all column--large',
+    render: function (data, type, row) {
+      return row.committee.name;
```
noahmanger: This should build the link to the committee page.

static/js/modules/columns.js
```diff
@@ -506,6 +513,31 @@ var reports = {
...
+    data: 'creditor_debtor_name',
+    orderable: false,
+    className: 'all column--large',
+  },
+  {
+    data: 'nature_of_debt',
+    orderable: false,
+    className: 'all column--large',
+  },
+  dateColumn({data: 'load_date', orderable: true, className: 'min-tablet hide-panel column--med'}),
```
noahmanger: Not sure what value we should use for the date here, but I don't think it should be `load_date` since that's an internal value for when it was loaded into the db.

xtine: Hmm. There's no other date in the API. @jontours: is there any other date filter that can be added like `incurred_date`?

jontours: The only other pertinent date in the database not yet exposed is `receipt_dt`.  I'm not gonna assume that's the same as the date the debt was incurred, maybe we can verify with the FEC?

noahmanger: Paging @PaulClark2 : what's the logical value to show for a date column with debts? And what would that column be called in the database?

noahmanger: Circling back on this. We're just waiting on the proper date field to wrap this up. @PaulClark2 what date should we be using here?

PaulClark2: I'm going to need to follow-up with someone in RAD. There aren't any useful dates on the Schedule D. The receipt date in the SD table is the receipt date of the report.

jenniferthibault:
@xtine do you know what the status of this PR is and if the work has been transferred into the cms repo?

xtine:
@jenniferthibault: this is an iceboxed issue because it's reliant on API work that didn't get done for some reason or another. We can close this and just reference the PR work here it to the Icebox issue in `fec-cms` (or create one if it doesn't exist).

jenniferthibault:
Thanks Christine! I moved this to a new issue in the `fec-cms` repo: https://github.com/18F/fec-cms/issues/1631

Closing this PR

Issue #1613, opened 2016-09-28
Create page for itemized loans


noahmanger:
So that users see the loans made and received by committees, implement a data table page for itemized schedule C data.

---

[API endpoint.](https://api.open.fec.gov/developers/#!/party-coordinated_expenditures/get_schedules_schedule_c)

**Filters and columns:**
These are what I understand the filters and columns we need to be:
 
- Committee ID (typeahead filter)
- Loaner name (text filter)
- Date (date range without time-period select) (I think this is the date of the filing it was included on, but not positive)
- Payment to date (financial range filter)
- Original amount of loan  (financial range filter)

**Details panel:**
Include all the fields available on the [Schedule C form](https://cloud.githubusercontent.com/assets/1696495/22171021/e3c541e0-df3a-11e6-9cdd-fd27c0692750.png).

The details panel should also link to the original image, just like receipts and disbursements.

**Completion criteria:**
- [x] Loans page available at `/loans`
- [x] Data is filterable by available filters
- [x] Details panel contains additional data
- [x] Link to page is included on "Advanced data" page

noahmanger:
Updated the original issue with more details about filters and columns. @jwchumley @PaulClark2 can verify those are right (and @jontours can help with API questions).

noahmanger:
Resolved by #1828. Nice work @johnnyporkchops !

PR #1828, opened 2017-02-08
Feature/itemized loans

johnnyporkchops:
Itemized loans datatable 
Issue: https://github.com/18F/openFEC-web-app/issues/1613


openfecwebapp/templates/partials/itemized-loans-filter.html
```diff
@@ -0,0 +1,25 @@
...
+{% import 'macros/filters/ie-reports.html' as reports %}
+
+
+{% block filters %}
+<div class="js-accordion accordion--neutral" data-content-prefix="filter" data-open-first="true">
+  <button type="button" class="js-accordion-trigger accordion__button">Committee information</button>
+  <div class="accordion__content">
+  {{ typeahead.field('committee_id', 'Committee Name or ID') }}
+</div>
+  {{ text.field('loan_source_name', 'Loaner name') }}
```
xtine: Include the filters within the `js-accordion` block to give them the filter padding.

static/js/pages/itemized-loans.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+var $ = require('jquery');
+
+var tables = require('../modules/tables');
+var columns = require('../modules/columns');
+var helpers = require('../modules/helpers');
```
xtine: You can remove the line since there aren't any helper functions being called.

static/js/pages/itemized-loans.js
```diff
@@ -0,0 +1,23 @@
...
+
+$(document).ready(function() {
+  var $table = $('#results');
+  new tables.DataTable($table, {
+    autoWidth: false,
+    title: 'Itemized loans',
+    path: ['schedules','schedule_c'],
+    //path: ['loans', 'get_schedules_schedule_c'],
```
xtine: Make sure to remove these comments.

openfecwebapp/templates/partials/loans-filter.html
```diff
@@ -0,0 +1,23 @@
+{% extends 'partials/filters.html' %}
+
+{% import 'macros/filters/text.html' as text %}
+{% import 'macros/filters/states.html' as states %}
+{% import 'macros/filters/typeahead-filter.html' as typeahead %}
+{% import 'macros/filters/years.html' as years %}
```
noahmanger: Don't need the years macro.

openfecwebapp/templates/partials/loans-filter.html
```diff
@@ -0,0 +1,23 @@
...
+{% import 'macros/filters/date.html' as date %}
+{% import 'macros/filters/range.html' as range %}
+{% import 'macros/filters/ie-reports.html' as reports %}
```
noahmanger: Don't need ie-reports macro

openfecwebapp/templates/partials/loans-filter.html
```diff
@@ -0,0 +1,23 @@
+{% extends 'partials/filters.html' %}
+
+{% import 'macros/filters/text.html' as text %}
+{% import 'macros/filters/states.html' as states %}
```
noahmanger: Don't need states macro.

static/js/modules/columns.js
```diff
@@ -506,6 +506,26 @@ var reports = {
   }
 };
 
+var loans = [
+  {
+    data: 'committee',
+    orderable: false,
+    className: 'all column--large',
+    render: function (data, type, row) {
```
noahmanger: This `render` function should also build a link to the page, like the other tables do:
```js
      if (data) {
        return columnHelpers.buildEntityLink(
          data.name,
          helpers.buildAppUrl(['committee', data.committee_id]),
          'committee'
        );
      } else {
        return '';
      }
```

static/js/modules/columns.js
```diff
@@ -506,6 +506,26 @@ var reports = {
...
+      return row.committee.name;
+    }
+  },
+  {
+    data: 'loan_source_name',
+    orderable: false,
+    className: 'all column--large',
+  },
+  dateColumn({data: 'load_date', orderable: true, className: 'min-tablet hide-panel column--med'}),
```
noahmanger: I'm not sure what the value for the `date` column should be, but I don't think `load_date` is right, as that's just about when it was loaded into the system. @PaulClark2, when displaying a table of itemized schedule Cs, should we show a date? And if so, the date of what?

noahmanger (changes requested):
Looking good. Just had a few notes and questions.

noahmanger:
Also, I added docs here to explain how to get the details panel working. https://github.com/18F/openFEC-web-app/wiki/Datatable-Page-Setup#create-a-details-panel-template

xtine:
I got the details panel working on my Debts PR (https://github.com/18F/openFEC-web-app/pull/1824) for reference. /cc @patphongs

static/js/modules/columns.js
PaulClark2: This is for the details panel, right? There are two dates we should probably show if there is room, `Date Incurred` and 'Date Due`.

noahmanger: What about in the table? Is there one that makes more sense to use for sorting?

PaulClark2: `Date Incurred`

johnnyporkchops: There does not seem to be a Date Due field in the data for itemized-loans

noahmanger (approved):

PR #1842, opened 2017-02-16
Fixed typo

jontours:
Easy for review.


jontours:
Candidate name is still not showing on staging so not sure if this fix is enough.

PR #1846, opened 2017-02-17
Use refactored AO index with 1 ES doc per AO

vrajmohan:


anthonygarvan:
omg so excited about this. So much crappy code removed!

PR #1849, opened 2017-02-24
Feature/auto deploy govcloud

LindsayYoung:
So we can have auto deploys on the govcloud branch without integrating it into our next release


ccostino:
Thank you, @LindsayYoung!  Giving it a shot. 🚀

Issue #1644, opened 2016-10-20
Add the ability to browse previous two-year periods on overview charts 


noahmanger:
So that users can view data for earlier elections, add the ability to page back in time to earlier two-year periods.

---

This is a carry-over issue from the original https://github.com/18F/openFEC-web-app/issues/1603

We're going to release this without the ability to page back in time, but will want to add it eventually.

---

## Update
Now that we're almost to the end of the year, I think we should implement this so that people can go back to the previous cycle once we're in a new one.

**Completion criteria:**
- [x] Using the arrows to go before January of the first year of a range redraws the chart with data from the previous 2 year cycle
- [x] Using the right arrow to go after the final December redraws the chart with data from the next 2 year cycle

noahmanger:
I'd like to bump this into the next sprint since we're entering a new two-year period.

noahmanger:
Don't need to turn on 2018 data until February.

PR #1853, opened 2017-02-25
Feature/overview chart enhancement

noahmanger:
This adds the ability to navigate between cycles on the line charts. Now, if you're on the last month of a cycle and you use the arrow buttons to go back in time, it destroys the chart and redraws it with data for the previous cycle; and likewise for arrowing past the next cycle. 

image

A few details:
- I moved the data fetching to the `LineChart` component, rather than passing the processed data into the constructor, since it needs to refetch data for each cycle.
- It uses the global constant `DEFAULT_TIME_PERIOD` for the starting cycle, and the `MAX_YEAR` constant for determining the furthest in the future you can go. It sets a `MIN_YEAR` constant for the earliest cycle we have data for. 
- I also set a static Y-axis max of $4 billion. Previously, the y-max was set based on the data, but this has the side-effect of distorting comparisons year-to-year (or between raising and spending). This makes those comparisons easier, but it means that it may need to be adjusted in the future. If there's a better way to do this, I'm all ears. 

Resolves https://github.com/18F/openFEC-web-app/issues/1644

### DIFF
--- a/static/js/modules/line-chart.js
+++ b/static/js/modules/line-chart.js
@@ -15,6 +15,7 @@ var bisectDate = d3.bisector(function(d) { return d.date; }).left;
 
 var MIN_CYCLE = 2008;
 var MAX_CYCLE = END_YEAR;
+var MAX_RANGE = 4000000000; // Set the max y-axis to 4 billion
 
 function LineChart(selector, snapshot, type, index) {
   this.element = d3.select(selector);
@@ -88,7 +89,7 @@ LineChart.prototype.fetch = function(cycle) {
 LineChart.prototype.buildChart = function(chartData) {
   var data = chartData;
   var entityTotals = {};
-  var max = 0;
+  var max = MAX_RANGE;
 
   data.forEach(function(d) {
     d.date = new Date(d.date);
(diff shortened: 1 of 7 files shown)


xtine:
Overall, works great. I just have one suggestion. When I click the back arrow to go back a full cycle (like Jan 1, 2013 - Dec 31, 2014) and then click the forward arrow to redraw the chart to the next cycle (Jan 1, 2015 - Dec 31, 2016), the line is at the end of the cycle. It would make more sense to draw the line starting at beginning rather than the end of the chart.

sotz8jf18c

noahmanger:
Good idea! Just added that.

static/js/modules/line-chart.js
```diff
@@ -176,61 +259,105 @@ LineChart.prototype.formatXAxis = function(axis) {
...
     })
     .attr('r', 4);
 };
 
-LineChart.prototype.populateSnapshot = function(d) {
+LineChart.prototype.getCursorStartPosition = function() {
+  // Determines whether to start the cursor at the begining or end of a time period
+  // this.startCursorAtEnd is set to true by default, but when navigating
+  // to next cycle, it is set to false so that the cursor starts at the beginning
+  return this.startCursorAtEnd === true ? this.chartData[this.chartData.length - 1] : this.chartData[0];
```
xtine: For readability can you break out this conditional operator?

PR #1861, opened 2017-02-28
retaining full state on pagination

anthonygarvan:
oops. Never implemented pagination for all the new filters.


vrajmohan:
We should do something soon about the growing method signature.

Issue #1864, opened 2017-03-01
election pages - future Senate electons


PaulClark2:
Senate Class 1 is up for election in 2018. The election pages for classes 2 and 3 are confusing.

Class 2: https://beta.fec.gov/data/elections/senate/IL/2020/
Class 3: https://beta.fec.gov/data/elections/senate/AZ/2022/

![senate-class2-2020-election-page](https://cloud.githubusercontent.com/assets/11817734/23460807/255e532e-fe55-11e6-8fe0-5ac4f6e6168b.png)

![senate-class3-2022-election-page](https://cloud.githubusercontent.com/assets/11817734/23460811/29b2f966-fe55-11e6-92bc-0e59ff6d2c13.png)

Also see https://github.com/18F/openFEC-web-app/issues/1863

noahmanger:
Resolved by https://github.com/18F/openFEC-web-app/pull/1865

Issue #1863, opened 2017-03-01
election pages -- invalid election years for President and Senate


PaulClark2:
...data/elections/president/ allows users to enter invalid presidential election years. 

We should not allow a 2018 presidential election page. The dates displayed on the page, as it exists now, are confusing. 

https://beta.fec.gov/data/elections/president/2018/
![president-election-page](https://cloud.githubusercontent.com/assets/11817734/23460331/e07c10c2-fe52-11e6-8f1c-4ca7e9707d2a.png)

PaulClark2:
...data/elections/senate/[STATE]/ allows users to enter invalid Senate election years. 

https://beta.fec.gov/data/elections/senate/MO/2020/ MO does not have a Class 2 (2020) Senate election.

The dates displayed on the page are confusing.

![senate-class2-invalid-election-2020-page](https://cloud.githubusercontent.com/assets/11817734/23460504/bd462d30-fe53-11e6-9070-8255a1b3dc4c.png)

PaulClark2:
Also see https://github.com/18F/openFEC-web-app/issues/1864

noahmanger:
Resolved by https://github.com/18F/openFEC-web-app/pull/1865

PR #1865, opened 2017-03-02
Dynamically generate list of election cycles

noahmanger:
This fixes several of the election cycle weirdness that @paulclark2 noticed in https://github.com/18F/openFEC-web-app/issues/1863 and https://github.com/18F/openFEC-web-app/issues/1864.

There's two issues. First, the cycle select currently maxes out at the current cycle (in this case, 2018). So even if a user navigated to a future election page, the select wouldn't match the other dates on the page: 

image

This patch makes it so that if the cycle entered in the URL path is greater than the current cycle, then the select will build a range up to that point (otherwise it will max out at the current cycle):

image

Second, it checks the cycle for presidential pages so that if it's not a valid presidential year, it throws a 404. 

Third, it adds a `SENATE_CLASSES` constant and checks that when on a state's page, that the year in the URL path is actually a year when one of this state's classes is up for election, otherwising throwing a 404. And also now only shows valid elections:
image

This does not resolve the issue of incorrect election dates showing up in election search results, as those are handled by the API.

@PaulClark2 do you think this resolves those issues?

### DIFF
--- a/openfecwebapp/utils.py
+++ b/openfecwebapp/utils.py
@@ -107,8 +107,9 @@ def date_ranges():
         ),
     }
 
-def get_cycles():
-    return range(current_cycle(), constants.START_YEAR, -2)
+def get_cycles(max_cycle=None):
+    max = max_cycle if max_cycle else current_cycle()
+    return range(max, constants.START_YEAR, -2)
 
 def election_title(cycle, office, state=None, district=None):
     base = ' '.join([str(cycle), 'Election', 'United States', office.capitalize()])
--- a/openfecwebapp/utils.py
+++ b/openfecwebapp/utils.py
@@ -167,3 +167,21 @@ def process_cash_data(totals):
 
 def process_ie_data(totals):
     return financial_summary_processor(totals, constants.IE_FORMATTER)
+
+
+def get_senate_cycles(senate_class):
+    if senate_class == 1:
+        max = 2018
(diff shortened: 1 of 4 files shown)


jontours:
@noahmanger is the list, for example, of a state's valid senatorial cycles are currently generated from the `election\search` api endpoint?  Trying to keep abreast of all the ways that endpoint is used.

noahmanger:
@jontours this PR just introduces some new constants (and utils that use them) to generate lists of cycles for each class, and then associates classes with states. It could be something provided by the API though.

jontours:
Ah, gotcha.  I'll see if it is something simple to do given the resources we have (I think it would be).

openfecwebapp/utils.py
```diff
@@ -166,3 +167,16 @@ def process_cash_data(totals):
...
+def get_senate_cycles(senate_class):
+    next_election = constants.NEXT_SENATE_ELECTIONS[str(senate_class)]
+    return range(next_election, constants.START_YEAR, -6)
+
+def get_state_senate_cycles(state):
+    senate_cycles = []
+    for senate_class in [1, 2, 3]:
+        if state.upper() in constants.SENATE_CLASSES[str(senate_class)]:
+            senate_cycles += get_senate_cycles(senate_class)
+    print(senate_cycles)
```
xtine: rm print statement

Issue #1860, opened 2017-02-28
Show itemized disbursements on candidate page


noahmanger:
So that users can see how candidates spend their money, show itemized disbursements from all authorized committees on the candidate page. 

---

![image](https://cloud.githubusercontent.com/assets/1696495/23385925/230094dc-fd08-11e6-8214-17ce1fa4c992.png)

**Completion criteria:**
- [x] Show which committees the disbursements are from
- [x] Allow users to switch 2-year periods (we can't support the full-cycle view yet)
- [x] Filter `/schedules/schedule_b/` for all results in the current time period by the candidate's authorized committees

noahmanger:
Resolved by https://github.com/18F/openFEC-web-app/pull/1866

PR #1866, opened 2017-03-03
Feature/new candidate disbursements and about candidate

xtine:
Section on new candidate page for itemized disbursements. https://github.com/18F/openFEC-web-app/issues/1860

screencapture-127-0-0-1-3000-candidate-h0ca32101-new-1488502610815

Also added "About this candidate" tab.

screencapture-127-0-0-1-3000-candidate-s2wi00219-new-1489548392859

Need styles for about tab: https://github.com/18F/fec-style/pull/647

### DIFF
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -220,7 +220,7 @@ def load_candidate_totals(candidate_id, cycle, election_full=True):
     return {}
 
 
-def load_candidate_statement_of_candidacy(candidate_id, cycle, election_full=False):
+def load_candidate_statement_of_candidacy(candidate_id, cycle):
     response = _call_api(
         'filings',
         candidate_id=candidate_id, cycle=cycle, form_type='F2'
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -179,7 +179,6 @@ def load_single_type(data_type, c_id, *path, **filters):
     data = _call_api(data_type, c_id, *path, **filters)
     return result_or_404(data)
 
-
 def load_nested_type(parent_type, c_id, nested_type, *path, **filters):
     return _call_api(parent_type, c_id, nested_type, *path, per_page=100, **filters)
 
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -228,7 +228,7 @@ def load_candidate_statement_of_candidacy(candidate_id, cycle):
 
     if response['results']:
         return response['results'][:2]
-    return {}
+    return None
 
 
 def result_or_404(data):
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -221,16 +221,10 @@ def load_candidate_totals(candidate_id, cycle, election_full=True):
 
 
 def load_candidate_statement_of_candidacy(candidate_id, cycle, election_full=False):
-    if election_full:
-        response = _call_api(
-            'filings',
-            candidate_id=candidate_id, form_type='F2'
-        )
-    else:
-        response = _call_api(
-            'filings',
-            candidate_id=candidate_id, cycle=cycle, form_type='F2'
-        )
+    response = _call_api(
+        'filings',
+        candidate_id=candidate_id, cycle=cycle, form_type='F2'
+    )
 
     if response['results']:
         return response['results'][:2]
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -220,6 +220,23 @@ def load_candidate_totals(candidate_id, cycle, election_full=True):
     return {}
 
 
+def load_candidate_statement_of_candidacy(candidate_id, cycle, election_full=False):
+    if election_full:
+        response = _call_api(
+            'filings',
+            candidate_id=candidate_id, form_type='F2'
+        )
+    else:
+        response = _call_api(
+            'filings',
+            candidate_id=candidate_id, cycle=cycle, form_type='F2'
+        )
+
+    if response['results']:
+        return response['results'][:2]
+    return {}
+
+
 def result_or_404(data):
     if not data.get('results'):
         abort(404)
--- a/openfecwebapp/templates/candidates-single-new.html
+++ b/openfecwebapp/templates/candidates-single-new.html
@@ -47,19 +47,28 @@
           <a
             class="side-nav__link"
             role="tab"
-            data-name="other-spending"
+            data-name="about-candidate"
             tabindex="1"
             aria-controls="panel2"
-            href="#section-2">Spending by others to support or oppose</a>
+            href="#section-2">About this candidate</a>
+        </li>
+        <li class="side-nav__item" role="presentation">
+          <a
+            class="side-nav__link"
+            role="tab"
+            data-name="other-spending"
+            tabindex="2"
+            aria-controls="panel3"
+            href="#section-3">Spending by others to support or oppose</a>
         </li>
         <li class="side-nav__item" role="presentation">
           <a
             class="side-nav__link"
             role="tab"
             data-name="itemized-disbursements"
-            tabindex="2"
-            aria-controls="panel3"
-            href="#section-3">Itemized Disbursements</a>
+            tabindex="4"
+            aria-controls="panel4"
+            href="#section-4">Itemized Disbursements</a>
         </li>
       </ul>
     </nav>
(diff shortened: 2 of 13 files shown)


xtine:
Paging doesn't seem to work. @jontours can you double check if the `page` filter is working properly on the API for `schedule_b`?

xtine:
Fixed the paging issue by setting `paginator: tables.SeekPaginator` in the datatable so it uses indexes instead of page filter.

noahmanger:
Is this ready for review @xtine ?

xtine:
@noahmanger: this is ready now. I had a bit of difficulty trying to figure out the "filter this data" button for multiple committees since the `{{ url_for }}` helper has the limitation of unique filters so I can't pass more than one committee. But since most candidates I've seen usually show one committee I've set it so if there's just one it shows the button and doesn't if there are multiple for now.

openfecwebapp/templates/candidates-single-new.html
```diff
@@ -52,6 +52,15 @@ <h1 class="entity__name">{{ name }}</h1>
             aria-controls="panel2"
             href="#section-2">Spending by others to support or oppose</a>
         </li>
+        <li class="side-nav__item" role="presentation">
+          <a
+            class="side-nav__link"
+            role="tab"
+            data-name="itemized-disbursements"
+            tabindex-"1"
```
noahmanger: Should be `=` not `-`.

openfecwebapp/templates/partials/candidate/itemized-disbursements.html
```diff
@@ -0,0 +1,68 @@
...
+{% import 'macros/null.html' as null %}
+{% import 'macros/cycle-select.html' as select %}
+
+<section id="section-3" role="tabpanel" aria-hidden="true" aria-labelledby="section-1-heading">
+  <div class="heading--section">
+    <h2>Itemized Disbursements</h2>
+  </div>
+
+  <div class="content__section">
+    {{ select.cycle_select(election_years, cycle, duration=duration, id="cycle-1") }}
```
noahmanger: The cycle select shouldn't have the option to do "full cycle" since we only have up to two years data at a time:
image

static/js/pages/candidate-single.js
```diff
@@ -55,26 +59,51 @@ var electioneeringColumns = [
...
+        helpers.buildAppUrl(['committee', row.committee_id]),
+        'committee'
+      );
+    }
+  },
+  {
+    data: 'recipient_name',
+    className: 'all',
+    orderable: false,
+    defaultContent: 'NOT REPORTED'
```
noahmanger: We haven't been using `defaultContent` elsewhere, so I think best to just leave it blank.

static/js/pages/candidate-single.js
```diff
@@ -94,10 +123,58 @@ var tableOpts = {
...
+    columns: filingsColumns,
+    order: [[2, 'desc']],
+    dom: tables.simpleDOM,
+    pagingType: 'simple',
+    hideEmpty: true
+  });
+}
+
+function initDisbursementsTable() {
+  var $table = $('table[data-type="itemized-disbursements"]');
```
noahmanger: Can we set the default per-page to just 10? 30 on these pages feels like a lot.

noahmanger (changes requested):
Looking good, but caught a few things.

openfecwebapp/templates/partials/candidate/itemized-disbursements.html
xtine: Hm, should we should remove the whole Time period section? I guess it doesn't make sense because you can choose the year from the Election dropdown anyway.

noahmanger: We could try just make the select two-year periods. Though I'm not sure how this will play with the rest of the page and the "Full election" parameter that's set.

xtine: Do you mean the dropdown showing two year periods?

noahmanger: Yep.

xtine: screen shot 2017-03-08 at 4 42 31 pm

openfecwebapp/templates/macros/cycle-select.html
```diff
@@ -1,4 +1,4 @@
-{% macro cycle_select(cycles, cycle=none, duration=2, location='query', id='cycle-select', class='') %}
+{% macro cycle_select(cycles, cycle=none, duration=2, location='query', id='cycle-select', class='', show_two_year_period=False) %}
```
noahmanger: This isn't quite right, but I think I did a bad job explaining. It should show all two-year periods, not just the election years. So for a candidate who ran in 2010 and 2016, it should show: 2009-2010, 2011-2012, 2013-2014, 2015-2016. This set of years is available in the `cycles` template var. 

The `committee_cycle_select` macro below does this for committee pages, so maybe it makes sense to repurpose it for this case. 

I'm down to pair on this too, as the logic gets real wonky real fast.

xtine:
Even after following down a wrong rabbit hole, I figured out how to show all two-year periods for candidates. At first I had all two-year periods for a candidate selectable, but then I would still get 404 pages for some years. Ended up that I still had the `election-full=True` filter set so it would not give any results in the candidate history API even though it was still a valid two year period. So just setting `election-full=True` on any disbursement two year period selections would then give a non-empty result.

screen shot 2017-03-09 at 7 58 19 pm

openfecwebapp/templates/partials/candidate/about-candidate.html
```diff
@@ -0,0 +1,93 @@
...
+    </div>
+
+    <div class="entity__figure row">
+
+      <h3>Candidate information</h3>
+
+      <table class="t-sans usa-width-three-fourths">
+        <tr>
+          <td class="figure__label">Seat:</td>
+          <td class="figure__value">{{ office_full }}</td>
```
noahmanger: For house races this should also include `district`.

noahmanger (commented):
I'm getting an error when trying to change the cycle on [McCain's page here](http://localhost:3000/candidate/P80002801/new/?cycle=2008&election_full=true&tab=itemized-disbursements)  . Do you know what's up with that?

xtine:
@noahmanger: it's another issue when the disbursements page selects a two-year period that isn't in an election year.

openfecwebapp/templates/partials/candidate/itemized-disbursements.html
```diff
@@ -0,0 +1,70 @@
...
+        </li>
+      </ul>
+
+      <button type="button" class="js-export button button--cta button--export" data-export-for="itemized-disbursements">Export</button>
+    </div>
+
+    <table
+        class="data-table data-table--heading-borders"
+        data-type="itemized-disbursements"
+        data-committee-id="{% for c in committee_groups['P'] | reverse %}{{ c.committee_id }},{% endfor %}{% for c in committee_groups['A'] | reverse %}{{ c.committee_id }},{% endfor %}"
```
xtine: Just realized that there's already `committee_ids` coming from the view, so I just made a simpler loop that also works for the Filter this data button url as well.

xtine:
@noahmanger: this is ready for another review. not sure why the build isn't passing because the python unit tests pass for me locally but on travis it's failing at ao landing page?

openfecwebapp/api_caller.py
```diff
@@ -221,6 +220,17 @@ def load_candidate_totals(candidate_id, cycle, election_full=True):
...
 
 
+def load_candidate_statement_of_candidacy(candidate_id, cycle):
+    response = _call_api(
+        'filings',
+        candidate_id=candidate_id, cycle=cycle, form_type='F2'
+    )
+
+    if response['results']:
+        return response['results'][:2]
```
noahmanger: I'm not sure if we want to show just the first two or show all. Thinking...

noahmanger (commented):
Looking good. The only thing is about showing the Form 2s. I think rather than hard-code the versions in and only showing the two most recent, we should show _all_ for the selected election and show their actual version from the API. What do you think?

openfecwebapp/templates/partials/candidate/about-candidate.html
```diff
@@ -0,0 +1,145 @@
...
+          <td class="figure__value">{{ party_full|lower|title }}</td>
+        </tr>
+        {% if statement_of_candidacy %}
+        <tr>
+          <td class="figure__label">Statement of candidacy:</td>
+          <td class="figure__value">
+            <ul>
+            {% for statement in statement_of_candidacy %}
+              <li>
+                {% if loop.index == 1 %}
```
xtine: It's already ordered by date, so the latest one will be the most recent.

In regards to showing all of the forms for the election, some of them have like 10+ revised statements of candidacy for an election and can really bog down that section.

noahmanger: Ah that makes sense.

noahmanger (approved):

PR #1868, opened 2017-03-03
Adding the form numbers to the form type filter

noahmanger:
Responding to a bit of feedback that came in https://github.com/18F/FEC/issues/1135 , which also echoed earlier feedback, this adds form numbers to the form type dropdown on the filings page. 

image

This seems like a sensible change that will help experienced users disambiguate the filter options easier. I just did it because it only took two minutes to do.


jenniferthibault:
What do you think about not abbreviating so that it feels more approachable? I.e., `(Form 3)` instead of `(F3)`

noahmanger:
I like it. 
image

PR #1870, opened 2017-03-03
Removing presidential results that with incorrect dates

noahmanger:
This is a temporary hack to remove presidential elections from the election lookup tool on non-presidential years. Eventually this will be handled by the API but in the mean time this will get rid of a glaring error.

image

image

### DIFF
--- a/static/js/modules/election-lookup.js
+++ b/static/js/modules/election-lookup.js
@@ -212,6 +212,18 @@ ElectionLookup.prototype.serialize = function() {
   return _.extend(helpers.filterNull(params));
 };
 
+ElectionLookup.prototype.removeWrongPresidentialElections = function(results, cycle) {
+  // Hack to remove the presidential result in non-presidential years
+  // Eventually this will be handled by the API
+  if (Number(cycle) % 4 > 0) {
+    return _.filter(results, function(result) {
+      return result.office !== 'P';
+    });
+  } else {
+    return results;
+  }
+};
+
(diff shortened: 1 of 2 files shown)


jwchumley:
Thanks, Noah!


________________________________

static/js/modules/election-lookup.js
```diff
@@ -212,6 +212,18 @@ ElectionLookup.prototype.serialize = function() {
   return _.extend(helpers.filterNull(params));
 };
 
+ElectionLookup.prototype.removeWrongPresidentialElections = function(results, cycle) {
+  // Hack to remove the presidential result in non-presidential years
+  // Eventually this will be handled by the API
+  if (Number(cycle) % 4 > 0) {
```
ccostino: Good calling that out, thanks!

tests/unit/modules/election-lookup.js
```diff
@@ -176,4 +176,13 @@ describe('election lookup', function() {
...
   });
+
+  it('removes incorrect presidential elections', function() {
+    var raw = [
+      {cycle: 2018, office: 'P'},
+      {cycle: 2018, office: 'S'}
+    ];
+    var results = this.el.removeWrongPresidentialElections(raw, '2018');
+    expect(results).to.deep.equal([{cycle: 2018, office: 'S'}]);
+  });
```
ccostino: Thanks for including a test, too!

PR #1871, opened 2017-03-03
[WIP] Feature/add citations filter

anthonygarvan:
This introduces React into legal search, as well as introducing the citations filter.


static/js/legal/FilterPanel.js
```diff
@@ -0,0 +1,32 @@
...
+  this.setState({'expanded': !this.state.expanded});
+}
+
+render() {
+    return <li>
+      <button className="accordion__button" aria-controls={this.props.id} aria-expanded={this.state.expanded}
+      onClick={this.toggle}>
+        {this.props.header}
+      </button>
+      <div className="accordion__content" id={this.props.id} aria-hidden={!this.state.expanded} style={{display: this.getDisplay()}}>
```
shawnbot: I would suggest using the [`hidden` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) instead of `aria-hidden="true"` and `style="display: none"` to hide it. See our [accessibility guide](https://pages.18f.gov/accessibility/hidden-content/) for more info.

static/js/legal/Filters.js
```diff
@@ -0,0 +1,74 @@
...
+            <TextFilter key="ao_no" name="ao_no" label="AO number" value={props.query.ao_no}
+                handleChange={props.setQuery} getResults={props.getResults} />
+            <TextFilter key="ao_requestor" name="ao_requestor" label="Requestor Name (or AO Name)" value={props.query.ao_requestor}
+                  handleChange={props.setQuery} getResults={props.getResults} />
+              <Dropdown key="ao_requestor_type" name="ao_requestor_type" label="Requestor Type" value={props.query.ao_requestor_type}
+                options={requestorOptions} handleChange={props.instantQuery} />
+            <ul className="accordion--neutral" data-content-prefix="first">
+              <FilterPanel id="first-content-0" header="Documents" startOpen={true}
+                  content={[<TextFilter key="q" name="q" label="Keywords" value={props.query.q}
+                  handleChange={props.setQuery} getResults={props.getResults} />,
```
toolness: I noticed this `<FilterPanel>` is self-closing, yet there's a `content` prop that might be more readable if it were supplied as children.

For a really nice tutorial on React children, see [A deep dive into children in React](http://mxstbr.blog/2017/02/react-children-deepdive/).  That post is long but you only have to read the first few paragraphs to figure out how to use children.

static/js/legal/Filters.js
```diff
@@ -0,0 +1,74 @@
...
+                <CheckboxFilter key="ao_is_pending" name="ao_is_pending" label="Show only pending requests"
+                    checked={props.query.ao_is_pending} handleChange={props.instantQuery} />,
+                <CheckboxList key="ao_category" name="ao_category" label="Document Type" value={props.query.ao_category || ['F']}
```
toolness: These `<CheckboxList>` components seem to be at a higher level of indentation as the `<FilterPanel>`, which made me think they were children of it, but they seem to be siblings... maybe reduce their indentation level?

static/js/legal/Filters.js
anthonygarvan: Thanks!

PR #1875, opened 2017-03-07
Fixes rendering on ao landing page

noahmanger:
https://fec-stage-proxy.18f.gov/data/legal/advisory-opinions/ is currently failing because this view and template hadn't been refactored to account for the new AO schema. This makes the necessary fix. 

cc @vrajmohan


vrajmohan:
Good catch!

Issue #1859, opened 2017-02-28
Show itemized individual contributions on candidate page


noahmanger:
So that users can easily see who's given to a candidate, show their individual contributions given to all of the candidate's authorized committees on their candidate page.  

---
![image](https://cloud.githubusercontent.com/assets/1696495/23386085/4d8a906c-fd09-11e6-9326-f6fabfdb6180.png)

![image](https://cloud.githubusercontent.com/assets/1696495/23385838/6a7f6e56-fd07-11e6-9a58-8c2e360e9226.png)

In addition to showing contributions [aggregated by state and size](https://github.com/18F/openFEC-web-app/issues/1858) this issue is to show the raw data itself. 

**Completion criteria:**
- [x] Show the committees that this is data for
- [x] User can toggle between two-year period (though not full-cycle, as you can't get that many years at once for schedule A and B)
- [x] User can toggle to "All transactions"
- [x] Results from `/schedules/schedule_a/` filtered down to individual contributions received during the time period for all authorized committees for this candidate in the time period.

noahmanger:
Resolved by https://github.com/18F/openFEC-web-app/pull/1876

Issue #1858, opened 2017-02-28
Show individual contributions by size and state on new candidate page


noahmanger:
So that users can know where a candidate's support comes from, show their contributions aggregated by **state** and **size** on their page. 

---
[InVision](https://gsa.invisionapp.com/share/ESA8FN046#/218195098_Candidate_-_Indiv_Contribs)

This is to implement one aspect of the "individual contributions" tab on the new candidate pages. It's basically the same as it is on committee pages, except only the "state" and "size" tabs:

![image](https://cloud.githubusercontent.com/assets/1696495/23385679/55a32096-fd06-11e6-857d-092c9f750e41.png)

![image](https://cloud.githubusercontent.com/assets/1696495/23385714/8bf3e8ec-fd06-11e6-8a14-80a013d0bfbd.png)

These are the endpoints to use:
- `/schedules/schedule_a/by_state/by_candidate/`
- `/schedules/schedule_a/by_size/by_candidate/`

**Completion criteria:**
- [x] Users can toggle between "by state" and "by size"
- [x] Contributions by state map implemented: 
  - [x] Clicking a state highlights the row in the table and vice versa
  - [x] Clicking the total number opens up `/receipts/individual-contributions/` filtered to the time period and the candidate's authorized committees (see implementation on the election pages) 
- [x] Contributions by size table implemented:
  - [x] the total number opens up `/receipts/individual-contributions/` filtered to the time period and the candidate's authorized committees (see implementation on the election pages)

noahmanger:
Resolved by https://github.com/18F/openFEC-web-app/pull/1876

Issue #1852, opened 2017-02-24
Committee Cycle Bug


patphongs:
I believe there are multiple instances of this bug, so if this logic were fixed, I think it will fix a lot of committee broken links. This is related to this issue #1845 

This bug seems to happen when `election_full=true`. If the committee filing does not exist for a given year (whether it be because they were terminated or has not filed yet), the `url_for` function needs to have logic to accommodate for those particular instances. Right now, I believe it is just taking the submitted cycle parameter as is with no additional logic? cc @jwchumley @PaulClark2 @LindsayYoung @ccostino 

# Terminated Committee Example
## GARN, E. J. JAKE
Link: https://beta.fec.gov/data/candidate/S4UT00027/?cycle=1986&election_full=true
### Campaign Committees
GARN CAMPAIGN COMMITTEE links for Money raised, Money spent, Reports filed are all broken because it is trying to pull a non-existent cycle (1986), even though the committee was terminated in 1984. 

### Joint fundraising committees
GARN/CFTR RECEPTION COMMITTEE link is broken because it is trying to pull a non-existent cycle (1986), my assumption is that this committee was also terminated in 1984.

# Not Yet Filed Committee Example
## COCHRAN, THAD
https://beta.fec.gov/data/candidate/S8MS00055/?cycle=2020&election_full=true

### Campaign Committees
CITIZENS FOR COCHRAN link is broken because it is trying to pull 2020 cycle information for the committee, but they have not filed yet for that cycle.

LindsayYoung:
What if for the campaign committee links on a full cycle candidate page, we don't add the cycle to the call that will take you to the most recent records for the committee? I think that might be the quickest cleanest fix for now.

It might be awkward if you are looking back in time because it may jump you forward but, if our timing is clearly labeled, it should be something people can deal with and get where they want to go.

PaulClark2:
I don't think this is going to fix the problem.

If I'm on a candidate page, https://beta.fec.gov/data/candidate/S8MS00055/?cycle=2020&election_full=true, then I click on "Money raised" https://beta.fec.gov/data/committee/C00091892/?cycle=2020&tab=receipt, I get a response that doesn't make sense. I get an "Oops!" message. 

Cochran has raised money during the period between 2015 and 2020 (technically 2015 and 2017).  

This isn't ok.  His committee has just over $1 million dollars in receipts.

LindsayYoung:
[...]

PR #1876, opened 2017-03-08
Feature/new candidate contributions

xtine:
Individual contributions for new candidate pages grouped by all transaction, contribution size, and contributor state.
https://github.com/18F/openFEC-web-app/issues/1859
https://github.com/18F/openFEC-web-app/issues/1858

screen shot 2017-03-08 at 12 27 06 pm

---

screen shot 2017-03-08 at 12 27 14 pm

---

screen shot 2017-03-08 at 12 27 19 pm

### DIFF
--- a/openfecwebapp/templates/partials/candidate/about-candidate.html
+++ b/openfecwebapp/templates/partials/candidate/about-candidate.html
@@ -107,7 +107,7 @@
       {% for committee in committee_groups['P'] | reverse %}
         <div class="callout callout--primary{% if loop.last %} u-no-margin{% endif %}">
           <h4 class="callout__title">
-            <a href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle) }}">{{ committee.name }}</a>
+            <a href="{{ url_for('committee_page', c_id=committee.committee_id) }}">{{ committee.name }}</a>
           </h4>
           <span class="callout__subtitle term" data-term="principal campaign committee">Principal campaign committee</span>
         </div>
@@ -117,7 +117,7 @@
       <ul class="list--bulleted row">
       {% for committee in committee_groups['A'] | reverse %}
         <li>
-          <a class="t-sans" href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle) }}">{{ committee.name }}</a>
+          <a class="t-sans" href="{{ url_for('committee_page', c_id=committee.committee_id) }}">{{ committee.name }}</a>
         </li>
       {% endfor %}
       </ul>
@@ -131,7 +131,7 @@
       <ul class="list--bulleted">
       {% for committee in committee_groups['J'] | reverse %}
         <li>
-          <a class="t-sans" href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle) }}">{{ committee.name }}</a>
+          <a class="t-sans" href="{{ url_for('committee_page', c_id=committee.committee_id) }}">{{ committee.name }}</a>
         </li>
       {% endfor %}
       </ul>
--- a/openfecwebapp/templates/partials/candidate/about-candidate.html
+++ b/openfecwebapp/templates/partials/candidate/about-candidate.html
@@ -1,5 +1,3 @@
-{% import 'macros/missing.html' as missing %}
-{% import 'macros/null.html' as null %}
 {% import 'macros/cycle-select.html' as select %}
 
 <section id="section-2" role="tabpanel" aria-hidden="true" aria-labelledby="section-2-heading">
--- a/openfecwebapp/templates/partials/candidate/individual-contributions.html
+++ b/openfecwebapp/templates/partials/candidate/individual-contributions.html
@@ -21,14 +21,39 @@
   </ul>
   <div class="content__section">
     <div class="results-info results-info--simple">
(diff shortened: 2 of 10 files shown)


openfecwebapp/templates/partials/candidate/individual-contributions.html
```diff
@@ -0,0 +1,152 @@
...
+      <div id="contributor-state" class="panel-toggle-element">
+
+        <div class="results-info results-info--simple">
+          <ul class="u-float-left">
+            <li class="tag__category">
+              <div class="tag__item">Report year: {{cycle - 1}} to {{cycle}}</div>
+            </li>
+          </ul>
+
+          <button type="button" class="js-export button button--cta button--export" data-export-for="contributor-state">Export</button>
```
noahmanger: We need to somewhere add the logic to disable this if the total count is over the limit (100k results).

static/js/pages/candidate-single.js
```diff
@@ -4,11 +4,41 @@
...
 var columns = require('../modules/columns');
 
+function buildStateUrl($elm) {
+  return helpers.buildUrl(
+    ['committee', $elm.data('committee-id'), 'schedules', 'schedule_a', 'by_state'],
+    {cycle: $elm.data('cycle'), per_page: 99}
+  );
+}
+
+function highlightRowAndState($map, $table, state, scroll) {
```
noahmanger: Is this the same function that's used on the committee state map? Should it be exported from some common place so it's not duplicated?

openfecwebapp/templates/partials/candidate/individual-contributions.html
xtine: Is there another place where the button is disabled with large total counts?

xtine:
Pulled the map event code into its own modules file, and simplified the state map code in candidate/committee files.

xtine:
@noahmanger: If you go to the current candidate page for Hillary (https://beta.fec.gov/data/candidate/P00003392/?cycle=2016&election_full=true), the link out for Hillary Clinton for President results in a 404 because of the cycle. Is this related to `committee_groups['A']`?

noahmanger:
Ah yes. @xtine this came up earlier in https://github.com/18F/openFEC-web-app/issues/1852 . The issue is that the committee was active in a previous two-year period and was then shut down, so there's no page for `?cycle=2016`.

The solution we nailed down there is to exclude the cycle paramater when linking to committee pages from the most recent time period. 

That will solve the broken link problem, but won't solve the issue of including old committees in the "Showing data for:" section. I'm not sure how to deal with that (other than removing entirely).

openfecwebapp/templates/partials/candidate/individual-contributions.html
noahmanger: It's how the main datatable pages work (checkout the `datatable.refreshExport()` method).

static/js/modules/maps-event.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
+
+var $ = require('jquery');
```
noahmanger: Yay for pulling this out! Would it make more sense included in `maps.js` though? Fewer places to look in the future.

openfecwebapp/templates/partials/candidate/about-candidate.html
```diff
@@ -107,7 +107,7 @@ <h2 id="section-2-heading">About this candidate</h2>
       {% for committee in committee_groups['P'] | reverse %}
         <div class="callout callout--primary{% if loop.last %} u-no-margin{% endif %}">
           <h4 class="callout__title">
-            <a href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle) }}">{{ committee.name }}</a>
+            <a href="{{ url_for('committee_page', c_id=committee.committee_id) }}">{{ committee.name }}</a>
```
noahmanger: Ack, sorry, I don't think I explained that well. We only want to exclude the cycle parameter if we're looking at the most recent time period. That way if you're looking at historical data, it still takes you to the historical committee page. Or does that bring up all kinds of weird logic that would just make it a mess?

xtine: Hmm.. in `constants.py`, is `DEFAULT_TIME_PERIOD` the most recent time period? If so then I can just omit passing cycle to url if it matches that in `committee_page`.

xtine:
@noahmanger: Would it be sufficient to just only show committees that match the current cycle in itemized disbursements? Example:
```
      {% for committee in committee_groups['P'] | reverse %}
      {% if committee.cycle == cycle %}
      <li>
        <a class="t-sans" href="{{ url_for('committee_page', c_id=committee.committee_id, cycle=cycle) }}">{{ committee.name }}</a>
      </li>
      {% endif %}
      {% endfor %}
```

noahmanger:
Oh yeah! That does it.

openfecwebapp/templates/partials/candidate/individual-contributions.html
xtine: I was confused at first, but then realized that it wasn't disabling the button because I didn't label the button correctly (individual-contributions).

static/js/modules/maps-event.js
xtine: The state map is also used in the elections page and doesn't have the map and table click events.

Issue #1877, opened 2017-03-08
Implement new "advanced data" page


noahmanger:
We have designs for a new "Advanced data" layout from https://github.com/18F/fec-cms/issues/833 so this task is to implement them. I currently have this in progress and expect to be done with it this sprint, but making an issue for tracking it.

---

[Designs](https://gsa.invisionapp.com/share/HQAOQLDF9#/221534844_Advanced-Data_Raising)

**Completion criteria:**
- [x] Organize page into new layout with side tabs
- [x] Add links to legacy bulk downloads and other data sets
- [x] Add disclaimer language at the bottom
- [x] Add link to API
- [ ] ~Add link to "About the data" page~ This page still needs to be made

Issue #1838, opened 2017-02-15
Add links to new data pages on Advanced Data


noahmanger:
So that users can find the pages for loans, debts and party expenditure, add links to the Advanced Data page.

---

Party expenditures should go under "Spending" but I'm not sure about loans and debts. Maybe we need a new section for these? If so, we will need @emileighoutlaw to add a blurb of content perhaps.

---

## Update
Party expenditures is added, so we just need a section for loans and debts.
- [ ] Write content for section
- [ ] Add to page

emileighoutlaw:
In our content, we've been separating loans, debts, and advances out as its own thing, separate from raising and spending. So that would be nicely consistent with the content, if we went in this direction.

noahmanger:
I'll add these links in this work (https://github.com/18F/openFEC-web-app/issues/1877) but we still need content explaining what loans and debts are. Is there a working definition we've used elsewhere?

emileighoutlaw:
We have content on the `loans, debts and advances` landing page, but I don't think we can reuse it for this (it doesn't quite translate between contexts). 

@AmyKort — do you think this is something you and I could pair on together? Ask someone from the content managers or data team to help us write?

AmyKort:
Yes!

noahmanger:
This work is now included in [this issue](https://github.com/18F/openFEC-web-app/issues/1885) here. We'll deal with all of the content for this page in one fell swoop.

PR #1878, opened 2017-03-09
Feature/advanced data

noahmanger:
PR implementing the new layout for the Advanced Data page.

Resolves https://github.com/18F/openFEC-web-app/issues/1877
Resolves https://github.com/18F/openFEC-web-app/issues/1838

---

## Update
This updates the Advance Data page to use the sidebar tabs like so:

image

I implemented this with all of the same copy as we had before, but in some cases broke things out slightly to fit the new organization. I still think we need to put some more thought into this content, but this will do for now.

It also:
- Adds the disclaimer 
- Adds the link to the API
- Adds links to legacy data tools

### DIFF
--- a/openfecwebapp/templates/advanced.html
+++ b/openfecwebapp/templates/advanced.html
@@ -100,15 +100,14 @@
     <div class="main__content--right-full">
       <section class="row" id="raising" aria-hidden="false" role="tabpanel" >
         <h2>Raising</h2>
-        <p>Receipts are anything of value (money, goods, services or property) received by a political committee. Authorized committees take in all receipts for a candidate’s campaign. Receipts include both contributions and other forms of support. This data is updated according to a committee's filing schedule and processing time may vary.</p>
         <div class="post-feed">
           <div class="post">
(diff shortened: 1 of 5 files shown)


noahmanger:
This is ready to go.

noahmanger:
Also: in the site map we had the PACronyms stuff under the "About the data" page, but what about just moving it directly to this page on the Committees tab, like:

image

openfecwebapp/templates/advanced.html
```diff
@@ -6,155 +6,339 @@
...
-          <li class="side-nav__item"><a class="side-nav__link" href="#spending">Spending</a></li>
-          <li class="side-nav__item"><a class="side-nav__link" href="#filings">Reports and filings</a></li>
-        </ul>
-      </nav>
-    </div>
-    <section class="main__content--right">
-      <div id="candidates" class="option">
+  {{ header.header(title, '') }}
+<div class="tab-interface">
+<div class="main container">
```
xtine: Can you indent this?

xtine:
@jenniferthibault: The code is fine, so it's fine to merge if it looks good to you.

PR #1895, opened 2017-03-23
Fix permissions of the cf_env_setup.sh script

ccostino:
This changeset fixes the permissions of the cf_env_setup.sh script so that it is executable.


ccostino:
Whoever is able to review this, please feel free!

PR #1906, opened 2017-03-28
Feature/gov cloud changes

LindsayYoung:
Adds Bandit 
Get rid of lingering Sentry
New Relic name standardization


manifest_stage.yml
```diff
@@ -7,5 +7,5 @@ env:
   FEC_CMS_URL: https://fec-stage-proxy.app.cloud.gov
   FEC_WEB_API_URL: https://api-stage.open.fec.gov/
   FEC_WEB_DEBUG: true
-  FEC_WEB_ENVIRONMENT: stage
+  FEC_WEB_ENVIRONMENT: prod
```
ccostino: This is the staging manifest, we should change this back to `stage`, right?

Issue #1894, opened 2017-03-22
Make data landing page resilient to API failures


noahmanger:
Currently, if an API endpoint that the data landing page relies on is down, the page throws a key error and crashes. We should update the view logic to be more resilient to prevent these issues from crashing the page in the future.

LindsayYoung:
I *think* we want to change the line in api_caller.py:
https://github.com/18F/openFEC-web-app/blob/develop/openfecwebapp/api_caller.py#L35
```
    return results.json() if results.ok else {}
```
to 
```
    return results.json() if results.ok else {'results': []}
```
I think that will make it a bit more graceful of a fail if an endpoint is misbehaving.

PR #1910, opened 2017-03-28
Preventing key errors if the API is down

noahmanger:
Small fix to prevent key errors that occur if API is down. This checks to make sure there's a `results` key in the `response` and otherwise returns `None`. 

Let me know if there's a better way to do this.

cc @LindsayYoung 

Resolves https://github.com/18F/openFEC-web-app/issues/1894


LindsayYoung:
Thank you so much! We will still want to fix things, of course but this removes a lot of pressure if one of the endpoints misbehave

PR #1911, opened 2017-03-29
Feature/server side datatables

noahmanger:
This is a proof of concept for adding basic server-side rendering of the candidate and committee data table pages, which can help with search engine indexing. 

image

The idea is that for beta.fec.gov/data/candidates and beta.fec.gov/data/candidates , we generate a basic alphabetically ordered list of all candidates and committees server-side. If JS is enabled, the regular data table and filtering experience takes over. If not, the server-side data stays.

The only thing you can do is page through the results and click through to committee pages (so there's no filters), which _should_ be enough for a search spider to do it's the work of crawling through and indexing all the pages. 

A future step could be building this out to use this approach for the other pages and/or adding some simple filters, which would ensure these pages work in some minimal way for non-js users.

cc @xtine @onezerojeremy

### DIFF
--- a/openfecwebapp/constants.py
+++ b/openfecwebapp/constants.py
@@ -254,7 +254,7 @@ table_columns = OrderedDict([
     ('candidates-office-president', ['Name', 'Party', 'Receipts', 'Disbursements']),
     ('candidates-office-senate', ['Name', 'Party', 'State', 'Receipts', 'Disbursements']),
     ('candidates-office-house', ['Name', 'Party', 'State', 'District', 'Receipts', 'Disbursements']),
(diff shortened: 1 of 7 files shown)


xtine:
I see some discrepancy of election years between the JS and no-JS candidate pages:

screen shot 2017-03-30 at 5 26 48 pm
screen shot 2017-03-30 at 5 27 07 pm

Also, the no-JS pages don't have titles, could we should a header?

noahmanger:
Huh. Good eye on the years. I'll look into that. And yeah, will add the title.

noahmanger:
Wow I totally forgot to follow up on this. Even though this isn't necessary for site search, this will still be good to get in.

Made your change, @xtine 
image

PR #1914, opened 2017-03-29
Bugfix/consume citation text

anthonygarvan:


noahmanger:
Does this need to get into the release?

anthonygarvan:
Derp. This is duplicated elsewhere.

PR #1916, opened 2017-03-30
adding build legal to build task

anthonygarvan:


package.json
```diff
@@ -47,10 +47,10 @@
     "vinyl-paths": "^1.0.0"
   },
   "scripts": {
-    "build": "gulp build-js && gulp build-sass",
+    "build": "gulp build-js && npm run build-legal && gulp build-sass",
```
ccostino: This seems like a circular/recursive "reference" to me if you will, is it okay to use `npm run ...` to in turn execute another `npm run ...`?

anthonygarvan: yeah npm run just executes a shell command. That could be a gulp task, another npm task, anything.

anthonygarvan: If `npm build` ran `npm build-legal` which ran `npm build`, _that_ would be trouble.

ccostino: 😆 yes, yes it would!  Thanks for explaining!

PR #1927, opened 2017-04-03
adding url support, fixing pagination bug and marshmallow bug

anthonygarvan:


static/js/legal/LegalSearch.js
```diff
@@ -9,24 +9,36 @@ const Pagination = require('./Pagination');
 class LegalSearch extends React.Component {
   constructor(props) {
     super(props);
-    this.state = { q: $('#query').val(), from_hit: 0, advisory_opinions: []}
+    const initState = URI.parseQuery(window.location.search);
```
noahmanger: Still getting my head around es6, but I thought you weren't supposed to use `const` when you're changing things? Should this be a `let` instead?

static/js/legal/LegalSearch.js
```diff
@@ -9,24 +9,36 @@ const Pagination = require('./Pagination');
...
     this.getResults();
   }
 
-  setQuery(e) {
+  setQuery(e, callback) {
     const value = e.target.type === 'checkbox' ? e.target.checked : e.target.value;
-    this.setState({ [e.target.name]: value, lastFilter: e.target.name});
+    const newState = {[e.target.name]: value, lastFilter: e.target.name};
+
+    if(e.target.name !== 'from_hit') {
```
noahmanger: What is this line doing?

static/js/legal/LegalSearch.js
anthonygarvan: No, it's `const` because the _reference_ to the object isn't changing. It would be `let` if I did `let initState = {obj1: 1}; initState = {obj2: 2}`, but it's `const` for `const initState = {obj1: 1}; initState.obj1 = 2`.

static/js/legal/LegalSearch.js
anthonygarvan: if any of the substantive searches change, I want to return results starting from page 1 (offset 0). But if `from_hit` is triggered, I want to increment / decrement as intended.

This came out of some testing I did where I went to the second page, then changed the filter, than was still at page 2 of the _new_ search, which seemed really strange to me.

PR #1930, opened 2017-04-05
Adding conditionals for cases when candidates don't have totals

noahmanger:
This fixes the conditional logic when rendering candidate pages so the pages don't break if the candidate doesn't have financial totals for the selected cycle.

To verify, make sure this page works: http://localhost:3000/candidate/P80001571/?election_full=true&cycle=2020


xtine:
Thanks for posting a handy testing link. 👍

PR #1931, opened 2017-04-06
Feature/add react tests

anthonygarvan:


xtine:
What are the `.snap` files?

anthonygarvan:
They are automatically generated "snapshots" of the rendered component. The first time you run the tests, it generates those .snap files, after that it will fail the tests if the new component doesn't match the snapshots.

noahmanger:
So for snapshot testing, if something changes and it's supposed to, the test will fail, but then you just update the snapshot?

This all makes sense, but is there any other logic or behavior we should also be testing in addition to the rendering? Not familiar enough with the components.

anthonygarvan:
Almost all of the components are "pure" and do not hold state, but yeah, I should probably figure out a way to add a few tests to the `LegalSearch` component as well (that holds the state). Can we get this merged so that I have something to build off of for future PRs?

noahmanger:
Sounds good to me!

PR #1932, opened 2017-04-06
Advanced data edits

noahmanger:
This mimics the CMS's approach for setting classic and transition URLs as env variables, and makes a few changes to use the correct base URLs for certain links on the advanced data page.

It also makes a small copy edit requested by @PaulClark2 .

Can you make sure this is right @patphongs ?

### DIFF
--- a/openfecwebapp/config.py
+++ b/openfecwebapp/config.py
@@ -12,7 +12,8 @@ api_key = env.get_credential('FEC_WEB_API_KEY', '')
 api_key_public = env.get_credential('FEC_WEB_API_KEY_PUBLIC', '')
 cache = os.getenv('FEC_WEB_CACHE')
 cache_size = int(os.getenv('FEC_WEB_CACHE_SIZE', 1000))
-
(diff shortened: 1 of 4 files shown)


openfecwebapp/config.py
```diff
@@ -12,7 +12,8 @@
 api_key_public = env.get_credential('FEC_WEB_API_KEY_PUBLIC', '')
 cache = os.getenv('FEC_WEB_CACHE')
 cache_size = int(os.getenv('FEC_WEB_CACHE_SIZE', 1000))
-
+transition_url = env.get_credential('FEC_TRANSITION_URL', 'http://www.fec.gov')
```
patphongs: I think the transition URL is stable enough now where you can change this variable to https://transition.fec.gov

noahmanger: I just changed it to default to `transition.fec.gov` but still have the possibility of overwriting with an env var just in case we need to change it suddenly.

patphongs:
@noahmanger Links are good to go!

PR #1938, opened 2017-04-10
[MERGE WITH GIT FLOW] Update New Relic Python Agent

ccostino:
This changeset updates the New Relic Python agent to the latest version available.  We were experiencing some major slow down and some errors intermittently, so hopefully this will address the issue.


anthonygarvan:
looks good.

PR #1942, opened 2017-04-11
Hotfix/use async worker2

vrajmohan:


bin/run.sh
```diff
@@ -1,3 +1,3 @@
 # turning off slack for now!
 # invoke notify
-gunicorn -w 2 openfecwebapp.app:app
+gunicorn --log-level debug -k gevent -w 2 openfecwebapp.app:app
```
ccostino: Just a note for others - we're going to have this set for the time being to watch what's happening and try and glean more info, then shut it off.

requirements.txt
```diff
@@ -10,6 +10,7 @@ python-dateutil==2.4.2
 requests==2.13.0
 newrelic==2.82.0.62
 gunicorn==19.7.1
+gevent==1.1.2
```
ccostino: Should we have 1.2.1 here instead?  We can adjust in a follow-up PR after reverting log level.

PR #1947, opened 2017-04-13
Fixing canonical URLs

noahmanger:
This shows correct canonical URLs for pages by adding a `canonical_base` config variable, which is set to beta.fec.gov now but will eventually be fec.gov, and then a `get_canonical_url()` utility.


xtine:
Looks good, just needs to fix conflict.

noahmanger:
Oh, thanks. Got it.

PR #1950, opened 2017-04-13
Fix/pending aos

noahmanger:
Adds `ao_category` to the API call to show pending AOs on the AO landing page. 

image


vrajmohan:
Looks great!

PR #1957, opened 2017-04-18
Feature/unify search

noahmanger:
This change goes along with the site search in https://github.com/18F/fec-cms/pull/972 . It unifies the candidate and committee searches on the data site so you no longer need to select a category before searching. 

Results are returned using the same pattern as the site search:

search results - fec


openfecwebapp/filters.py
```diff
@@ -82,3 +82,11 @@ def fmt_cycle_min_max(cycles):
...
         return '{}–{}'.format(min(cycles), max(cycles))
     return cycles[0]
+
+@app.template_filter()
+def get_max(list):
+    return max(list)
+
+@app.template_filter()
+def get_min(list):
+    return min(list)
```
noahmanger: I was surprised they didn't already exist!

Issue #1888, opened 2017-03-17
Update "no data" messages


noahmanger:
Once we do the site flip, we will likely want to update these messages that appear when no data has been reported by a candidate or committee for a certain time period:

![image](https://cloud.githubusercontent.com/assets/1696495/24025005/e0035730-0a74-11e7-8d1f-398f45510546.png)

emileighoutlaw:
Bringing this more in line with our other error messages (issue: https://github.com/18F/fec-cms/issues/881), how about this:

> ## No results
> Sorry, we didn’t find any documents filed for **XXXX** in **XXXX**.

> [Try a different search]

> Think this was a mistake? Send us an _email_ or file a _GitHub issue_. 

cc @AmyKort

noahmanger:
My $.02: I'd move away from the "Sorry, we didn't find any" language. This message is used when there's just no reports filed for a committee, so I think it would be good to separate it from any sense of there being something wrong. I think we had this more cautious language up at first when there was more of a risk, but now I think we can assert a little more confidence. Does that make sense?

emileighoutlaw:
Yep — that makes sense. Do I remember correctly that there sometimes be a decent amount of lag-time between a filing deadline and when documents show up in our search?

What I'm wondering is if we should take a step back from "sorry" but still be gentle:

> ## No results
> We didn’t find any documents filed for **XXXX** in **XXXX**.

or if we can/should be more firm about the documents not existing:

> ## No results
> No documents have been filed for **XXXX** in **XXXX**.

emileighoutlaw:
> ## No results
> We don’t have any financial data for **XXXX** in **XXXX**. 

> Sometimes no results are returned when:
> - There hasn't been a filing deadline yet
> - A filing deadline has passed, but the FEC hasn’t finished processing the financial data

> To find the most recent activity for this candidate, try changing the election cycle. 

> [ Search another year ] 

> Think this was a mistake? Send us an _email_ or file a _GitHub issue_.



?

noahmanger:
What about:

> We don’t have any financial data for **XXXX** in **XXXX**. 

> This can either be because:
> - There hasn't been a filing deadline yet, or
> - A filing deadline has passed, but the FEC hasn’t finished processing the financial data

> To find the most recent activity for this candidate [or committee], try changing the election cycle. 

> Think this was a mistake? Send us an _email_ or file a _GitHub issue_.

[...]

PR #1958, opened 2017-04-18
Feature/no data messages

noahmanger:
This updates the "no data" messages from https://github.com/18F/openFEC-web-app/issues/1888

It also makes a few small improvements to candidate pages:
- Adds `entity__figure--narrow` class to set a max-width on entity figures, like:
image

- Rearranges the tabs to match the original designs

Still making final copy edits to the messages.


openfecwebapp/templates/macros/missing.html
```diff
@@ -1,14 +1,15 @@
 {% macro missing_financials(entity, cycle) %}
-<div class="message message--no-icon">
-  <h2 class="message__title" id="section-1-header" tabindex="0">No results</h2>
-  <p>Sorry, we didn't find any financial data for <span class="t-bold">{{ entity }}</span> in the <span class="t-bold">{{ cycle | fmt_year_range }}</span> election cycle.</p>
+<div class="message message--info">
+  <p>We don't have any financial data for <span class="t-bold">{{ entity }}</span> in <span class="t-bold">{{ cycle | fmt_year_range }}</span>.</p>
+  <p>This can either be because:</p>
```
emileighoutlaw: Now that I'm looking at this again, I'm worried that the "either" makes it seem like these are the only two possible reasons, which I don't think is the case. What if we said 

_Sometimes this happens when a candidate [or committee]:_ 

instead?

noahmanger: That doesn't make sense with the next bullet being "There hasn't been a filing deadline yet" so what about just removing the "either"?

openfecwebapp/templates/candidates-single.html
```diff
@@ -56,28 +56,28 @@ <h1 class="entity__name">{{ name }}</h1>
           <a
             class="side-nav__link"
             role="tab"
-            data-name="other-spending"
-            tabindex="2"
+            data-name="individual-contributions"
+            tabindex-"3"
```
xtine: `-` instead of `=`

noahmanger: Oh thanks

PR #1959, opened 2017-04-18
Feature/ao entities

noahmanger:
Now that we have the entity filters as implemented in #1955 , I realized we could also add the entity table to the AO page to show requestor, counsel and commenters. 

While I was in there, I noticed that since we refactored the AO schema a while ago, the big final opinion button was also missing, so I fixed that.

And last, I went ahead and added the side nav layout from [the design](https://gsa.invisionapp.com/share/QCA404INT#/229485601_AO_Canonical-Collapsed-Demo).

So here's nicer AO pages:
ao 2013-06 - fec

cc @jenniferthibault @nickykrause


noahmanger:
Good eye. Got it:

image

PR #1964, opened 2017-04-19
Fetch AO results on enter

noahmanger:
This adds a keydown handler to search AOs when hitting enter in text filters to address an issue we've  noticed in [usability testing](https://github.com/18F/fec-testing/issues/43).


anthonygarvan:
This is a great start, but there's no need to lift state in this case. The keypress function can just live in the TextFilter component. I used this as a starting point and resolved in #1968 .

noahmanger:
Ah nice! Thanks. I tried to do it there first but I see what I did wrong now.

PR #1966, opened 2017-04-19
Feature/misc fixes

noahmanger:
This makes several changes that just came up in the meeting with @ChristianHilland, @llienfec, Jason, @PaulClark2 and @jwchumley 👍 

Sets the default time period to 2018, which means that receipts and disbursements pages now default to 2017-2018 and the charts on the data landing page default to 2017-2018:

image

Removes the financial summary link from `Total receipts` and adds it to `Itemized receipts`:
image

Removes the "Disbursements received from other committees". This left only one table in the "Between committees" tab, so I thought it instead made sense to group it with the other disbursements, like:
image

Adds a sentence to the "All candidates" item on Advanced data explaining what the difference is:
image

Fixes the broken links to raising and spending overviews

Gives extra space for the map on the election pages (requires a style PR):
image

### DIFF
--- a/openfecwebapp/constants.py
+++ b/openfecwebapp/constants.py
@@ -2,7 +2,7 @@ from collections import OrderedDict
 
 START_YEAR = 1979
 END_YEAR = 2018
-DEFAULT_TIME_PERIOD = 2016
+DEFAULT_TIME_PERIOD = 2018
 
 states = OrderedDict([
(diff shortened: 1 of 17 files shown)


noahmanger:
Here's a new pass at labels for the disbursements tab:
image

What do you think @llienfec and @PaulClark2 ?

xtine:
👍

PR #1972, opened 2017-04-21
Feature/fix election lookup dates

noahmanger:
This adds a call to the `/election-dates/` endpoint to check for upcoming elections for districts, such as special elections. If it finds one, it uses that date. Otherwise, it uses the date of the next general election. 

image

It also makes a few minor UI tweaks for greater clarity. Because the election may be in either of the years of a two-year period, I changed single years to be year ranges, a la:
image

Also, as someone noted the other day, the "current officeholder" doesn't make sense for historic elections, and also seems to be inaccurate for cases like special elections, so I just removed it.

### DIFF
--- a/static/js/modules/election-lookup.js
+++ b/static/js/modules/election-lookup.js
@@ -371,7 +371,8 @@ ElectionLookup.prototype.drawLocations = function($svg) {
 
 ElectionLookup.prototype.getTitle = function() {
   var params = this.serialized;
-  var title = params.cycle + ' candidates';
+  var minYear = Number(params.cycle) - 1;
+  var title = minYear + '–' + params.cycle + ' candidates';
   if (params.zip) {
     title += ' in ZIP code ' + params.zip;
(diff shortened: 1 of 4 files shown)


static/js/modules/election-lookup.js
```diff
@@ -193,6 +194,24 @@ ElectionLookup.prototype.init = function() {
   });
 };
 
+ElectionLookup.prototype.getUpcomingElections = function() {
+  // Call the API to get a list of upcoming election dates
+  var now = new Date();
```
xtine: I know at least in the CMS there's a few places we check for dates, do you think it would be helpful if we have a helper function that just returns current date and other common date functionality?

noahmanger: That's a good idea. I've found myself writing this pattern a few times.

LindsayYoung:
This is great! thanks @noahmanger

PR #1981, opened 2017-04-25
Feature/accessibility updates

xtine:
Update aria-controls label for mobile nav trigger, and fix aria-labelledby typos.


noahmanger:
Heh. Funny how things work better when you spell them right 🙈

PR #1982, opened 2017-04-25
adding templating for reg and statute citations

anthonygarvan:
cc @jenniferthibault I tightened up the spacing a little by changing `<p>`'s to `<div>`'s. If you or @nickykrause take a look to review, you can just recommend existing classes to apply (preferred) or if we really need to we can make a new class (or maybe adding a new utility would be better?).


noahmanger:
Beat me too it! Looks good to me too :)

PR #1986, opened 2017-04-26
jQuery 3.2.1

noahmanger:
This updates jQuery to the latest version. I clicked through the site and everything seems to be working. I also ready through the [upgrade guide](https://jquery.com/upgrade-guide/3.0/#overview) and checked for instances of the things they warned about and I didn't find anything concerning.

~The one snag is that a single test is failing, which seems to perhaps relate to the change in the `$.Deferred()` implementation.~

~But I'm stumped, so I'd appreciate if someone else could take a look.~

Update: the test issue is resolved.

### DIFF
--- a/static/js/modules/download.js
+++ b/static/js/modules/download.js
@@ -137,7 +137,7 @@ DownloadItem.prototype.refresh = function() {
     contentType: 'application/json',
     dataType: 'json'
   });
-  this.promise.then(this.handleSuccess.bind(this));
+  this.promise.done(this.handleSuccess.bind(this));
   this.promise.fail(this.handleError.bind(this));
 };
 
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "handlebars": "^4.0.5",
     "hbsfy": "2.2.1",
     "intl": "1.0.0-rc-4",
-    "jquery": "2.2.4",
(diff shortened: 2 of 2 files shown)


static/js/modules/download.js
```diff
@@ -137,7 +137,7 @@ DownloadItem.prototype.refresh = function() {
     contentType: 'application/json',
     dataType: 'json'
   });
-  this.promise.then(this.handleSuccess.bind(this));
+  this.promise.done(this.handleSuccess.bind(this));
```
ccostino: Interesting, this required a change to `done`?  In reading the guide I would've thought `then` would still be preferred as that is the new way forward with the `Deferred` object changes.  It appears we're still using the old style, though, [according to this part](https://jquery.com/upgrade-guide/3.0/#callback-exit), which would make sense given this change.

noahmanger: Yep, I was confused about this too. But there was an issue with the wrong context not being passed and this solution fixed it. Since it's a test and not a feature, seemed ok to just leave it.

ccostino: Sounds good to me!  Should we work in the `fec-style` version bump in this PR to just ensure full testing of the updates?

noahmanger: Sounds god.

PR #1993, opened 2017-04-27
Adding tabindex to nav links

noahmanger:
This fixes the mega menu in Safari by adding tabindex to the links. 

The issue is that the mega menu calls the click handler when the triggering element has focus. Chrome automatically adds focus to an element when you click it, but Safari doesn't unless there's a tabindex. 

Also requires https://github.com/18F/fec-style/pull/688

### DIFF
--- a/openfecwebapp/templates/partials/navigation.html
+++ b/openfecwebapp/templates/partials/navigation.html
@@ -8,17 +8,17 @@
         </a>
       </li>
       <li class="site-nav__item" data-submenu="data">
-        <a href="{{ url_for('search') }}" class="site-nav__link {% if parent == 'data' %}is-parent{% endif %}">
+        <a href="{{ url_for('search') }}" tabindex="0" class="site-nav__link {% if parent == 'data' %}is-parent{% endif %}">
           <span class="site-nav__link__title">Campaign finance data</span>
         </a>
       </li>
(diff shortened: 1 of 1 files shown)


openfecwebapp/templates/partials/navigation.html
```diff
@@ -8,17 +8,17 @@
...
       </li>
       <li class="site-nav__item" data-submenu="data">
-        <a href="{{ url_for('search') }}" class="site-nav__link {% if parent == 'data' %}is-parent{% endif %}">
+        <a href="{{ url_for('search') }}" tabindex="0" class="site-nav__link {% if parent == 'data' %}is-parent{% endif %}">
           <span class="site-nav__link__title">Campaign finance data</span>
         </a>
       </li>
       <li class="site-nav__item site-nav__item--secondary" data-submenu="help">
-        <a href="{{ cms_url }}/help-candidates-committees/" class="site-nav__link">
+        <a href="{{ cms_url }}/help-candidates-committees/" tabindex="0" class="site-nav__link">
```
ccostino: For accessibility reasons, should the `tabindex` be different for each one?

noahmanger: AFAIK it doesn't matter. What matters is that there is one so that the browser knows its focusable, and then they're treated in the order they appear in the DOM.

PR #2000, opened 2017-05-01
Pass office to template to restore correct filters

noahmanger:
As reported in https://github.com/18F/FEC/issues/4032, the [House Candidates](https://beta.fec.gov/data/candidates/house/) page is missing the state (and district) filter. I looked into it and the template code was already there, but it was using the wrong variable to check the office. So this is just a small fix to restore the original designed functionality:

image

Resolves https://github.com/18F/FEC/issues/4032

This also includes a small fix to restore the details panels when looking at raw receipts data.


LindsayYoung:
thanks

Issue #1985, opened 2017-04-26
Full-cycle problem when tabbing around candidate page


noahmanger:
I noticed a small quirk on candidate pages that will pop up for candidates for president and senate who've filed to run for a future election (like https://beta.fec.gov/data/candidate/P80001571/).

When you first go to the page, the cycle is set to the date of the election (so, 2020). When you tab to "Individual contributions" or "Itemized disbursements", the cycle is still 2020, but it should be 2018. As a result, no data shows.

noahmanger:
Related, when on a special election candidate page, the select on itemized tabs is empty:
![image](https://cloud.githubusercontent.com/assets/1696495/25456518/f37fcb44-2a87-11e7-9e84-b7d09baa0c09.png)

noahmanger:
Resolved by https://github.com/18F/openFEC/pull/2372

noahmanger:
Oops, wrong issue.

PR #2013, opened 2017-05-06
Feature/fix candidate cycle logic

noahmanger:
This fixes a few interrelated snags on the candidate pages having to do with the logic around cycles / election years / full 4 and 6 year time periods.

**The problem:** So you're going to a presidential candidate who ran in 2016. The current behavior when going to a candidate page is to load it with `cycle=2016` and `election_full=true`, and then the page will show aggregated data for the 2014 and 2016 cycles. This works fine. 

But the problem is with future dates. Senate and presidential candidates often file to run for office for a year that's 4 or 6 years in the future. The current behavior in this cases is to load the candidate page with `cycle=2020` and `election_full=true`. This causes three problems:

1. Behind the scenes, we use the 2020 value for all API queries for totals and itemized data; e.g. `/schedule_a/?two_year_transaction_period=2020`. While there might be data for the 2018 cycle, which we're in, there will never be data for the 2020 cycle until we're in the year 2019. This results in empty contribution and disbursement tables when looking at the default time period for a candidate currently running for president or senate (https://github.com/18F/openFEC-web-app/issues/1985).

2. When calling the `/totals/` endpoint with `full_election=true`, it currently returns no results because the time period requested includes a time period that hasn't happened yet. That's why when you go to [Schumer's profile](https://beta.fec.gov/data/candidate/S8NY00082/?cycle=2022&election_full=True) there are no results, but if you toggle to 2017-2018, the data shows up.

3. As @jontours uncovered in https://github.com/18F/openFEC/issues/2364, passing a future cycle to the `schedule_e` aggregates also results in null data.

**The solution:** Rather than use the regular `cycle` value for all of the logic explained above, I added logic to set a `max_cycle`. If the `cycle` is before or equal to the current cycle we're in, it stays the same. If it's in the future (i.e. 2020 or later), `max_cycle` will be our current cycle. All of the API calls then use `max_cycle` to get data. 

Similarly, if the request was for `election_full=true` but the `cycle` is in the future, it sets `show_full_election` to `false` so that we're not trying to aggregate data that doesn't yet exist.

**However!** In doing this, I uncovered what seems to be an API bug. #2000 fixed the `/totals/` call by using the correct `full_election` param instead of `election_full`. However, there seems to be a bug in the case of a candidate who only has financial info for _part_ of a full cycle ([like this](https://fec-dev-proxy.app.cloud.gov/data/candidate/P80001571/?cycle=2016&election_full=true)), the API returns nothing. This isn't visible on production because the API call is passing the wrong param, so it's just being ignored, but you can see on dev.

Resolves https://github.com/18F/openFEC-web-app/issues/1985
Resolves https://github.com/18F/openFEC/issues/2364

### DIFF
--- a/openfecwebapp/templates/partials/candidate/other-spending-tab.html
+++ b/openfecwebapp/templates/partials/candidate/other-spending-tab.html
@@ -1,83 +1,84 @@
 {% import 'macros/cycle-select.html' as select %}
 
 <section id="section-3" role="tabpanel" aria-hidden="true" aria-labelledby="section-3-heading">
+  <div class="heading--section">
+    <h2 id="section-3-heading">Spending by others to oppose/support</h2>
+  </div>
   <div class="content__section">
-    <div class="heading--section">
-      <h2 id="section-3-heading">Spending by others to oppose/support</h2>
-    </div>
+    <p class="usa-width-two-thirds">This tab shows spending that opposes or supports this candidate. None of the funds are directly given to or spent by the candidate.</p>
+  </div>
 
+  <div class="slab slab--inline slab--neutral u-padding--left u-padding--right">
     <div class="content__section">
       {{ select.cycle_select(election_years, cycle, duration=duration, id="cycle-3") }}
     </div>
 
-    <p class="usa-width-two-thirds">This tab shows spending that opposes or supports this candidate. None of the funds are directly given to or spent by the candidate.</p>
-  </div>
+    <div class="entity__figure row">
+      <div class="results-info results-info--simple">
+        <h3 class="results-info__title">Independent expenditures</h3>
+      </div>
+      <table
+          class="data-table data-table--heading-borders"
+          data-type="independent-expenditures"
+          data-candidate="{{ candidate_id }}"
+          data-cycle="{{ max_cycle }}"
+          data-election-full="{{ show_full_election }}"
+        >
+        <thead>
+          <tr>
+            <th scope="col">Aggregate amount</th>
+            <th scope="col">Spent by</th>
+            <th scope="col">Support/Oppose</th>
+          </tr>
+        </thead>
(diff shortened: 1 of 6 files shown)


openfecwebapp/views.py
```diff
@@ -185,35 +185,61 @@ def groupby(values, keygetter):
...
             max(cycle for cycle in tmpl_vars['aggregate_cycles'] if cycle in committee['cycles'])
             if election_full
             else candidate['two_year_period']
         )
 
+    # Group the committees by designation
     committee_groups = groupby(committees, lambda each: each['designation'])
     committees_authorized = committee_groups.get('P', []) + committee_groups.get('A', [])
 
+    tmpl_vars['committee_groups'] = committee_groups
```
noahmanger: I did some general moving code around and adding comments, so some of these are false diffs.

openfecwebapp/templates/partials/candidate/other-spending-tab.html
```diff
@@ -1,83 +1,84 @@
...
+    <h2 id="section-3-heading">Spending by others to oppose/support</h2>
+  </div>
   <div class="content__section">
-    <div class="heading--section">
-      <h2 id="section-3-heading">Spending by others to oppose/support</h2>
-    </div>
+    <p class="usa-width-two-thirds">This tab shows spending that opposes or supports this candidate. None of the funds are directly given to or spent by the candidate.</p>
+  </div>
 
+  <div class="slab slab--inline slab--neutral u-padding--left u-padding--right">
```
noahmanger: The markup changes here are just to add styles to match the other tabs (with the white boxes on gray backgrounds).

jontours:
Ah nice fix!  Makes sense to me.  I can look into the bug with the `/totals` endpoint.

PR #2021, opened 2017-05-09
adding loading image during results fetch

anthonygarvan:
Thanks for the headstart @noahmanger , this was quick.


noahmanger:
Nice work!

PR #2040, opened 2017-05-13
HOTFIX: minify css and js

xtine:
This forces minification of scss files (updating library) and uglification of js assets.


gulpfile.js
```diff
@@ -86,7 +86,7 @@ function write(streams, name) {
   return concat(function(body) {
     return file(dest, body, {src: true})
       .pipe(rev())
-      .pipe(gulpif(production, uglify()))
+      .pipe(uglify())
```
noahmanger: I don't think we should remove the environment check here. Uglifying this locally makes it impossible to debug.

xtine: Hm, good point. I just checked and I think it's a case of the production variable not working correctly if prod isn't being uglified.

xtine: I believe FEC needs to be double checked because I think it's not being uglified on local for that as well.

noahmanger: Yeah seems like something is wrong. When I set that variable locally, it gets uglified. Can you remove it from this and we can deal with that issue separately?

noahmanger:
Since we're doing a release this week, I think we should just merge into develop.

Issue #2044, opened 2017-05-16
change name 'unique only' to 'unique individuals only'


LindsayYoung:
Change name 'unique only' to 'unique individuals only'

https://beta.fec.gov/data/receipts

<img width="250" alt="screen shot 2017-05-16 at 4 42 56 pm" src="https://cloud.githubusercontent.com/assets/4406333/26127268/c19de4c4-3a56-11e7-95f3-94f528a060cd.png">

Also since we have a view for individuals, we probably should not set it true by default. that is  how the link starts off here 

https://beta.fec.gov/data/advanced/

Here is the explanation from the API documentation if you want to reference https://api.open.fec.gov/developers/#!/receipts/get_schedules_schedule_a

PaulClark2:
I agree. For the `All receipts` page `unique individuals only` should _not_ be selected by default.

jwchumley:
Agreed.

PR #2051, opened 2017-05-17
Feature/unique fix

noahmanger:
This relabels the "unique receipts" filter to be more descriptive and removes it from being set in links to receipts from the advanced data page.

image

I also moved it to be under the "individual" vs "committee" filter, which feels like a more natural place for it. We had originally pulled it out because it was set by default and we wanted people to see it, but now that it's not checked by default I think it makes sense here. 

cc @LindsayYoung @PaulClark2 

Resolves https://github.com/18F/openFEC-web-app/issues/2044


LindsayYoung:
Yes, thank you!

PR #2056, opened 2017-05-19
fixed spelling error in 3rd paragraph

johnnyporkchops:
Changed "conisderation" to "consideration" per request from @AmyKort


ccostino:
I manually merged this into `develop` so we should be fine - it was deployed when the other hotfix went up (#2057).

PR #2073, opened 2017-05-24
Submit the data search form to the current page

noahmanger:
This resolves the issue where entering another search term from the data search results page sends users to the home page, as reported in https://github.com/18F/FEC/issues/4084

PR #2075, opened 2017-05-24
Hotfix/restore coverage dates

noahmanger:
Once the API fix for coverage dates is in (https://github.com/18F/openFEC/pull/2418) this can be merged to restore them to the frontend. I also rolled in the 1 character change from https://github.com/18F/openFEC-web-app/pull/2073.

### DIFF
--- a/openfecwebapp/app.py
+++ b/openfecwebapp/app.py
@@ -40,7 +40,7 @@ def get_absolute_url():
 
 def get_canonical_url():
     path = furl.furl(request.url).path
-    return '{}/data{}'.format(config.canonical_base, path)
+    return '{}{}'.format(config.canonical_base, path)
 
 
 def _get_default_cycles():
--- a/openfecwebapp/templates/search-results.html
+++ b/openfecwebapp/templates/search-results.html
@@ -13,7 +13,7 @@
     <header>
       <h1 class="heading--main">Candidate and committee profiles</h1>
     </header>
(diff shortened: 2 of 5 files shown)


openfecwebapp/app.py
```diff
@@ -40,7 +40,7 @@ def get_absolute_url():
 
 def get_canonical_url():
     path = furl.furl(request.url).path
-    return '{}/data{}'.format(config.canonical_base, path)
+    return '{}{}'.format(config.canonical_base, path)
```
ccostino: Just confirming we don't need the `/data` here - I know we changed some things around this but wanted to be sure!

openfecwebapp/templates/search-results.html
```diff
@@ -13,7 +13,7 @@
     <header>
       <h1 class="heading--main">Candidate and committee profiles</h1>
     </header>
-    <form action="/" method="GET" class="slab slab--inline slab--neutral">
+    <form action="" method="GET" class="slab slab--inline slab--neutral">
```
ccostino: Also confirming that we don't need the `/` here either.

openfecwebapp/app.py
noahmanger: Yep. The `/data` is actually included in the `path` already.

openfecwebapp/templates/search-results.html
noahmanger: Nope. That was causing the form to submit to the home page and not the main data landing page.

Issue #2074, opened 2017-05-24
Make it easier to access bulk downloads from data pages


noahmanger:
As reported in https://github.com/18F/FEC/issues/4076, the link from a data page to the bulk downloads is hard to access in the tooltip. We should make it easier.

Probably worth starting by looking at the interaction design. There's probably a better way to send people to this page than through a link in the tooltip.

cc @nickykrause

nickykrause:
cc @noahmanger, also @jenniferthibault if you are watching this issue at all:

Some initial thoughts on how to address this issue, at least in the short term. I prefer option 1. 

![bulk data](https://cloud.githubusercontent.com/assets/1450173/26467610/028b707e-4159-11e7-8220-578c29d981f8.png)

jenniferthibault:
@nickykrause 1 or 4 feels like a better match to me because they don't replace the button. What do you think about positioning the alert on the left-hand side of the button instead so that it appears/disappears without needing to bump the button over as it does, which feels like it would be disruptive.

Overall the info alert in 4 feels like a correct (but new!) use of the pattern, so I'd guess we'd need to adjust a new small-format class for it. I think I lean most to 4 with the swapped position of the alert, then button.

nickykrause:
@jenniferthibault . option 4 was my second favorite, so I'm cool with that. Also, 💯 to your comment about moving the button. I agree! Here's option 5, then, with your comments incorporated:

---

![5](https://cloud.githubusercontent.com/assets/1450173/26468121/ea5bde6a-415a-11e7-8217-bc06b1e6f954.png)

Not sure about the microcopy here. Maybe it would be better not to start with the dependent clause, and just to say: `Use the bulk data exporter to export more than 100,000 records.` I didn't go for that originally because it puts the link in the top line, which puts the underlining in the middle of the two lines, like so:

![group](https://cloud.githubusercontent.com/assets/1450173/26468241/4d08d13a-415b-11e7-8c75-af6f776010b0.png)

But perhaps that is fine.

jenniferthibault:
@jameshupp could you lend us your expertise on interface copywriting here? Is it better to put the dependent clause before or after the instruction?

noahmanger:
This makes a lot of sense to me. Nice work!

noahmanger:
Resolved by https://github.com/18F/openFEC-web-app/pull/2083

PR #2083, opened 2017-05-25
Feature/bulk data message

noahmanger:
This addresses a couple interrelated issues that have surfaced over time:
- Users have expressed difficulty figuring out how to control the dates and enter contributor names
- The link in to the bulk data from the disabled export button tooltip is actually unclickable and not obvious

This implements the message pattern as designed in https://github.com/18F/openFEC-web-app/issues/2074 and also pulls out the contributor name and date filters to the top, open panel:
image

It makes a similar change to disbursements:
image

The pattern we're going for is to always expose filters for the two parties of any given transaction in the top, open panel (we already do this with independent expenditures) as well as any constraining factors, like time period. 

Other details:
- The message links to https://www.fec.gov/data/advanced/?tab=other for bulk downloads
- The "Committee name or ID" field was relabelled to "Recipient name or ID" on receipts and individual contributions
- The "Committee name or ID" field was relabelled to "Spender name or ID" on disbursements (same as it is on independent expenditures
- The "Time period" filter was relabelled to "Transaction time period" to make clearer that it's referring to the time period the transaction was made within

cc @nickykrause


noahmanger:
Thanks @jenniferthibault for pairing with me to figure this out in code, and @nickykrause for the smart design ideas.

jenniferthibault:
One small addition: we also relabelled the accordions from `Transaction information` to `Transaction details` to re-emphasize the model that there are some basic filters at the top, but more _detailed_ options in each category below. 

@nickykrause still very open to your thoughts and reactions—just because it's in a PR doesn't mean it's decided, it was just much, much less tedious to move the components in code than rebuilding them in the design files.

noahmanger:
And this is how it looks on candidate / committee pages:
image

nickykrause:
@jenniferthibault @noahmanger all looks good to me, but I appreciate the tag! 👍

static/js/modules/tables.js
```diff
@@ -30,10 +30,10 @@ var downloadCapFormatted = helpers.formatNumber(DOWNLOAD_CAP);
 var MAX_DOWNLOADS = 5;
 var DOWNLOAD_MESSAGES = {
   recordCap:
-    'Exports are limited to ' +
+    'Use <a href="' + window.BASE_PATH + '/advanced?tab=other">' +
```
ccostino: What is `BASE_PATH` defined as here?

noahmanger: On non-local envs: `/data`

PR #2091, opened 2017-05-31
Sort documents in reverse chronological order

vrajmohan:
Fixes 18F/openFEC#2415


anthonygarvan:
Unfortunately I think this will need to be scrapped (sorry!). This story is to add sorting to the API, not the web app- the advisory opinion search no longer goes through the web app (the code you changed is deprecated and should have been removed- that's on me).

Issue #2093, opened 2017-05-31
Table column widths are wrong in Firefox


noahmanger:
As reported originally by https://github.com/18F/FEC/issues/4153 , there's an issue in Firefox (and sounds like IE) where the table column widths are weird, which leads to strange overflows. 

On [/filings](https://www.fec.gov/data/filings/):
![image](https://cloud.githubusercontent.com/assets/1696495/26648419/486eff78-45f7-11e7-92d9-8a7a702423f2.png)

On a committee page:
![image](https://cloud.githubusercontent.com/assets/1696495/26648439/57eb9498-45f7-11e7-9a88-064cf31082b6.png)

There might be more cases too.

Issue #2041, opened 2017-05-15
sort on filings page


PaulClark2:
The sort icon of the document section of the filings page doesn't do anything when you click on it. 

CFD --> Committees --> All comittees --> C00282921 --> WINSTON & STRAWN LLP POLITICAL ACTION COMMITTEE --> Filings

![image](https://cloud.githubusercontent.com/assets/11817734/26065504/26df3ac2-3962-11e7-8e09-5ea0d65ba115.png)

PR #2095, opened 2017-05-31
Fixing sort problems on committee filing tables

noahmanger:
The committee filings table isn't manually sortable, since it uses a complex sorting query to get filings to display in a certain order. However, due to a quirk in the datatables library, it was rendering it as if the first column was sorting, which made it seem like you should be able to sort it in the other direction. This removes the apparent (but not functional) sort styling from the document description and receipt date columns:

image

Resolves https://github.com/18F/FEC/issues/4155
Resolves https://github.com/18F/openFEC-web-app/issues/2041

This also makes it so that reports tables only show links to PDFs if the `pdf_url` is actually present for a given report.

Resolves https://github.com/18F/FEC/issues/4158

---

## Update
This also fixes an issue we're experiencing with wonky column widths in Firefox as a result of using `table-layout: fixed` with a mix of percentage and px based column widths. Firefox is using the percentage value and actually calculating the width correctly, whereas Chrome expands the column if there's space available. The better way to go, however, is to just use `table-layout: auto` and let the browser handle all the work of sizing the columns. This works well in general, and so I removed the column size classes in almost all instances. 

The places where I kept it was `column--med` for the version columns, but changed it to a min-width value set in px and then `column--small` for all date columns, because those have predictable widths and so can be constrained in order to give more space to other columns.

This requires an update to fec-style.

The end result is no more of this:
image
image

And more of this:
image

image

Resolves https://github.com/18F/FEC/issues/4153
Resolves https://github.com/18F/openFEC-web-app/issues/2093

### DIFF
--- a/static/js/modules/columns.js
+++ b/static/js/modules/columns.js
@@ -307,7 +307,7 @@ var filings = {
   pages: pagesColumn,
   version: versionColumn,
   receipt_date: receiptDateColumn,
-  receipt_date_unordable: {
+  receipt_date_unorderable: {
     data: 'receipt_date',
     className: 'min-tablet hide-panel column--small',
     orderable: false,
--- a/static/js/modules/columns.js
+++ b/static/js/modules/columns.js
@@ -127,23 +127,23 @@ var renderCommitteeColumn = function(data, type, row, meta) {
 };
 
 var candidates = [
-  {data: 'name', className: 'all column--large', render: renderCandidateColumn},
-  {data: 'office_full', className: 'min-tablet hide-panel-tablet column--med'},
+  {data: 'name', className: 'all', render: renderCandidateColumn},
+  {data: 'office_full', className: 'min-tablet hide-panel-tablet'},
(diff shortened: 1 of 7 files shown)


static/js/modules/columns.js
```diff
@@ -315,8 +307,17 @@ var filings = {
...
+  receipt_date_unordable: {
+    data: 'receipt_date',
+    className: 'min-tablet hide-panel column--small',
+    orderable: false,
+    render: function(data, type, row) {
+      var parsed = moment(row.receipt_date, 'YYYY-MM-DDTHH:mm:ss');
+      return parsed.isValid() ? parsed.format('MM/DD/YYYY') : 'Invalid date';
+    }
+  },
+  coverage_start_date: dateColumn({data: 'coverage_start_date', className: 'min-tablet hide-panel column--small', orderable: false}),
```
xtine: Spelling -> `receipt_date_unorderable`?

PR #2090, opened 2017-05-30
Feature/remove export

noahmanger:
This removes the export button from the candidate contributions by state and contributions by size tables and adds support for handling 500 server errors.

Now if a user gets an error with text status of `error` you will see this message:

image

Resolves https://github.com/18F/openFEC-web-app/issues/2038
Resolves https://github.com/18F/openFEC-web-app/issues/2089

cc @LindsayYoung

PR #2096, opened 2017-05-31
Hotfix/remove export

noahmanger:
Replaces https://github.com/18F/openFEC-web-app/pull/2090


static/js/modules/download.js
```diff
@@ -180,6 +182,25 @@ DownloadItem.prototype.close = function() {
   this.container.subtract();
 };
 
+DownloadItem.prototype.handleServerError = function() {
```
ccostino: Nice, thanks for adding this!

PR #2108, opened 2017-06-06
Fixed more info links on advanced data committees tab

johnnyporkchops:
Links were pointing to old reg/rptng page, so they point to help4cc page anchors

### DIFF
--- a/openfecwebapp/templates/advanced.html
+++ b/openfecwebapp/templates/advanced.html
@@ -187,7 +187,7 @@
               <div class="icon-heading grid__item">
                 <i class="icon-heading__image icon-circle--checklist"><span class="u-visually-hidden">Icon of a checklist</span></i>
                 <div class="icon-heading__content">
-                  <a href="{{ cms_url }}/registration-and-reporting/essentials-house-and-senate-candidates-and-committees/">Help for candidates and their committees</a>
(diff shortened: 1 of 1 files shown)


noahmanger:
Does this need to be a hotfix, or can we just merge into develop?

Also, please include `hotfix` in the title of hotfix PRs so they don't accidentally get merged in with Github.

johnnyporkchops:
It does not have to be a hot fix if you do not think its urgent

noahmanger:
I can go either way. It's simple enough. @LindsayYoung ?

johnnyporkchops:
My understanding of hofixes is that they are for things that are either urgent or so simple that they don't require a alot of development or review, but  could have that wrong. If there is a release coming up then I am sure its fine for whatever works for better for the reviewer/merger. This is just some dead links that Wie pointed out in slack this week.

noahmanger:
I caught one more of these that needed to be replaced (on the candidates tab) and also I could only get the anchor tags to work if there was no `/` before the URL fragment, so I removed those. 

Can someone double-check though?

noahmanger:
@patphongs can you take a look at this and verify the links are right? I had to push a couple commits.

noahmanger:
Changed this to target develop and we'll just include it in the release.

patphongs:
Thanks! I didn't get to this today but will review tomorrow.

PR #2110, opened 2017-06-07
Removing disbursement purpose filter

noahmanger:
This removes the disbursement purpose filter per the discussion in Slack. The logic behind this filter is still too complicated and inconsistent, and this functionality will soon be partially replaced by the line number filter.

image

cc @jwchumley


LindsayYoung:
Thanks!

PR #2112, opened 2017-06-08
add legal search tags

anthonygarvan:


noahmanger:
@anthonygarvan is this ready for review? Is there anything left that we should plan on doing with it while you're out?

noahmanger:
One thing I noticed testing this is that the requestor type tag should show the option text, not the value:
image

Also, occassionally I'm finding clearing a tag keeps the empty `tag__category` there:
image

Otherwise seems to be working great.

anthonygarvan:
@noahmanger ready. Couldn't reproduce the empty tag bug, any ideas / tips?

static/js/legal/Tags.js
```diff
@@ -0,0 +1,112 @@
...
+  const namedFields = _.extend([] , requestorOptions);
+  _.extend(namedFields, documentTypes);
+  namedFields.push({value: 'ao_is_pending', text: "Pending"});
+  var citationRequireAll = props.query.ao_citation_require_all && props.query.ao_citation_require_all !== 'false';
+  function removeQuery(name, value) {
+    function handleRemove(e) {
+      var newEvent;
+      if(Array.isArray(props.query[name])) {
+        let currentValues = _.extend(props.query[name]);
+        currentValues.splice(currentValues.indexOf(value), 1)
```
noahmanger: This one should be fixed.

static/js/legal/Tags.js
```diff
@@ -0,0 +1,117 @@
...
+        return <li key={tagName} className="tag__category">{getTag(query[tagName].maxName, 'Ending ' + query[tagName].max)}</li>
+      }
+    } else {
+      return <li key={tagName} className="tag__category">{getTag(tagName, query[tagName])}</li>;
+    }
+  }
+
+  function getTagCategories() {
+    let query = _.extend({}, props.query);
+    delete query['from_hit'];
```
noahmanger: What about storing a `const` somewhere with all these things that we don't want to display? That way we could reference it above in the `getUserSearchCriteria` and would only need to update in one place?

static/js/legal/Tags.js
```diff
@@ -0,0 +1,117 @@
...
+    }
+    if(query.ao_min_request_date || query.ao_max_request_date) {
+      query.request_date = {min: query.ao_min_request_date,
+                            minName: 'ao_min_request_date',
+                            max: query.ao_max_request_date,
+                            maxName: 'ao_max_request_date'};
+      delete query.ao_min_request_date
+      delete query.ao_max_request_date
+    }
+    console.log(query);
```
noahmanger: remove console.log

static/js/legal/Tags.js
anthonygarvan: It's only used in one place currently, so I'd rather not overgeneralize. But syntactically a list is much better than a bunch of delete statements, so I'll change that.

anthonygarvan:
@noahmanger issues addressed. The `filtered results` vs `results` text was working properly for me-- keep in mind "Final Opinion" filter is turned on by default, and that the empty `tag__category` bug (which has been fixed) also counted as a "filter", which would have caused the text to be off.

static/js/legal/Tags.js
```diff
@@ -74,11 +74,9 @@ function Tags(props) {
 
   function getTagCategories() {
     let query = _.extend({}, props.query);
-    delete query['from_hit'];
-    delete query['search'];
-    delete query['type'];
-    delete query['search_type'];
-    delete query['ao_citation_require_all']
+    const excludedFields = ['from_hit', 'search', 'type', 'search_type', 'ao_citation_require_all']
```
noahmanger: Nice. 👍

Issue #2039, opened 2017-05-12
load by state profile information only if user requests it


LindsayYoung:
It also looks like we load the individual contribution by state information on the initial page load before people have clicked on the tab. For better performance, we will want to load that client side and only if the user clicks on the tab. 

I was watching the network on this page load on a candidate and committee page

https://beta.fec.gov/data/candidate/H8CA05035/?tab=summary
https://beta.fec.gov/data/committee/C00213512/?cycle=2018&tab=summary

noahmanger:
Ah! Good catch. This is a fix we can make.

noahmanger:
Grabbed this fix in #2113

PR #2106, opened 2017-06-05
Feature/raising spending candidate pages

xtine:
Updates for raising/spending tabs on candidate pages. https://github.com/18F/openFEC-web-app/issues/2087
- Update spending/raising tabs
  - Add sub items
- Show total receipts/disbursements and filter button link 

screen shot 2017-06-05 at 1 47 05 pm
screen shot 2017-06-05 at 1 47 25 pm

Requires: https://github.com/18F/fec-style/pull/727

PR #2113, opened 2017-06-08
Feature/committee page cleanup

noahmanger:
This contains a whole bunch of fixes to round out the work that began in https://github.com/18F/openFEC-web-app/pull/2106 to bring greater consistency to profile pages.

## 1. Page organization consistency
First, for groups that have additional forms of spending other than schedule B disbursements — i.e. groups that make IEs, electioneering comms, or communication costs — those tabs are now also organized under the new "Spending" tab:

image

image

image

## 2. Unified heading style
Second, it unifies the heading style used throughout these pages. Previously the titles and links to filter data varied depending on the table. Now, we use this pattern everywhere, using the title + button + border lockup we use elsewhere on the site:
image
image
image
image
image

## 3. Misc. cleanup and fixes
- Set the cycle select on committee pages to be the same fixed-width as it is on candidate pages
- Only load state map data when the tab is loaded, preventing an extra unnecessary API call on the initial page load (resolves https://github.com/18F/openFEC-web-app/issues/2039 @LindsayYoung )
- Uses the dynamic coverage dates for the total receipts and disbursement figures on candidate pages, rather than hard-coded values
- Adds "transaction date" tags to data tables on candidate and committee pages that show the first and last date of the cycle:
image
- Removes an unnecessary `tabs.html` macro that we were using sometimes and not others, so it tidied the codebase up a little to just get rid of it.

Much thanks to @jenniferthibault on pairing through the nitty gritty details of this.


xtine:
This is great! Nice to unify the headers, but also always great to remove code (`tabs.html`). 👍

PR #2118, opened 2017-06-09
Feature/committee efilings

noahmanger:
This is a blunt way of adding raw filings to committee pages:
image

The way it works is:
- On the server-side: if we're looking at the current cycle and the committee isn't a senate committee, call `/efile/filings` and see if the committee has any filings in the last two days
- If so, add the raw filings section to the template (this is why the call needs to be made server-side first)
- On the client side, call `/efile/filings` again to get the filings and build the data table

Pro:
- It's done and works well

Cons:
- It makes two API calls to the same endpoint. This is a little bit of an unfortunate consequence of the way the pages and data tables are architected. Theoretically possible to resolve, but would take more time.
- It doesn't dedupe the raw filings from the processed filings, but with only two days worth, it's unlikely to be a huge issue and just replicates the functionality in the classic viewer.

cc @LindsayYoung @jwchumley

### DIFF
--- a/openfecwebapp/app.py
+++ b/openfecwebapp/app.py
@@ -147,6 +147,7 @@ app.jinja_env.globals.update({
     'series_group_has_data': series_group_has_data,
     'cycle_start': cycle_start,
     'cycle_end': cycle_end,
+    'two_days_ago': utils.two_days_ago(),
     'election_url': get_election_url,
     'constants': constants,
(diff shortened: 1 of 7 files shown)


LindsayYoung:
Looking now :)

LindsayYoung:
This is a great addition thanks for getting to this so quickly

PR #2119, opened 2017-06-10
Feature/candidate other spending totals

noahmanger:
This adds total support and oppose numbers to the other spending tab on candidate profiles:

image

image

image

Because these numbers aren't available over the API directly, this PR does it all with client-side code that runs when you go to this tab for the first time. It calls each endpoint (and if there's more than 100 records, calls it multiple times to get all of them) and then totals them all up.

@jenniferthibault let me know if you want to pair to fine-tune styles.

cc @LindsayYoung @xtine 

Resolves https://github.com/18F/fec-cms/issues/839


jenniferthibault:
@noahmanger thanks so much for including multiple screen shots lined up, super helpful!

I might have commented on the design issue too late, but what did you think of adding the "Support/Oppose" headers to the IE's & CC's ?

If we're not going that route, the only unexpected style I'm seeing is that there's more space between the totals numbers and the border line on IEs than on CCs and ECs. The larger unit of spacing on IE's the the correct amount to use here. 

Mocked up over your screen shots:

screen shot 2017-06-12 at 12 55 47 pm

noahmanger:
Oh! I misunderstood that point. Yeah I think that's a good idea. I was wanting an extra label too.

noahmanger:
That work? 
image

jenniferthibault:
Yes! ❤️ it, thank you!

noahmanger:
I also added a loading indicator:
loading

xtine (approved):

PR #2142, opened 2017-06-19
Hide net contributions and operating expenditures

noahmanger:
This hides the net operating expenditures and net contributions numbers (and their child lines) as the underlying logic needs some work.

The cash summary portion of financial summaries now looks like:
image

Worth nothing that these numbers are only incorrect for certain committees, but when they're wrong they're also wrong in the classic viewer.


LindsayYoung:
Thanks! Checked this out locally and makes sense. I think the local test data on the API could use some more presidential.

Issue #2055, opened 2017-05-19
Add image number to main filings page table


PaulClark2:
Request from FEC staff: `any chance we can add the image number in a new column on the filings page for a committee`

Being about to see the image number on this page https://beta.fec.gov/data/filings/?data_type=processed would be helpful to staff.

noahmanger:
Just going through the backlog. This isn't a hard thing to add:
![image](https://user-images.githubusercontent.com/1696495/27307605-533959e6-54ff-11e7-8072-b41afc177614.png)

But I wonder if this information makes more sense in the panel? Is it useful to see the image number for a lot of filings all at once?

PaulClark2:
Having the image number available without needing to open the panel makes it easier for many staff to do their jobs more quickly. 

I think there is another issue requesting coverage dates (or at least coverage end date). This is also about helping staff access information more quickly.

noahmanger:
That makes sense to me. I separately have a PR adding those coverage dates to filing tables on committee pages: https://github.com/18F/openFEC-web-app/pull/2139

Issue #2054, opened 2017-05-19
Image number search for statements and reports and specific pages with a document


PaulClark2:
Request from FEC staff member: `My suggestion was adding the image number search to the main search on the “Campaign finance data” page.  We don’t need to actually put in wording that states you can search using the image number, but could we enable that functionality since it is so vital to internal staff?`

This image number search would be different from the image number search on the transaction pages. This search would allow users to enter an image number and return a document. The Viewer allows users to search for any image number. The docquery call look like this: http://docquery.fec.gov/cgi-bin/fecimg/?IMAGE_NUMBER

![image](https://cloud.githubusercontent.com/assets/11817734/26246289/40012568-3c66-11e7-8051-54b21b9ca7f5.png)

noahmanger:
I was just playing around with this and it's easy enough to do a regex check for an 11 or 16 digit number and then redirect to the page-by-page PDF view. It's just a question of UX.

If we're going to add it, we should expose it for everyone. I could see a couple ways.

We could call it out as a separate field, similar to the AO number:
![image](https://user-images.githubusercontent.com/1696495/27307851-b670877c-5500-11e7-9fcb-25385ebc2e95.png)

Or add it as secondary help text:
![image](https://user-images.githubusercontent.com/1696495/27307860-c44f9540-5500-11e7-9ce7-09ae0c80cad4.png)

PaulClark2:
I like option #2 with the secondary text better. @jwchumley @AmyKort do either of you have a preference?

AmyKort:
I also like option 2.

jwchumley:
Option #2 for me as well.


________________________________

noahmanger:
Not urgent, but I'd like to know what @jenniferthibault or @nickykrause think.

nickykrause:
@noahmanger I also prefer option 2 (the helptext). Could it be simpler, like this?: 

`Examples: Obama for America; C00431445; Bush, George W.; P00003335; or enter an image number for a filing`

Just wondering if it needs to be on a second line and contain all the detail it contains, or if an abbreviated version would suffice.

noahmanger:
Works for me. Yeah, the second line felt a little clunky.

jenniferthibault:
I don't have anything substantive to add, just +1'ing @nickykrause 's suggestion :)

PR #2143, opened 2017-06-19
Feature/image number improvements

noahmanger:
This is just an experiment to tackle a couple unrelated image number feature requests. The first is to add an image number column to the main filings table and the second is to add functionality to the main data search so that entering an image number pulls up a filing.

We still need to figure out if/how we want to do this, but it was an easy thing to test so filing a PR so I don't lose track of the work.

Resolves https://github.com/18F/openFEC-web-app/issues/2055
Resolves https://github.com/18F/openFEC-web-app/issues/2054


LindsayYoung:
is there a companion pr that implements the search?

LindsayYoung:
Thanks!

Issue #2141, opened 2017-06-19
Complete consistent table header styling for AOs


jenniferthibault:
Since we're inching closer to having the tags for AOs in place, these styling tweaks should help make the patterns line up closer to 1:1.

**Pattern from data tables and attributes present**
- Title is h1 class
- Underline under title is grey and meets the edge of the filter panel and edge of the screen
- Page title is the data set, without verbiage ("Receipts" instead of "Searching receipts")
- Table headers have a 2px border top that meets the edge of the filter panel and edge of the screen
- Vertical lines between columns extend into header area
- Line height of column headers is tighter (check that we're using the same classes)
- The first row has a white background instead of grey

<img width="1277" alt="screen shot 2017-06-19 at 4 10 10 pm" src="https://user-images.githubusercontent.com/11636908/27303788-e83d35e4-5509-11e7-9d48-97108c998f22.png">

**Current status of AO tables**
- (missing the attributes listed above)
<img width="1277" alt="screen shot 2017-06-19 at 4 21 26 pm" src="https://user-images.githubusercontent.com/11636908/27304233-6763eaa6-550b-11e7-96b4-0a05b73937e1.png">

noahmanger:
Resolved by #2144

PR #2144, opened 2017-06-20
Legal search cleanup

noahmanger:
This started as an endeavor to clean up some of the filter style notes that @jenniferthibault documented in https://github.com/18F/openFEC-web-app/issues/2141 and ended up rolling in a few other cleanup items.

First, it adds the correct classes to the AO search results so they mirror the other data tables.

image

Second, I updated the breadcrumbs across legal search results pages so that they include their immediate parent pages, as well as the main top-level legal resource page. This has been a small thing that's been driving me nuts.

Third, I removed a bunch of the old unnecessary server-side AO search results code, since it all happens client-side now. Mainly that involved refactoring the api caller functions to use `**kwargs` instead of a bunch of `ao_*` named arguments. 

Interestingly enough, doing so, made it really easy to plop in a few of the MUR filters while we wait to extend the new React code base to MURs. 

image

### DIFF
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -57,6 +57,7 @@ def load_legal_search_results(query, query_type='all', offset=0, limit=20, **kwa
         if kwargs[key]:
             filters[key] = kwargs[key]
 
+    if query or query_type in ['advisory_opinions', 'murs']:
         filters['hits_returned'] = limit
         filters['type'] = query_type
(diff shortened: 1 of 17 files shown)


jenniferthibault:
This is awesome! Thanks for getting all these annoying little inconsistencies, AND adding some MUR filters 🎉

nickykrause:
Amazing! Thank you @noahmanger! Adding the `MUR number` filter will be a win for sure

PR #2145, opened 2017-06-20
Feature/remove old election maps

noahmanger:
This change accounts for the fact that district maps before 2012 were different than the data we show, so it hides the district maps on both the election look up and individual house election pages when viewing a cycle before 2011-2012. 

Lookup:
image

Individual page:
image

Before merging, I want to verify with @LindsayYoung that 2012 is the right cut-off to use. 

Resolves https://github.com/18F/FEC/issues/4136


noahmanger:
Also ping @jameshupp for your 👀  on the copy in that message that displays.

noahmanger:
All ready for review. Because the latest redistricting happened this year, I set the cut-off date for 2017-2018 rather than 2012. 

Also, much thanks to @jseppi for helping out getting the districts together. He also wrote a shell script that we can run to easily rebuild in the future.

jameshupp:
Just so I'm clear: this means it's going to say 2017 or later for every state? Or it's going to be whatever the first election cycle since redistricting was for that particular state, whatever that year is?

noahmanger:
It'll be hidden for all locations prior to the 2017-2018 cycle:

image

LindsayYoung:
@jameshupp I took your question as a clarifying question and will merge this. If you have any additional questions or comments we can always make changes later.

Thanks!

Issue #2124, opened 2017-06-12
Implement line number filter front-end


noahmanger:
So that users can browse receipts and disbursements by the line they were reported on, add the line number filter to the front-end.

---

[Design issue](https://github.com/18F/openFEC-web-app/issues/2016)

[Receipts design](https://user-images.githubusercontent.com/1450173/26939730-da41aeea-4c3d-11e7-99a2-08b6e075d636.png)

[Disbursements design](https://user-images.githubusercontent.com/1450173/26939742-e368a668-4c3d-11e7-80fc-53b732d72f88.png)

Depends on the [API implementation](https://github.com/18F/openFEC/issues/2228).

---

Implementation details:
- [x] The filters should be in the "Receipt/disbursement details" panel under the label "Receipt type" or "Disbursement type"
- [x] There should be three dropdowns for each type of form: House and senate committees (form 3), presidential (form 3P) and pacs/parties (form 3X)
- [x] Each dropdown should include the plain text label of each line with the line number in parentheses
- [x] When viewing contributions before 2007 the filters should be disabled and a message should be shown
- [x] Each filter should generate a tag that includes the full text label + the type of committee that reported it, with line number in parentheses

noahmanger:
Resolved by https://github.com/18F/openFEC-web-app/pull/2150

PR #2150, opened 2017-06-22
Feature/filter line number

xtine:
Front end filters for form 3/3p/3x on receipts and disbursements. https://github.com/18F/openFEC-web-app/issues/2124

Filters:
screen shot 2017-06-23 at 9 51 07 am

Hide line number filters on years before 2007:
screen shot 2017-06-23 at 9 50 54 am

Tags with prefixed labels:
screen shot 2017-06-23 at 10 52 03 am


static/js/modules/filters-event.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+var $ = require('jquery');
+var _ = require('underscore');
+
+var FilterPanel = require('fec-style/js/filter-panel').FilterPanel;
+var URI = require('urijs');
+
+function lineNumberFilters() {
+  this.filterPanel = new FilterPanel();
```
xtine: I tried to just have the listener on `$('#filters form')` but it doesn't seem to work. I'm guessing it's because `FilterSet()` activates all the various form handlers for events to be attached to?

noahmanger: Hm that's strange. It's not actually a `form` element, so maybe that's it? Trying something...

noahmanger: Ah yeah, doing just `$('#filters').on` seems to do it.

xtine:
Got the events to work on just `$('#filters')`. Thanks for that suggestion. Also fixed the multiple filter numbers showing by making them a `{{ checkbox_dropdown_multiple }}` that prevents that issue.

noahmanger:
Awesome work here!

Issue #2155, opened 2017-06-26
Bug: Line number filters don't show tag when loading from URL


noahmanger:
I noticed that if you go to a URL with the line number filter applied ([example](https://fec-stage-proxy.app.cloud.gov/data/receipts/?two_year_transaction_period=2016&data_type=processed&min_date=01%2F01%2F2015&max_date=12%2F31%2F2016&line_number=F3P-17D)) the filter is applied but there's not a tag generated. 

Can you look into this @xtine ?

xtine:
Ah, I think somehow, the function to do a line number check upon load was accidentally removed https://github.com/18F/openFEC-web-app/blob/release/public-20170629/static/js/modules/filters-event.js#L12

noahmanger:
This was fixed a while ago.

PR #2156, opened 2017-06-26
bugfix: check line number filter on load

xtine:
Fixes: https://github.com/18F/openFEC-web-app/issues/2155


noahmanger:
I see what that's doing, which is right, but doesn't address the tag not being generated:
image

xtine:
Interestingly, it will bring up the tags if more than one line number filter is selected, but not one. Looking into this.

Issue #2125, opened 2017-06-12
Link financial summary lines to filtered views of transactions


noahmanger:
So that users can get a granular view on the data reported by committees, link each line of the financial summaries to filtered views.

---

Once we have the line number filter implemented (https://github.com/18F/openFEC-web-app/issues/2124), we can link all of the financial summary lines to filtered views.

![image](https://user-images.githubusercontent.com/1696495/27056737-43e67a40-4f7e-11e7-8879-6c0bc454fd87.png)

Each one of those lines corresponds to a specific form line number. 

**Completion criteria:**
- [x] Each line in house/senate committee summaries are linked
- [x] Each line in pac/party summaries are linked
- [x] Each line in presidential committee summaries are linked
- [x] Lines on all candidate pages are linked
- [x] Lines are only linked in time periods after 2006

noahmanger:
Resolved by https://github.com/18F/openFEC-web-app/pull/2169

PR #2169, opened 2017-07-07
Feature/financial summary lines

xtine:
Link filtered views on financial summaries. https://github.com/18F/openFEC-web-app/issues/2125

Will need another set of 👀 to double check I didn't miss or mislabel any form lines.

### DIFF
--- a/openfecwebapp/constants.py
+++ b/openfecwebapp/constants.py
@@ -465,7 +465,7 @@ SPENDING_FORMATTER = OrderedDict([
     ('operating_expenditures',  # F3, F3P, F3X
         {'label': 'Operating expenditures', 'term': 'operating expenditures',
             'level': '2', 'type': {'link': 'disbursements',
-                'P': 'F3P-23', 'H': 'F3-17', 'S': 'F3-17', 'O': 'F3X-21'}}),
+                'P': 'F3P-23', 'H': 'F3-17', 'S': 'F3-17'}}),
     ('shared_fed_operating_expenditures',  # F3X
         {'label': 'Allocated operating expenditures - federal', 'level': '3'}),
     ('shared_nonfed_operating_expenditures',  # F3X
--- a/openfecwebapp/constants.py
+++ b/openfecwebapp/constants.py
@@ -474,7 +474,7 @@ SPENDING_FORMATTER = OrderedDict([
         {'label': 'Other federal operating expenditures', 'level': '3',
             'type': {'link': 'disbursements', 'O': 'F3X-21B'}}),
     ('transfers_to_other_authorized_committee',  # F3, F3P
-        {'label': 'Transfers to authorized committees', 'level': '2', 'type': {
+        {'label': 'Transfers to other authorized committees', 'level': '2', 'type': {
             'link': 'disbursements', 'H': 'F3-18', 'S': 'F3-18', 'P': 'F3P-24'
         }}),
     ('fundraising_disbursements',  # F3P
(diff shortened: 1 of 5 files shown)


openfecwebapp/constants.py
```diff
@@ -345,74 +345,220 @@
...
+        {'label': 'Total individual contributions', 'level': '3'}),
+    ('individual_itemized_contributions',  # F3, F3P, F3X
+        {'label': 'Itemized individual contributions', 'level': '4',
+            'link': 'individual_contributions'}),
+    ('individual_unitemized_contributions',  # F3, F3P, F3X
+        {'label': 'Unitemized individual contributions', 'level': '4'}),
+    ('political_party_committee_contributions',
+        {'label': 'Party committee contributions', 'level': '3', 'type': {
+            'link': 'receipts',
+                  'P': 'F3P-17B', 'H': 'F3-11B', 'S': 'F3-11B', 'O': 'F3X-11B'
```
noahmanger: Just a suggestion for future clarity: what about instead of P, H, S and 0, it was just the form number: 3, 3P, 3X (or 3, P, X)? H and S will always be the same (form 3), so it would also remove an unnecessary item.

xtine: The P, H, S, O are because it cross references with the passed `office` and `committee_type` variables from the view. Although now thinking about it I could just try to have the view pass an additional form number variable.

Also I've been thinking about how to merge the `line_numbers` constants with the formatter ones. So perhaps this PR can be merged in while I figure out how to best unify the constants  so we have one object instead of three.

noahmanger: Yeah that would make a lot of sense to consolidate. Ok, we can leave this as-is, but then do that next.

openfecwebapp/constants.py
```diff
@@ -345,74 +345,220 @@
...
+        {'label': 'Other committee contributions', 'level': '3', 'type': {
+            'link': 'receipts',
+                  'P': 'F3P-17C', 'H': 'F3-11C', 'S': 'F3-11C', 'O': 'F3X-11C'
+        }}),
+    ('federal_funds',  # F3, F3P
+        {'label': 'Presidential public funds', 'level': '3'}),
+    ('candidate_contribution',  # F3, F3P
+        {'label': 'Candidate contributions', 'level': '3', 'type': {
+            'link': 'receipts',
+                  'P': 'F3P-17D', 'H': 'F3-11D', 'S': 'F3-11D', 'O': 'F3X-11D'
```
noahmanger: `0` should not have an option. This is only on F3 and F3P.

openfecwebapp/constants.py
```diff
@@ -345,74 +345,220 @@
...
+        }}),
+    ('transfers_from_affiliated_party',  # F3X
+        {'label': 'Transfers from affiliated committees', 'level': '2',
+            'type': {'link': 'receipts', 'O': 'F3X-12'}}),
+    ('transfers_from_affiliated_committee',  # F3P
+        {'label': 'Transfer from affiliated committees', 'level': '2',
+            'type': {'link': 'receipts', 'P': 'F3P-18'}}),
+    ('transfers_from_other_authorized_committee',  # F3
+        {'label': 'Transfer from authorized committees', 'level': '2',
+            'type': {'link': 'receipts', 'H': 'F3-12', 'S': 'F3-12'}}),
```
noahmanger: Transfers from authorized committees needs `'P': 'F3P-18'`.

xtine:
Thanks for the corrections, @noahmanger !

PR #2188, opened 2017-07-13
Removing more lines from F3 totals

noahmanger:
One more pass at removing errant presidential totals items from congressional candidate pages.


LindsayYoung:
Thanks so much!

Issue #2177, opened 2017-07-11
Update front-end to allow for full time period searches


noahmanger:
So that users can search for receipts and disbursements across all time, update the front-end filters to not restrict to a two-year period.

---

This is dependent on the API work happening but we can work on figuring out what we want this to do.

## Update filter logic
This is the trickiest part, since we want to start by first expanding to 6 years to make sure we're ok performance-wise. Presumably we could get this to work using the current filter validation logic, but by changing from a 2 year range to a 6 year range. It raises some questions:

- Should users select any even year (2016, 2018, 2020, etc.) and then the period becomes the six years preceding that year? I think this makes the most sense, because that way you could have an given full senate cycle.
- Do we show the date picker?
- Do we need to pass something like `six_year_transaction_period` or is it enough to just enforce the time constraint client-side?

I got a start experimenting with this in https://github.com/18F/openFEC-web-app/pull/2213 and https://github.com/18F/fec-style/pull/747. 

## Update links to receipts & disbursements pages
- [x] Remove two-year period filter and validation logic
- [ ] Consider whether or not we want a default time period of some length, even if it can be removed
- [x] Go through and update all links to these pages that include the transaction period parameter
- [x] When on a candidate profile and looking at a 4 or 6 year period, the links should go to the full 4 or 6 years by using `min_date` and `max_date`
- [x] Update all links on election pages
- [x] Ensure proper formatting of all date params

I got a start with this in https://github.com/18F/openFEC-web-app/pull/2202 , though that PR goes a little far in completely removing the time period restriction all together. 

## Update candidate pages
- [x] Update candidate page to query 4 and 6 years of transactions
- [ ] Candidate page totals need to be adjusted to allow 4 or 6 year totals, but this requires API work. Is it ok to keep these restricted to two year ranges? If so, make sure that coverage dates make sense. Or maybe they should be hidden if looking at a full election range?

noahmanger:
The trickiest part about this is figuring out the right interaction for a six-year range. Unlike with two years where it makes sense to ask somebody to select a two year period and then adjust dates within it, 6 year ranges are more difficult. Here's a couple attempts, with their challenges.

### Option 1: Complex, adaptation of what we have now
![image](https://user-images.githubusercontent.com/1696495/29101005-5d2e4b28-7c64-11e7-9ca3-d42a8d800f91.png)

One direction would be to take the current component but have the cycle select count for 6 years and have all the date selection and validation and everything work the same way.

[...]

PR #2202, opened 2017-07-19
[WIP] Feature/remove time period

noahmanger:
This is an initial pass of front-end changes in preparation for removing the `two_year_transaction_period` filter so that we can test it once the API work is done. 

First, it replaces the time-period constricted date filters with regular min and max fields. We may decide we want something more sophisticated, but it's a start. 

Second, I basically went through and replaced all instances where a `two_year_transaction_period` was passed as a URL parameter with `min_date` and `max_date`, as defined by the coverage start and end dates for the page you're looking at. 

Last, I also did a little cleanup throughout by using the global `context` object to pass more data to the data tables on the candidate pages, rather than setting everything as data attributes on the tables themselves.

Don't merge until the API work is done.

**To do:**
- [ ] Decide what to do about coverage dates on candidate pages for the aggregates when viewing a 4 or 6 year period

Resolves https://github.com/18F/openFEC-web-app/issues/2177


noahmanger:
Closing in favor of https://github.com/18F/openFEC-web-app/pull/2213

PR #2213, opened 2017-07-24
6 year date selection

### LINKED ISSUE #2133: Link aggregates to 4 or 6 year ranges
On the other spending tab of a candidate page, if you're looking at a 4 or 6 year range, the aggregates in the IE/EC/CC tables all still include the min and max dates for a 2 year range.

image

In order for the itemized data to accurately reflect the totals, we should make the min date equal to the start of the 4 or 6 year period.

The current behavior is because we have a single helper function that generates min and max dates for a two-year period: https://github.com/18F/openFEC-web-app/blob/master/static/js/modules/helpers.js#L154

h/t @PaulClark2 for catching this. cc @xtine .

noahmanger:
This makes several changes to the front-end to support 6 years of schedule A and B data at a time.

- Adds a new macro for a date validation filter, which requires https://github.com/18F/fec-style/pull/759 . This filter checks that the years entered are within 6 years of each other.
- Updates links from candidate, committee and election pages to pass through min and max dates, rather than the two year transaction period
- On candidate pages, when looking at a full 4 or 6 year period, the full date range is passed through in links to the itemized data

This also includes a fix so that other spending aggregates on candidate page link to the full 4 or 6 year time period for IEs, ECs, and CCs (resolves https://github.com/18F/openFEC-web-app/issues/2133).


static/js/modules/column-helpers.js
```diff
@@ -88,26 +88,13 @@ function buildEntityLink(data, url, category, opts) {
...
-  }
-}
-
 function buildTotalLink(path, getParams) {
   return function(data, type, row, meta) {
     data = data || 0;
     var params = getParams(data, type, row, meta);
     var span = document.createElement('div');
     var includeTransactionPeriod = false;
+    var duration = meta.settings.nTable.dataset.duration ? Number(meta.settings.nTable.dataset.duration) : 2;
```
xtine, line 97: Why are these variables so long?

static/js/pages/committee-single.js
```diff
@@ -302,8 +302,10 @@ $(document).ready(function() {
   // Set up data tables
   $('.data-table').each(function(index, table) {
     var $table = $(table);
-    var committeeId = $table.attr('data-committee');
-    var cycle = $table.attr('data-cycle');
+    var committeeId = window.context.committeeID;
```
xtine, line 305: Is there a way in which we don't have to rely on global variables?

jenniferthibault:
Closing this in favor of work in https://github.com/18F/fec-cms/issues/1371 which will use the work started here, but will need to be refactored to work in the `fec-cms repo`

PR #2230, opened 2017-07-28
Feature/filings filters

noahmanger:
This implements the front-end changes to take advantage of the API data provided in https://github.com/18F/openFEC/pull/2550

- Adds state, office and party filters to /filings
- Adds a filter for "first filing date" to candidate and committee pages, along with a column for this and makes it so that these pages sort reverse-chronological on that field so that users see the most recent new candidates and committees by default

Resolves https://github.com/18F/openFEC/issues/2378

### DIFF
--- a/openfecwebapp/constants.py
+++ b/openfecwebapp/constants.py
@@ -233,11 +233,11 @@ house_senate_types = OrderedDict([
 ])
 
 table_columns = OrderedDict([
-    ('candidates', ['Name', 'Office', 'Election years', 'Party', 'State', 'District']),
+    ('candidates', ['Name', 'Office', 'Election years', 'Party', 'State', 'District', 'First filing date']),
     ('candidates-office-president', ['Name', 'Party', 'Receipts', 'Disbursements']),
     ('candidates-office-senate', ['Name', 'Party', 'State', 'Receipts', 'Disbursements']),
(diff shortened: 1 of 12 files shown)


noahmanger:
In testing this out, it seems that `/candidates/` is missing `first_file_date`. I see a `last_file_date` but not `first_file_date`. @jontours can you check it out?

jontours:
I can see if there is a quick fix this morning.

jontours:
After a quick look there is no explainable reason why the feature isn't working.  When I tested this I ran update_schemas and candidates like McCain's presidential were showing correct `first_file_date` and now it's not there.  I'd suspect the dev database is not updated, I'll run update_schemas to force the update, but there seems to be larger problem at hand with none of the updates running.

jontours:
Ok, another update.  The data is in the database and it's still not showing.  I'm suspecting the api didn't actually deploy with the new code changes.  We've seen that happen before with Travis and cloudfoundry not actually deploying.  Maybe restage the api @ccostino?

ccostino:
@jontours it's possible given the build issues we've been having in Travis (though I think this change was done last week and should be there, if I remember correctly).  https://github.com/18F/openFEC/pull/2570 definitely needs to get merged to get builds in the API working again given the failing tests issue.  That should get the latest code up there again.

noahmanger:
This is all ready! Candidates page looks good now:
image

Issue #2198, opened 2017-07-17
Design new button linking to candidate comparison pages from candidate profile


noahmanger:
So that users can easily compare candidates running in a race, create a more prominent link to comparison pages from candidate profile.

---
Concept:
![concept](https://user-images.githubusercontent.com/1450173/28283639-ecdddae0-6af3-11e7-9881-4f7638b985ce.png)

Follow-up from https://github.com/18F/openFEC-web-app/issues/2167

**Completion criteria:**
- [x] Figure out where the button lives on the page
- [x] Figure out what the button text is
- [x] Figure out style of the button
- [x] Design end-to-end interaction for the button: where does it take you, how do you get back?

jenniferthibault:
### Button style & placement
Since this needs more attention than a regular link, but it a little much text for a regular button, it feels like a good case for our small card pattern.

We had initially marked out space for a quick intro line to the Financial summary (and other) sections, so I've pieced that in as part of this since it helps with some of the layouts.

👀 ing three different locations:

1. Fixed below the menu, which keeps it accessible from any tab but competes in visual hierarchy for attention with the menu

![screen shot 2017-07-27 at 8 58 19 pm](https://user-images.githubusercontent.com/11636908/28698211-c0b46a00-730f-11e7-8726-af01e3212b4e.png)

__________________________
2. Fixed in the title space. Would need to drop the standard card pattern for a large button instead, since the space is too short, but would keep it available from any tab

![screen shot 2017-07-27 at 8 58 01 pm](https://user-images.githubusercontent.com/11636908/28698216-c78cddb2-730f-11e7-84aa-4ff2513d01c3.png)

__________________________
3. In the right-hand column. This feels like a really good place for a card element, since it's a familiar use of space, but puts it only on the financial summary tab. Since this is usually the first tab, that's ok, but we'd probably want to repeat it _at the least_ on the "About this candidate" tab

![screen shot 2017-07-27 at 8 57 04 pm](https://user-images.githubusercontent.com/11636908/28698219-ce9a7b64-730f-11e7-8a01-b884b0d22d93.png)


__________________________
### End to end experience: how do users get back

- Selecting the candidate from the list on the election page
- Using the browser "back" button

<img width="990" alt="screen shot 2017-07-27 at 9 01 58 pm" src="https://user-images.githubusercontent.com/11636908/28698247-11a7b0c0-7310-11e7-8093-15d0f0c6e5f7.png">


[...]

PR #2231, opened 2017-07-28
Feature/elections improvements

### LINKED ISSUE #2204: Maps on campaign pages aren't rendering
In the election pages like:
https://www.fec.gov/data/elections/president/2016/
screen shot 2017-07-19 at 3 45 20 pm

if you click on the map you get an error
 console error "Cannot read property 'aDataSort' of undefined"

noahmanger:
This is some old work I started to re-do the elections pages in the new templates. It also contains the new comparison button for candidate pages, though needs the design from https://github.com/18F/openFEC-web-app/issues/2198 applied.

---
Update: this is ready now. I finished up the styling by pairing with @jenniferthibault , added the compare button to the candidate page, and also fixed a bug where the maps on the comparison page weren't showing (resolves https://github.com/18F/openFEC-web-app/issues/2204).

image

image

image

image


openfecwebapp/templates/elections.html
```diff
@@ -13,73 +12,106 @@
...
+            href="#section-1"
+            >Candidate financial totals</a>
+        </li>
+        <li class="side-nav__item" role="presentation">
+          <a
+            class="side-nav__link"
+            role="tab"
+            data-name="contributions"
+            tabindex="0"
+            aria-controls="panel3"
```
xtine: Is this supposed to be `panel2` (and next `panel3`)?

openfecwebapp/filters.py
```diff
@@ -85,3 +85,7 @@ def get_max(list):
 @app.template_filter()
 def get_min(list):
     return min(list)
+
+@app.template_filter()
+def strip_zero_pad(number):
```
xtine: Would this template filter be better named as just `strip` if we need to strip other characters than zero padded numbers?

xtine (approved):
Looks good, just have a few minor questions.

openfecwebapp/templates/elections.html
noahmanger: Fixed!

openfecwebapp/filters.py
noahmanger: Well this filter just strips out the `0` so I think that makes sense.

PR #2233, opened 2017-07-31
[WIP] Make api_caller logs more verbose

LindsayYoung:
This error message includes the status code of the API call that fails. Have not tested locally yet

### DIFF
--- a/openfecwebapp/api_caller.py
+++ b/openfecwebapp/api_caller.py
@@ -1,3 +1,4 @@
+import logging
 from operator import itemgetter
 import os
 from urllib import parse
@@ -15,6 +16,8 @@ from collections import OrderedDict
 
 MAX_FINANCIALS_COUNT = 4
 
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
 
 session = requests.Session()
 http_adapter = requests.adapters.HTTPAdapter(max_retries=2)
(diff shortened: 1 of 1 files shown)


openfecwebapp/api_caller.py
```diff
@@ -34,7 +37,11 @@ def _call_api(*path_parts, **filters):
 
     results = session.get(url, params=filters)
 
-    return results.json() if results.ok else {}
+    if results.ok:
+        return results.json()
+    else:
+        logger.error('API ERROR with status {0} for {1} with filters: {2}'.format(results.status_code, url, filters))
```
ccostino: This will give enough of a clue to know what to check for on the API side, but would it make sense to also include the exception message thrown by the API too, or do you think that's not necessary here?

Either way, this definitely provides much better visibility into a request that failed when going through the `api_caller`!

ccostino (commented):
I just tested this out locally and it works great, thanks @LindsayYoung!  I had one question/suggestion, but this could be good to go as is too I think.  Did you have anything else in mind to do here?

openfecwebapp/api_caller.py
LindsayYoung: By exception message do you mean the message from the API if there is one?

ccostino: Yup!

jenniferthibault:
@LindsayYoung @ccostino how should we handle this PR now that we've merged repos?

ccostino:
Oooh this is a good one to pick back up since it is not incorporated into the merge, thanks @jenniferthibault!

The changes are lightweight enough that we can just open a new PR on the `fec-cms` side and continue in there.  I'll get that done now so we can close this. :-)

ccostino:
https://github.com/18F/fec-cms/pull/1537 now in place, so this is good to close!

PR #2238, opened 2017-08-03
Feature/archive mur tweaks

noahmanger:
This removes the lines about archived MURs not being included, and also fixes a styling bug on the legal search results page as a result of the CSS refactor.


xtine:
@noahmanger: just needs a conflict fix and then I can merge in.

PR #2242, opened 2017-08-04
Use .csv extension for download filename

vrajmohan:


noahmanger:
I'm assuming we want to wait to merge this until the API change is done? https://github.com/18F/openFEC/pull/2584

vrajmohan:
True. Even more, we need to discuss how to deploy this so that no downloads are clobbered.

Issue #2244, opened 2017-08-04
Bug: "Home" breadcrumb link in Data and Legal sections broken


patphongs:
In user testing today, we uncovered that the home breadcrumb link is broken in the data and legal sections of the website. The `href` seems to be empty.

jenniferthibault:
Thanks for filing this @patphongs !

As seen:

https://www.fec.gov/data/advanced/

![screen shot 2017-08-04 at 4 58 29 pm](https://user-images.githubusercontent.com/11636908/28986726-283cd69a-7936-11e7-8064-13cd93532b03.png)

PR #2253, opened 2017-08-08
Adding cms_url to manifests

noahmanger:
The issue with the "Home" link in the breadcrumbs not linking to the home page (https://github.com/18F/openFEC-web-app/issues/2244) appears to be due to a missing `FEC_CMS_URL` env variable. This adds it to the manifests for all three environments.

Resolves https://github.com/18F/openFEC-web-app/issues/2244

### DIFF
--- a/manifest_dev.yml
+++ b/manifest_dev.yml
@@ -4,6 +4,7 @@ host: fec-dev-web
 services:
   - fec-creds-dev
 env:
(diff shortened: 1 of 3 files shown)


noahmanger:
AFAIK, yes. These are just used to build absolute links to CMS pages from the web app.

Issue #2179, opened 2017-07-11
Extend "late filing" retention on filings pages from 2 to 3 days


jwchumley:
A few sprints back we added the ability to display reports from the real_efile tables if they were too late to make the overnight process. I believe we set them to be displayed for 2 days assuming they would be in the processed data by then. That was a work around because the classic process that displays them for 3 days or until they show up in the processed data was too complicated to implement in the short time frame. 

I think we need to either extend the time to 3 days or implement the more complex -Display until the report shows up in the processed data because we had a report that did this:

Jul 9: Filed late in the evening
Jul 9: It was probably displaying correctly
Jul 10: it was probably displaying correctly
Jul 10 overnight process failed
Jul 11: It fell off the public record because the two days had passed.
Jul 12: It should be fine because the overnight process the previous night should load it.

@noahmanger

noahmanger:
So in this case, the issue is that the committee record is not in the API. The earlier change to display efiled reports for two days was _only_ on the web app side. So I don't think extending from 2 to 3 days will actually solve the issue.

cc @LindsayYoung

jwchumley:
I'll have to check and find out how the classic viewer handles it. It displayed there when it wasn't displaying on the new site. It may be that the viewer creates a virtual row for the committee from the efiling data.


________________________________

AmyKort:
Is this ready to close?

AmyKort:
Issue moved to [18F/fec-cms #1539](https://github.com/18F/fec-cms/issues/1539) via [**ZenHub**](https://www.zenhub.com/)

PR #2258, opened 2017-08-11
Extend efiling window + refactor

noahmanger:
This extends the window for showing efilings on a committee page from two days to three, for all the [reasons laid out here](https://github.com/18F/openFEC-web-app/issues/2179). 

It also makes this a little easier to configure by making the previous `two_days_ago` filter into `n_days_ago` and then setting a constant for `EFILING_WINDOW`.

Resolves https://github.com/18F/openFEC-web-app/issues/2179

### DIFF
--- a/openfecwebapp/app.py
+++ b/openfecwebapp/app.py
@@ -147,7 +147,7 @@ app.jinja_env.globals.update({
     'series_group_has_data': series_group_has_data,
     'cycle_start': cycle_start,
     'cycle_end': cycle_end,
-    'two_days_ago': utils.two_days_ago(),
+    'efiling_window': utils.n_days_ago(constants.EFILING_WINDOW),
     'election_url': get_election_url,
     'constants': constants,
     'cycles': utils.get_cycles(),
--- a/openfecwebapp/constants.py
+++ b/openfecwebapp/constants.py
@@ -5,6 +5,7 @@ END_YEAR = 2018
 DEFAULT_TIME_PERIOD = 2018
 DEFAULT_PRESIDENTIAL_YEAR = 2016
 DISTRICT_MAP_CUTOFF = 2018 # The year we show district maps for on election pages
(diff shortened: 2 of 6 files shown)


jenniferthibault:
@LindsayYoung how should we handle this PR now that we've merged repos?

jenniferthibault:
@LindsayYoung @ccostino this is the last PR/issue in the `openFEC-web-app` repo. Can you help figure out how we should handle it? The issue Noah pointed to above that this addresses has been moved to `fec-cms` in https://github.com/18F/fec-cms/issues/1539, which was closed because there was no response to prompts in the comment thread (but I'm not sure that means it was no longer relevant?)

ccostino:
It looks like based on the conversation history this work will handle the display of things on the front end, but there may not be a noticeable difference as the API doesn't return the information being requested at the moment (it only returns the 2-day period if I understand this correctly).

We could probably go ahead and move this work over to the CMS and merge it without an issue based on what I see here, then circle back on it and see if we still need to make an API adjustment.  Does that sound good, @LindsayYoung?

LindsayYoung:
I think we did this let me double check

LindsayYoung:
Ok it is not implemented, but it seems like that is because it hasn't been a priority. We thought it would fix a problem about latent reports, but it seems like the cases we had seen before, that wouldn't have helped anyway. I am going to close this for now, it is a simple change so we can just open a new ticket and do this in the future if we think it would be beneficial. (Feel free to re-open if you disagree)

Here is the current code: https://github.com/18F/openFEC-web-app/blob/e23ff54e822ed491bcd78b2ab407c544b39914ac/openfecwebapp/templates/macros/entity-pages.html#L43

Previous discussions:
https://github.com/18F/openFEC-web-app/issues/2179
https://github.com/18F/fec-cms/issues/1539

Issue #2256, opened 2017-08-10
Election results page does not list all districts


patphongs:
When searching an entire state, it does not list all of the districts in the search results. This has not been tested on every state, but this also seems to be a problem with Texas too. So possibly other large states might have this issue.

## Example from production:
California has 53 districts. But when searching for California, the search results only display 19 districts for the 2017-2018 cycle. Number of districts also change when you select a different 2 year period. It should list all 53 districts. 

![election-results-bug](https://user-images.githubusercontent.com/12799132/29182244-2e9f840c-7dcc-11e7-8425-6f08885f5a58.gif)

The dropdown district list when selecting California does show 53 districts, which is correct.

![screen shot 2017-08-10 at 12 26 58 pm](https://user-images.githubusercontent.com/12799132/29180895-3bd6298c-7dc7-11e7-9bcf-a22d3b8ecd0b.png)

noahmanger:
This is a front-end issue. We need to pass `per_page=100` to the API call.

noahmanger:
Including this fix in https://github.com/18F/openFEC-web-app/pull/2260

noahmanger:
Resolved by https://github.com/18F/openFEC-web-app/pull/2260

PR #2260, opened 2017-08-14
Add links to election pages from data landing

noahmanger:
I forgot to include actual links on this part of the data landing, so this adds them:

image

It also adds the correct link to the election page from candidate pages:
image

And updates the state + district selects on election pages to go to the page similar to how it works on the data landing page:
image

And last, I fixed an issue where not all district results were showing up when doing a state search by adding `per_page=100` to API calls for the election lookup. Resolves https://github.com/18F/openFEC-web-app/issues/2256

### DIFF
--- a/openfecwebapp/templates/candidates-single.html
+++ b/openfecwebapp/templates/candidates-single.html
@@ -107,7 +107,9 @@
             </ul>
         </li>
         <li class="side-nav__item">
-          <a class="button button--cta u-margin--top t-left-aligned button--two-candidates u-full-width">Compare to<br>opposing candidates</a>
+          <a class="button button--cta u-margin--top t-left-aligned button--two-candidates u-full-width"
+             href="{{ election_url(context(), cycle)}}"
+          >Compare to<br>opposing candidates</a>
         </li>
       </ul>
     </nav>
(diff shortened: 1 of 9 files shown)


noahmanger:
Added a few more things to this PR and updated the message accordingly.

openfecwebapp/templates/candidates-single.html
```diff
@@ -107,7 +107,9 @@ <h1 class="entity__name">{{ name }}</h1>
             </ul>
         </li>
         <li class="side-nav__item">
-          <a class="button button--cta u-margin--top t-left-aligned button--two-candidates u-full-width">Compare to<br>opposing candidates</a>
+          <a class="button button--cta u-margin--top t-left-aligned button--two-candidates u-full-width"
+             href="{{ election_url(context(), cycle)}}"
+          >Compare to<br>opposing candidates</a>
```
xtine: Can you fix the formatting/indent so the a tag is closed?

static/js/pages/elections.js
```diff
@@ -15,6 +15,7 @@ var tables = require('../modules/tables');
 var columnHelpers = require('../modules/column-helpers');
 var columns = require('../modules/columns');
 var helpers = require('../modules/helpers');
+var ElectionForm = require('../modules/election-form').ElectionForm;
```
xtine: Can you instantiate this outside of the require?

PR #2263, opened 2017-08-16
Feature/boolean search

noahmanger:
This is a companion PR to https://github.com/18F/fec-cms/pull/1205 , adding the boolean search field modal to the advanced legal resources search pages. It adds it to AOs, MURs, statutes and regs, and basically works the same in all cases.

demo

Note, rather than putting the `keyword-modal.js` code in fec-style, I just made two copies, one here and one in the CMS, since we're going to merge the apps soon. There's just a slight logic difference between the two, which I noted in the comments. The key difference is that for these searches, rather than submitting the form, it just updates the `?search=` param in the URL. This way it works with the React AO filters as well as the regular HTML form fields for the other searches. 

The one trade-off at the moment is that on the AO page, it forces a full refresh of the page. This can maybe be resolved in a later step, but didn't feel worth holding up.

I also fixed the JSHint errors by telling it to ignore files in the `/legal/` directory, which use newer JS that the current JSHint settings aren't configured for. And I moved the `build-legal` build script to the gulpfile for consistency and to get sourcemaps working. 

Thanks to @anthonygarvan for the guidance.


xtine:
Front-end looks good. Also modal works well in mobile as well.

PR #2276, opened 2017-08-23
Add Alabama to 2018 senate class

noahmanger:
Currently on staging, going to [Alabama Senate 2018](https://fec-stage-proxy.app.cloud.gov/data/elections/senate/AL/2018/), the cycle select shows 2021-2022, which is the date of the next election. 

image

This change adds Alabama to the 2018 Senate class.


LindsayYoung:
Thanks @noahmanger this makes sense we can get something dynamic in as a follow up

